PR #45, opened 2016-03-29
CODENVY-233: fix error of executing IM CLI command after update

dmytro-ndp:


tolusha:
Do we need to change the name of the update script?

dmytro-ndp:
Old name of update cli script *codenvy-cli-update-script.sh* is little bit misleading because we are about update IM CLI, and it contents odd extension *.sh*.

tolusha:
+1

PR #46, opened 2016-03-30
CODENVY-141: add link to Codenvy Classic 

ashumilova:


ashumilova:
link_hover
link_to_codenvy

PR #58, opened 2016-03-31
CODENVY-235: check write access to install directory; code cleanup

dmytro-ndp:

### DIFF
--- a/cdec/installation-manager-resources/src/main/resources/im-scripts/install-codenvy.sh
+++ b/cdec/installation-manager-resources/src/main/resources/im-scripts/install-codenvy.sh
@@ -10,6 +10,7 @@
 # --systemAdminName=<SYSTEM ADMIN NAME>
 # --systemAdminPassword=<SYSTEM ADMIN PASSWORD>
 # --fair-source-license=accept
+# --install-directory=<PATH_TO_SINGLE_DIRECTORY_FOR_ALL_RESOURCES>
 
 trap cleanUp EXIT
 
@@ -22,6 +23,8 @@ JRE_URL=http://download.oracle.com/otn-pub/java/jdk/8u45-b14/jre-8u45-linux-x64.
 
 PUPPET_AGENT_PACKAGE=puppet-3.8.6-1.el7.noarch
 PUPPET_SERVER_PACKAGE=puppet-server-3.8.6-1.el7.noarch
+
+INSTALL_DIRECTORY=$PWD
 
 EXTERNAL_DEPENDENCIES=("https://codenvy.com||0"
                        "https://install.codenvycorp.com||0"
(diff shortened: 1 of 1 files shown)


cdec/installation-manager-resources/src/main/resources/im-scripts/install-codenvy.sh
```diff
@@ -928,6 +941,15 @@ doCheckAvailableResourcesLocally() {
         exit 1;
     fi
 
+    if [[ ${writeAccessIssueFound} == true ]]; then
+        if [[ -n $INSTALL_DIRECTORY ]]; then
+            println "$(printError "ERROR: Installation directory") $(printImportantLink $INSTALL_DIRECTORY) $(printError "cannot be written to.")"
+        else
+            println $(printError "ERROR: Current directory doesn't have write access.")
```
tolusha: it is better to say: User doesn't have write permissions to directory

tolusha:
New param should be reflected in the doc

cdec/installation-manager-resources/src/main/resources/im-scripts/install-codenvy.sh
```diff
@@ -914,11 +920,18 @@ doCheckAvailableResourcesLocally() {
...
+        writeAccessIssueFound=true
+        writeAccessStateToDisplay=$(printf "%-44s" && printError "[NOT OK]")
+    fi
+
     println
     println "                MINIMUM        RECOMMENDED     AVAILABLE"
     println "RAM             $minRAMToDisplay $recRAMToDisplay $availableRAMToDisplay $RAMStateToDisplay"
     println "CPU             $minCoresToDisplay $recCoresToDisplay $availableCoresToDisplay $coresStateToDisplay"
     println "Disk Space      $minDiskSpaceToDisplay $recDiskSpaceToDisplay $availableDiskSpaceToDisplay $diskStateToDisplay"
+    println "Write Access    $writeAccessStateToDisplay"
```
tolusha: Access -> Permissions ?

cdec/installation-manager-resources/src/main/resources/im-scripts/install-codenvy.sh
```diff
@@ -928,6 +941,15 @@ doCheckAvailableResourcesLocally() {
         exit 1;
     fi
 
+    if [[ ${writeAccessIssueFound} == true ]]; then
+        if [[ -n $INSTALL_DIRECTORY ]]; then
+            println "$(printError "ERROR: Installation directory") $(printImportantLink $INSTALL_DIRECTORY) $(printError "cannot be written to.")"
```
tolusha: to -> into

tolusha:
What will be the default directory to install IM into ?

TylerJewell:
The default should be:
<current-directory>/codenvy ... meaning that we create a codenvy directory to hold everything first, and then put codenvy.properties, downloads, etc into that directory.

tolusha:
@dmytro-ndp I don't see of INSTALL_DIRECTORY is used.

cdec/installation-manager-resources/src/main/resources/im-scripts/install-codenvy.sh
dmytro-ndp: That is how Tyler defined the message in description to the https://jira.codenvycorp.com/browse/CODENVY-235
But, I agree - "Write Permission" is more suitable here.

cdec/installation-manager-resources/src/main/resources/im-scripts/install-codenvy.sh
dmytro-ndp: That is how it defined [here](https://jira.codenvycorp.com/browse/CODENVY-235?focusedCommentId=273935&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-273935).

cdec/installation-manager-resources/src/main/resources/im-scripts/install-codenvy.sh
TylerJewell: If you use "Write Permission" then you will have to adjust the columns as the formatting will be altered.

dmytro-ndp:
It will be used in the issue [CODENVY-317](https://jira.codenvycorp.com/browse/CODENVY-235).

cdec/installation-manager-resources/src/main/resources/im-scripts/install-codenvy.sh
dmytro-ndp: I don't think it is good message so as it doesn't say which user doesn't have access to which directory.

dmytro-ndp:
@tolusha
> New param should be reflected in the doc
Which doc do you mean? There is description already [here](https://github.com/codenvy/codenvy/blob/CODENVY-235/cdec/installation-manager-resources/src/main/resources/im-scripts/install-codenvy.sh#L13).

dmytro-ndp:
@tolusha 
> What will be the default directory to install IM into ?

That is the question for another issue. May be for [this one](https://jira.codenvycorp.com/browse/CODENVY-317).

cdec/installation-manager-resources/src/main/resources/im-scripts/install-codenvy.sh
dmytro-ndp: okay

tolusha:
others ok. +1

PR #61, opened 2016-04-01
CODENVY-141: fix codenvy classic url

riuvshin:
In phase 1 codenvy classic url should point to codenvy.com

@ashumilova @benoitf @TylerJewell


benoitf:
as discussed c3 will be for phase2

PR #72, opened 2016-04-07
CODENVY-316: update reference to factory documentation in UD

akurinnoy:


ashumilova:
+1

PR #78, opened 2016-04-07
Add hosts into the /etc/hosts of Codenvy on-prem from the new line

dmytro-ndp:
@tolusha: please, review my request. on-prem
This fix will allow to avoid situation when, after installing Codenvy, there will be next content of /etc/hosts initially without EOF at the end:

> 127.0.0.1 localhost127.0.0.1 puppet


tolusha:
+1

PR #82, opened 2016-04-08
Fix workspace updating on stop by timeout

mkuznyetsov:
Perform update, and then stop workspace.
Stopping will asynchronously update workspace, so make sure there will be our attribute.


riuvshin:
OK

evoevodin:
OK

akorneta:
ok

riuvshin:
MERGE

PR #83, opened 2016-04-08
CODENVY-313: fix requesting through authenticated proxy

dmytro-ndp:


tolusha:
+1

PR #86, opened 2016-04-09
Add default properties for Codenvy 4.1.0, 4.2.0-RC1-SNAPSHOT

dmytro-ndp:
@riuvshin: please, approve this request as well as set here https://github.com/riuvshin/cdec-bundle/blob/master/pom.xml version 4.1.0 to place Codenvy 4.1.0 bundle into nightly Updater Server for testing before it deploying to production.


riuvshin:
LGTM

PR #96, opened 2016-04-13
CODENVY-363: add error messages when user reach the RAM limits

olexii4:
@ashumilova, @evoevodin

### DIFF
--- a/core/codenvy-hosted-platform-api-impl/src/main/java/com/codenvy/api/workspace/LimitsCheckingWorkspaceManager.java
+++ b/core/codenvy-hosted-platform-api-impl/src/main/java/com/codenvy/api/workspace/LimitsCheckingWorkspaceManager.java
@@ -39,6 +39,7 @@ import org.eclipse.che.commons.lang.Size;
 import javax.inject.Inject;
 import javax.inject.Named;
 import javax.inject.Singleton;
+import java.text.DecimalFormat;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -56,6 +57,7 @@ import static org.eclipse.che.api.core.model.workspace.WorkspaceStatus.STOPPED;
 @Singleton
 public class LimitsCheckingWorkspaceManager extends WorkspaceManager {
 
+    private static final DecimalFormat DECIMAL_FORMAT = new DecimalFormat("#0.#");
     private static final Striped<Lock> CREATE_LOCKS = Striped.lazyWeakLock(100);
     private static final Striped<Lock> START_LOCKS  = Striped.lazyWeakLock(100);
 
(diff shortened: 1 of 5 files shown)


core/codenvy-hosted-platform-api-impl/src/main/java/com/codenvy/api/workspace/LimitsCheckingWorkspaceManager.java
```diff
@@ -235,14 +242,40 @@ void checkMaxEnvironmentRam(WorkspaceConfig config) throws LimitExceededExceptio
...
                                                         "This value is set by your admin with the 'limits.workspace.env.ram' property",
-                                                        maxRamPerEnv,
-                                                        workspaceRam));
+                                                        parseMegabyteSizeToString(maxRamPerEnv, "mb"),
+                                                        parseMegabyteSizeToString(workspaceRam, "mb")));
             }
         }
     }
 
+    private String parseMegabyteSizeToString(long ramSizeMB, String suffix) {
```
sleshchenko: Maybe it is better to use https://github.com/eclipse/che/blob/master/core/commons/che-core-commons-lang/src/main/java/org/eclipse/che/commons/lang/Size.java#L22-22 here.

ashumilova:
+1 for dashboard

sleshchenko:
:+1:

core/codenvy-hosted-platform-api-impl/src/main/java/com/codenvy/api/workspace/LimitsCheckingWorkspaceManager.java
evoevodin: Well, about the method: i don't think this is a good place for such kind of transformations. 

1. Basically you need this `parse` method  only in the one place `checkRamAndPropagateStart`
2. You don't need to parse from megabytes to megabytes.
3. If [Size]( https://github.com/eclipse/che/blob/master/core/commons/che-core-commons-lang/src/main/java/org/eclipse/che/commons/lang/Size.java) class doesn't work appropriate for you, then you should either simplify this method to convert megabytes to gigabytes(and that's it), or add a new method into the `Size` class, which may be reused by others.

core/codenvy-hosted-platform-api-impl/src/main/java/com/codenvy/api/workspace/LimitsCheckingWorkspaceManager.java
```diff
@@ -235,14 +242,40 @@ void checkMaxEnvironmentRam(WorkspaceConfig config) throws LimitExceededExceptio
                                                  .mapToInt(machineCfg -> machineCfg.getLimits().getRam())
                                                  .sum();
             if (workspaceRam > maxRamPerEnv) {
-                throw new LimitExceededException(format("The maximum RAM per workspace is set to '%dmb' and you requested '%dmb'. " +
+                throw new LimitExceededException(format("The maximum RAM per workspace is set to '%s' and you requested '%s'. " +
                                                         "This value is set by your admin with the 'limits.workspace.env.ram' property",
-                                                        maxRamPerEnv,
-                                                        workspaceRam));
+                                                        parseMegabyteSizeToString(maxRamPerEnv, "mb"),
```
evoevodin: Can you please explain why did you change this code?
I think the previous version was pretty ok, and we don't need extra overhead here.

core/codenvy-hosted-platform-api-impl/src/main/java/com/codenvy/api/workspace/LimitsCheckingWorkspaceManager.java
```diff
@@ -169,21 +171,26 @@ public WorkspaceImpl updateWorkspace(String id, Workspace update) throws Conflic
...
-                                                                            .getEnvironment(ws.getRuntime().getActiveEnv())
-                                                                            .get()
-                                                                            .getMachineConfigs())
-                                                               .mapToLong(this::sumRam)
-                                                               .sum();
+            final List<WorkspaceImpl> workspacesPerUser = getWorkspaces(user);
+            final long runningWorkspaces = workspacesPerUser.stream().filter(ws -> RUNNING == ws.getStatus()).count();
+            final long currentlyUsedRamMB = workspacesPerUser.stream().filter(ws -> STOPPED != ws.getStatus())
+                                                             .map(ws -> ws.getConfig()
+                                                                          .getEnvironment(
```
evoevodin: change the formatting back please `.getEnvironment(ws.getRuntime().getActiveEnv())`

core/codenvy-hosted-platform-api-impl/src/main/java/com/codenvy/api/workspace/LimitsCheckingWorkspaceManager.java
```diff
@@ -169,21 +171,26 @@ public WorkspaceImpl updateWorkspace(String id, Workspace update) throws Conflic
...
-            final long currentlyUsedRamMB = getWorkspaces(user).stream()
-                                                               .filter(ws -> STOPPED != ws.getStatus())
-                                                               .map(ws -> ws.getConfig()
-                                                                            .getEnvironment(ws.getRuntime().getActiveEnv())
-                                                                            .get()
-                                                                            .getMachineConfigs())
-                                                               .mapToLong(this::sumRam)
-                                                               .sum();
+            final List<WorkspaceImpl> workspacesPerUser = getWorkspaces(user);
+            final long runningWorkspaces = workspacesPerUser.stream().filter(ws -> RUNNING == ws.getStatus()).count();
```
evoevodin: Workspace uses RAM when it in `STARTING`, `RUNNING`, `STOPPING` state so you must change the code `RUNNING  = ws.getStatus()` to `STOPPED != ws.getStatus()`

core/codenvy-hosted-platform-api-impl/src/main/java/com/codenvy/api/workspace/LimitsCheckingWorkspaceManager.java
```diff
@@ -56,6 +57,8 @@
 @Singleton
 public class LimitsCheckingWorkspaceManager extends WorkspaceManager {
 
+    public static final DecimalFormat DECIMAL_FORMAT = new DecimalFormat("#0.#");
```
evoevodin: why `public` ?

evoevodin:
+1

PR #101, opened 2016-04-14
Fix java_naming_security_principal property; add properties for 4.1.1

dmytro-ndp:


riuvshin:
OK

PR #103, opened 2016-04-14
CODENVY-368: fix merge of properties, add scripts for migration

dmytro-ndp:
@tolusha, @riuvshin: please, review this request.

1) add merge of properties in time of update Codenvy from binary
2) add script to migrate from 4.0.1 to 4.1.0
3) add scripts to prepare Codenvy bundle and upload it to update server
4) don't replace default properties in time of update Codenvy

### DIFF
--- a/cdec/installation-manager-core/src/main/java/com/codenvy/im/artifacts/helper/CDECMultiServerHelper.java
+++ b/cdec/installation-manager-core/src/main/java/com/codenvy/im/artifacts/helper/CDECMultiServerHelper.java
@@ -259,12 +259,12 @@ public class CDECMultiServerHelper extends CDECArtifactHelper {
             }
 
             case 5:
-                return createCommand("sudo systemctl start puppetmaster");
+                return createStartServiceCommand("puppetmaster");
 
             case 6:
                 List<Command> commands = new ArrayList<>();
                 commands.add(createCommand("sudo systemctl start puppet", nodeConfigs));
-                commands.add(createCommand("sudo systemctl start puppet"));
+                commands.add(createStartServiceCommand("puppet"));
 
                 return new MacroCommand(commands, "Launch puppet agent");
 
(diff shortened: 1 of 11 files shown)


cdec/installation-manager-core/src/main/java/com/codenvy/im/artifacts/helper/CDECMultiServerHelper.java
```diff
@@ -299,9 +299,13 @@ public Command getUpdateCommand(Version versionToUpdate, Path pathToBinaries, In
...
             case 0:
-                return createCommand(format("rm -rf %1$s; " +
-                                            "mkdir %1$s; " +
-                                            "unzip -o %2$s -d %1$s", getTmpCodenvyDir(), pathToBinaries.toString()));
+                return new MacroCommand(ImmutableList.of(
+                    createCommand(format("rm -rf %1$s; " +
+                                         "mkdir %1$s; " +
+                                         "unzip -o %2$s -d %1$s; ", getTmpCodenvyDir(), pathToBinaries.toString())),
+                    createCommand("sudo systemctl stop crond"),
+                    createCommand("sudo systemctl stop puppet"),
```
tolusha: Why do we need to stop puppet?

dmytro-ndp: @riuvshin suggested to do so, because then we change puppet manifest and puppet could start applying configuration for new version of Codenvy too early and too long - before the new version of Codenvy has been installed.

riuvshin: any migration must be performed before update tomcat, also we must stop current codenvy server to protect DB from unexpected changes, so we need to stop puppet because it will restart codenvy while we running migrations.

tolusha:
+1

riuvshin:
+1

PR #104, opened 2016-04-15
Skip importing project without sources in factory

mkuznyetsov:


skabashnyuk:
ok

plugins/plugin-hosted/codenvy-hosted-ext-factory/src/main/java/com/codenvy/ide/factory/client/utils/FactoryProjectImporter.java
```diff
@@ -132,6 +132,10 @@ public void apply(List<ProjectConfigDto> projectConfigs) throws OperationExcepti
                 Set<String> projectNames = new HashSet<>();
                 String createPolicy = factory.getPolicies() != null ? factory.getPolicies().getCreate() : null;
                 for (ProjectConfigDto projectConfig : projectConfigs) {
+                    if (projectConfig.getSource().getLocation() == null) {
+                        continue;
```
sleshchenko: mb we should show warning message that we can not import project because it doesn't contain source location?

plugins/plugin-hosted/codenvy-hosted-ext-factory/src/main/java/com/codenvy/ide/factory/client/utils/FactoryProjectImporter.java
```diff
@@ -132,6 +132,10 @@ public void apply(List<ProjectConfigDto> projectConfigs) throws OperationExcepti
                 Set<String> projectNames = new HashSet<>();
                 String createPolicy = factory.getPolicies() != null ? factory.getPolicies().getCreate() : null;
                 for (ProjectConfigDto projectConfig : projectConfigs) {
+                    if (projectConfig.getSource().getLocation() == null) {
```
mshaposhnik: projectConfig.getSource() can't be null? For me seems it can.

mshaposhnik:
Other ok

plugins/plugin-hosted/codenvy-hosted-ext-factory/src/main/java/com/codenvy/ide/factory/client/utils/FactoryProjectImporter.java
sleshchenko: As I understand For It can't because source is mandatory parameter in factory https://github.com/eclipse/che/blob/CODENVY-365/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/shared/dto/ProjectConfigDto.java#L80-80

But mb it is better to make source optional when project is subproject. Instead of making optional type and location. WDYT?

plugins/plugin-hosted/codenvy-hosted-ext-factory/src/main/java/com/codenvy/ide/factory/client/utils/FactoryProjectImporter.java
sleshchenko: I understood that it is possible only if project is subproject and it is not need to import it. Please add this comment here

plugins/plugin-hosted/codenvy-hosted-ext-factory/src/main/java/com/codenvy/ide/factory/client/utils/FactoryProjectImporter.java
mshaposhnik: Yes, sure (i thought it was) This https://github.com/eclipse/che/pull/1056/files#diff-fc34b5aa12e2c9b6e3b005a8987613ccR603 line confused me

mkuznyetsov: fixed

PR #109, opened 2016-04-18
CHE-362. Changes according to adding display modes in  StatusNotification

RomanNikitenko:
Signed-off-by: Roman Nikitenko <rnikitenko@codenvy.com>


vparfonov:
ok

PR #110, opened 2016-04-18
Fix PullRequest Plugin according chnages in GitServiceClient

vparfonov:
Signed-off-by: Vitaly Parfonov <vparfonov@codenvy.com>


akorneta:
ok

PR #115, opened 2016-04-19
CHE-243: Return link header describing page links

evoevodin:
Add page links for gettings users.
@skabashnyuk please review

### DIFF
--- a/core/codenvy-hosted-api-admin/src/main/java/com/codenvy/api/dao/ldap/AdminUserDaoImpl.java
+++ b/core/codenvy-hosted-api-admin/src/main/java/com/codenvy/api/dao/ldap/AdminUserDaoImpl.java
@@ -18,6 +18,7 @@ import com.codenvy.api.user.server.dao.AdminUserDao;
 import com.google.inject.Inject;
 
 import org.eclipse.che.api.account.server.dao.AccountDao;
+import org.eclipse.che.api.core.Page;
 import org.eclipse.che.api.core.ServerException;
 import org.eclipse.che.api.core.notification.EventService;
 import org.eclipse.che.api.user.server.dao.PreferenceDao;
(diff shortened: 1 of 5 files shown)


core/codenvy-hosted-api-admin/src/main/java/com/codenvy/api/dao/ldap/AdminUserDaoImpl.java
```diff
@@ -64,19 +74,62 @@ public AdminUserDaoImpl(AccountDao accountDao,
...
+
+    // TODO: consider WeakReference holder for returned value.
+    // as this operation may be expensive on large medium/large data sets
+    private int getTotalUsersCount() throws ServerException {
+        InitialLdapContext context = null;
+        try {
+            context = contextFactory.createContext();
+            final SearchControls controls = new SearchControls();
+            controls.setSearchScope(SearchControls.SUBTREE_SCOPE);
+            controls.setReturningAttributes(EMPTY_ARRAY);
```
skabashnyuk: is this call mandatory?

evoevodin: yes, there is no way to count elements using ldap filters, the only way is to count elements in enumeration, so it is much cheaper to return N matched elements which are empty and count them.

skabashnyuk: When add plz this information to EMPTY_ARRAY constant.

evoevodin: I will add this information to method javadoc, thank you

skabashnyuk:
ок

PR #116, opened 2016-04-20
CODENVY-370: dashboard for OnPrem must not include the button "Codenv…

akurinnoy:
…y Classic"


benoitf:
I think this is not what I've discussed on JIRA entry BTW
but ok as quickfix

benoitf:
based on the issue we only need a way to customize the display of the top left button, not to know the assembly. As this custom way can be used by several assemblies

ashumilova:
+1

benoitf:
+1

PR #119, opened 2016-04-20
Revert "CHE-243: Return link header describing page links"

evoevodin:
Reverts codenvy/codenvy#115


evoevodin:
Must be merged into 4.3

PR #125, opened 2016-04-21
CODENVY-423: fixed swarm nodes info parsing

garagatyi:
Also require new property in puppet swarm.client.node_description_length

@riuvshin @skabashnyuk Please review


riuvshin:
tested! +1

PR #130, opened 2016-04-22
CODENVY-428 Improve workspace activity notification

mkuznyetsov:
Activity notification will be send as soon as activity is occured,
but not more than once in a specific amount of time (60 seconds).


garagatyi:
tests?

skabashnyuk:
ок

PR #141, opened 2016-04-25
CODENVY-434 Adapt Che URL and port config

stour:
Fix for platform-api-client-java integration tests.

@skabashnyuk @benoitf

### DIFF
--- a/platform-api-client-java/core/pom.xml
+++ b/platform-api-client-java/core/pom.xml
@@ -140,16 +140,12 @@
                                 </goals>
                                 <configuration>
                                     <target>
-                                        <delete dir="${codenvy.ide.path}/temp" />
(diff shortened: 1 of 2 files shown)


skabashnyuk:
ok

benoitf:
I think that jmx server port can still collide, no ? the same for the shutdown port
If there is an existing tomcat server (like jenkins) I think it may kill jenkins instead of the integration tests 'internal tomcat'

stour:
@benoitf I updated the PR in order to replace port 8005 and 8009 in server.xml and so avoid collision same as before.

benoitf:
+1

stour:
@skabashnyuk @benoitf could one of you please merge ?

benoitf:
it seems I don't have the right to merge

PR #163, opened 2016-05-04
CODENVY-433 Implement permissions based authorization for Recipe and Stack APIs

sleshchenko:
@gazarenkov @skabashnyuk @evoevodin Please review my changes


core/codenvy-hosted-api-permission/src/main/java/com/codenvy/api/permission/server/dao/PermissionsStorage.java
```diff
@@ -95,7 +75,7 @@
      * @throws ServerException
      *         when any other error occurs during permissions fetching
      */
-    List<PermissionsImpl> getByInstance(String domain, String instance) throws ServerException;
+    List<PermissionsImpl> getByInstance(String domain, String instance) throws ServerException, NotFoundException;
```
evoevodin: Document `NotFoundException`

core/codenvy-hosted-api-workspace/src/main/java/com/codenvy/api/workspace/server/stack/StackAction.java
```diff
@@ -0,0 +1,37 @@
...
+ * Dissemination of this information or reproduction of this material
+ * is strictly forbidden unless prior written permission is obtained
+ * from Codenvy S.A..
+ */
+package com.codenvy.api.workspace.server.stack;
+
+/**
+ * @author Sergii Leschenko
+ */
+public enum StackAction {
```
evoevodin: `RecipeAction` is exactly the same as this enum, so maybe it makes sense to have one enum describing actions ?

core/codenvy-hosted-api-workspace/src/main/java/com/codenvy/api/workspace/server/stack/StackPermissionStorage.java
```diff
@@ -0,0 +1,135 @@
...
+    private final StackDao stackDao;
+
+    @Inject
+    public StackPermissionStorage(StackDao stackDao) {
+        this.stackDao = stackDao;
+    }
+
+    @Override
+    public Set<PermissionsDomain> getDomains() {
+        return ImmutableSet.of(new StackDomain());
```
evoevodin: What about extracting set to the field?

core/codenvy-hosted-api-workspace/src/main/java/com/codenvy/api/workspace/server/stack/StackPermissionStorage.java
```diff
@@ -0,0 +1,135 @@
...
+        } catch (NotFoundException e) {
+            return false;
+        }
+
+        final Set<String> allowedActions = recipe.getAcl()
+                                                 .stream()
+                                                 .filter(entry -> entry.getUser().equals(user) || entry.getUser().equals("*"))
+                                                 .flatMap(entry -> entry.getActions().stream())
+                                                 .collect(Collectors.toSet());
+        return allowedActions.contains(action);
```
evoevodin: I'd say that `anyMatch` is better, as you don't need to allocate memory for `Set` + it is cheaper in the case of time.
```java
return recipe.getAcl()
             .stream()
             .filter(entry -> entry.getUser().equals(user) || entry.getUser().equals("*")) 
             .flatMap(entry -> entry.getActions().stream())
             .anyMatch(action::equals);
```

sleshchenko: +1 thx

core/codenvy-hosted-api-workspace/src/main/java/com/codenvy/api/workspace/server/stack/StackPermissionsFilter.java
```diff
@@ -0,0 +1,82 @@
...
+ * Restricts access to methods of {@link StackService} by users' permissions
+ *
+ * <p>Filter should contain rules for protecting of all methods of {@link StackService}.<br>
+ * In case when requested method is unknown filter throws {@link ForbiddenException}
+ *
+ * @author Sergii Leschenko
+ */
+@Filter
+@Path("/stack{path:(/.*)?}")
+public class StackPermissionsFilter extends CheMethodInvokerFilter {
```
evoevodin: Do you have any tests related to the interceptor compatibility ?
I mean will it fail if i change a method signature ?

sleshchenko: I'm going to add them. Something like https://github.com/codenvy/codenvy/blob/master/core/codenvy-hosted-api-workspace/src/test/java/com/codenvy/api/workspace/server/filters/WorkspacePermissionsFilterTest.java#L70-70

core/codenvy-hosted-platform-api-impl/src/main/java/com/codenvy/api/dao/mongo/DocumentedCodec.java
```diff
@@ -0,0 +1,52 @@
...
+import org.bson.codecs.DecoderContext;
+import org.bson.codecs.EncoderContext;
+import org.bson.codecs.configuration.CodecRegistry;
+
+/**
+ * Implementations of this abstract class can both encode and decode values of type {@code T} by {@link Document} instance.
+ *
+ * @author Sergii Leschenko
+ */
+public abstract class DocumentedCodec<T> implements Codec<T> {
```
evoevodin: I don't think that this name is really good as `DocumentCodec` already exists and often used so this name may be confusing. Consider `AbstractDocumentCodec` or another name for this class.

core/codenvy-hosted-api-permission/src/main/java/com/codenvy/api/permission/server/dao/PermissionsStorage.java
sleshchenko: documented

core/codenvy-hosted-api-workspace/src/main/java/com/codenvy/api/workspace/server/stack/StackPermissionsFilter.java
sleshchenko: Added here https://github.com/codenvy/codenvy/blob/acls/core/codenvy-hosted-api-workspace/src/test/java/com/codenvy/api/workspace/server/stack/StackPermissionsFilterTest.java

skabashnyuk:
ok

evoevodin:
ok

PR #166, opened 2016-05-05
Fix eclipse jdt dependency, codenvy repacked one should be used

mshaposhnik:


mkuznyetsov:
ok

skabashnyuk:
ok. @vparfonov ?

vparfonov:
ok

PR #167, opened 2016-05-05
CODENVY-426 show dedicated message when clone fails due to lack of SSH key

mshaposhnik:


TylerJewell:
@riuvshin If we accept this request, we should cherry pick this issue and codenvy-455 into a hot fix release.  Maybe combined with the swarm race condition fix.

mshaposhnik:
@skabashnyuk  @vparfonov pls check

plugins/plugin-bitbucket/codenvy-plugin-bitbucket-ext-bitbucket/src/main/java/org/eclipse/che/ide/ext/bitbucket/server/BitbucketKeyUploader.java
```diff
@@ -1,3 +1,4 @@
+
```
sleshchenko: remove this line

skabashnyuk:
ок

PR #169, opened 2016-05-06
Codenvy 473: fix Codenvy bootstrap script for TIAA CREF

dmytro-ndp:
1. Fix proxy errors.
2. Fix bad formatting while running on a normal terminal.
3. Don't interrupt installation on checking puppet master port if there is puppet master installed in system earlier.
4. Change "--silent" option to be named "--suppress".
5. Implement "--silent" option to mean that we only display limited outputs when doing the installation.


TylerJewell:
+1

dmytro-ndp:
@tolusha: could you, please, take a look on this request as well?

PR #193, opened 2016-05-13
CODENVY-292 Escape factory DB query attributes

mkuznyetsov:
Sometimes query attribute values can contain dots ("creator.userId" : "admin@codenvy.onprem").
It must be escaped, as it is stored in DB.


skabashnyuk:
ок

evoevodin:
@mshaposhnik approve this please

mshaposhnik:
affirmative

PR #194, opened 2016-05-13
Codenvy 473: fix IM according to TIAA CREF requests

dmytro-ndp:
Issues:
1) add '--list-remote' option of 'im-download' command to display all available versions in reverse order
2) fix setting proxy settings in codenvy properties when there is no http(s)_proxy environment variables.

### DIFF
--- a/cdec/installation-manager-core/src/main/java/com/codenvy/im/facade/IMArtifactLabeledFacade.java
+++ b/cdec/installation-manager-core/src/main/java/com/codenvy/im/facade/IMArtifactLabeledFacade.java
@@ -120,8 +120,8 @@ public class IMArtifactLabeledFacade extends InstallationManagerFacade {
 
     /** {@inheritDoc} */
     @Override
-    public List<UpdateArtifactInfo> getAllUpdates(@Nullable Artifact artifact) throws IOException, JsonParseException {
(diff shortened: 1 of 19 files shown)


cdec/installation-manager-core/src/main/java/com/codenvy/im/facade/IMArtifactLabeledFacade.java
```diff
@@ -120,8 +120,8 @@ public IMArtifactLabeledFacade(@Named("installation-manager.download_dir") Strin
 
     /** {@inheritDoc} */
     @Override
-    public List<UpdateArtifactInfo> getAllUpdates(@Nullable Artifact artifact) throws IOException, JsonParseException {
-        List<UpdateArtifactInfo> updates = super.getAllUpdates(artifact);
+    public List<UpdateArtifactInfo> getAllUpdates(@Nullable Artifact artifact, boolean fromInstalledVersion) throws IOException, JsonParseException {
```
tolusha: I would like to have 2 methods:
* getAllUpdates()
* getUpdatesAfterVersion()
It is more readable.

dmytro-ndp: agree, but it is no "afterVersion" but "afterInstalledVersrion"

PR #195, opened 2016-05-13
CODENVY-502 allow to create factories by providing github URL

benoitf:
Change-Id: Ic5707936dfb9682126f14a2ed7b1716becde31cd

### DIFF
--- a/assembly/onpremises-ide-packaging-war-factory/src/main/java/com/codenvy/onpremises/factory/filter/FactoryParamsFilter.java
+++ b/assembly/onpremises-ide-packaging-war-factory/src/main/java/com/codenvy/onpremises/factory/filter/FactoryParamsFilter.java
@@ -25,6 +25,7 @@ import javax.servlet.ServletException;
 import javax.servlet.ServletRequest;
 import javax.servlet.ServletResponse;
 import java.io.IOException;
+import java.util.Iterator;
 import java.util.Map;
 
 import static org.eclipse.che.api.factory.server.FactoryConstants.INVALID_PARAMETER_MESSAGE;
@@ -65,7 +66,7 @@ public class FactoryParamsFilter implements Filter {
                                     String.format(PARAMETRIZED_ILLEGAL_PARAMETER_VALUE_MESSAGE, "id", params.get("id")[1]));
                 return;
             }
-        } else if (params.get("name") != null && params.get("user") != null) {
+        } else if (params.get("name") != null || params.get("user") != null) {
             if (params.size() != 2) {
                 dispatchToErrorPage(req, resp, INVALID_FACTORY_URL_PAGE, INVALID_PARAMETER_MESSAGE);
                 return;
@@ -83,8 +84,21 @@ public class FactoryParamsFilter implements Filter {
             }
 
         } else {
-            dispatchToErrorPage(req, resp, INVALID_FACTORY_URL_PAGE, INVALID_PARAMETER_MESSAGE);
-            return;
+
+            // check that we have unique value for each parameters
+            // custom parameters will be checked at the next step when trying to retrieve it
+            Iterator<Map.Entry<String, String[]>> iterator = params.entrySet().iterator();
+            while (iterator.hasNext()) {
+                Map.Entry<String, String[]> element = iterator.next();
+                String key = element.getKey();
+                String[] array = element.getValue();
+                // check it's only one item length
+                if (array.length != 1) {
+                    dispatchToErrorPage(req, resp, INVALID_FACTORY_URL_PAGE,
+                                        String.format(PARAMETRIZED_ILLEGAL_PARAMETER_VALUE_MESSAGE, key, params.get(key)[1]));
+                }
+            }
+
         }
         filterChain.doFilter(req, resp);
     }
--- a/assembly/onpremises-ide-packaging-war-factory/src/main/java/com/codenvy/onpremises/factory/filter/FactoryRetrieverFilter.java
+++ b/assembly/onpremises-ide-packaging-war-factory/src/main/java/com/codenvy/onpremises/factory/filter/FactoryRetrieverFilter.java
@@ -32,10 +32,13 @@ import javax.servlet.ServletException;
 import javax.servlet.ServletRequest;
 import javax.servlet.ServletResponse;
 import javax.servlet.http.HttpServletRequest;
-
 import java.io.IOException;
+import java.util.Enumeration;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 
+import static javax.ws.rs.HttpMethod.POST;
 import static javax.ws.rs.core.UriBuilder.fromUri;
 
 /**
@@ -87,7 +90,8 @@ public class FactoryRetrieverFilter implements Filter {
                                                      .setMethod("GET")
                                                      .request()
                                                      .asDto(Factory.class);
-            } else {
+            } else if (httpReq.getParameter("user") != null && httpReq.getParameter("name") != null) {
+
                 final String getFactoryUrl = fromUri(apiEndPoint).path(FactoryService.class)
                                                                  .path(FactoryService.class, "getFactoryByAttribute")
                                                                  .build().toString();
@@ -109,6 +113,27 @@ public class FactoryRetrieverFilter implements Filter {
                     return;
                 } 
                 requestedFactory = matchedFactories.get(0);
+            } else {
+                // asked for a parameters factory
+
+                // first populate map of parameters
+                Map<String, String> map = new HashMap<>();
+                Enumeration<String> parameterNames = httpReq.getParameterNames();
+                while (parameterNames.hasMoreElements()) {
+                    String parameterName  = parameterNames.nextElement();
+                    map.put(parameterName, httpReq.getParameter(parameterName));
+                }
+
+                // Create URL
+                final String resolveFactoryUrl = fromUri(apiEndPoint).path(FactoryService.class)
+                                                                 .path(FactoryService.class, "resolveFactory")
+                                                                 .build().toString();
+                // perform call
+                requestedFactory = httpRequestFactory.fromUrl(resolveFactoryUrl)
+                                                     .setMethod(POST)
+                                                     .setBody(map)
+                                                     .request()
+                                                     .asDto(Factory.class);
             }
 
             req.setAttribute("factory", requestedFactory);
--- a/plugins/plugin-github/codenvy-plugin-github-factory-resolver/src/main/java/com/codenvy/plugin/factory/github/factory/resolver/GithubFactoryParametersResolver.java
+++ b/plugins/plugin-github/codenvy-plugin-github-factory-resolver/src/main/java/com/codenvy/plugin/factory/github/factory/resolver/GithubFactoryParametersResolver.java
@@ -1,0 +1,177 @@
+/*
+ *  [2012] - [2016] Codenvy, S.A.
+ *  All Rights Reserved.
+ *
+ * NOTICE:  All information contained herein is, and remains
+ * the property of Codenvy S.A. and its suppliers,
+ * if any.  The intellectual and technical concepts contained
+ * herein are proprietary to Codenvy S.A.
+ * and its suppliers and may be covered by U.S. and Foreign Patents,
+ * patents in process, and are protected by trade secret or copyright law.
+ * Dissemination of this information or reproduction of this material
+ * is strictly forbidden unless prior written permission is obtained
+ * from Codenvy S.A..
+ */
+package com.codenvy.plugin.factory.github.factory.resolver;
+
+import org.eclipse.che.api.core.BadRequestException;
+import org.eclipse.che.api.factory.server.FactoryParametersResolver;
+import org.eclipse.che.api.factory.shared.dto.Factory;
+import org.eclipse.che.api.machine.shared.dto.LimitsDto;
+import org.eclipse.che.api.machine.shared.dto.MachineConfigDto;
+import org.eclipse.che.api.machine.shared.dto.MachineSourceDto;
+import org.eclipse.che.api.workspace.shared.dto.EnvironmentDto;
+import org.eclipse.che.api.workspace.shared.dto.ProjectConfigDto;
+import org.eclipse.che.api.workspace.shared.dto.SourceStorageDto;
+import org.eclipse.che.api.workspace.shared.dto.WorkspaceConfigDto;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.validation.constraints.NotNull;
+import java.io.IOException;
+import java.net.HttpURLConnection;
+import java.net.URL;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import static java.util.Collections.singletonList;
+import static org.eclipse.che.dto.server.DtoFactory.newDto;
+
+/**
+ * Provides Factory Parameters resolver for github repositories.
+ *
+ * @author Florent Benoit
+ */
+public class GithubFactoryParametersResolver implements FactoryParametersResolver {
+
+    /**
+     * Logger.
+     */
+    private static final Logger LOG = LoggerFactory.getLogger(GithubFactoryParametersResolver.class);
+
+    /**
+     * Parameter name.
+     */
+    protected static final String URL_PARAMETER_NAME = "url";
+
+    /**
+     * Connection timeout of 10seconds.
+     */
+    private static final int CONNECTION_TIMEOUT = (int) TimeUnit.SECONDS.toMillis(10);
+
+    /**
+     * Regexp to find repository details (repository name, project name and branch and subfolder)
+     * Examples of valid URLs are in the test class.
+     */
+    protected static final Pattern GITHUB_PATTERN = Pattern.compile("^(?:http)(?:s)?(?:\\:\\/\\/)github.com/(?<repoUser>[^/]++)/(?<repoName>[^/]++)(?:/tree/(?<branchName>[^/]++)(?:/(?<subFolder>.*))?)?$");
+
+    /**
+     * Check if this resolver can be used with the given parameters.
+     *
+     * @param factoryParameters
+     *         map of parameters dedicated to factories
+     * @return true if it will be accepted by the resolver implementation or false if it is not accepted
+     */
+    @Override
+    public boolean accept(@NotNull Map<String, String> factoryParameters) {
+        // Check if url parameter is a github URL
+        return factoryParameters.containsKey(URL_PARAMETER_NAME) && GITHUB_PATTERN.matcher(factoryParameters.get(URL_PARAMETER_NAME)).matches();
+    }
+
+    /**
+     * Create factory object based on provided parameters
+     *
+     * @param factoryParameters
+     *         map containing factory data parameters provided through URL
+     * @throws BadRequestException when data are invalid
+     */
+    @Override
+    public Factory createFactory(@NotNull Map<String, String> factoryParameters) throws BadRequestException {
+
+        // no need to check null value of url parameter as accept() method has performed the check
+        String githubUrl = factoryParameters.get("url");
+
+        // Apply github url to the regexp
+        Matcher matcher = GITHUB_PATTERN.matcher(githubUrl);
+        if (!matcher.matches()) {
+            throw new BadRequestException(String.format(
+                    "The given github url %s is not a valid URL github url. It should start with https://github.com/<user>/<repo>",
+                    githubUrl));
+        }
+
+        String repoUser = matcher.group("repoUser");
+        String repoName = matcher.group("repoName");
+        String branchName = matcher.group("branchName");
+        String subfolder = matcher.group("subFolder");
+
+        // build factory for the given github url
+        Map<String, String> parameters = new HashMap<>();
+        if (branchName == null) {
+            // use master if not specified
+            branchName = "master";
+        } else {
+            parameters.put("branch", branchName);
+        }
+        if (subfolder != null) {
+            parameters.put("keepDir", subfolder);
+        }
+
+        // github location for the clone
+        String githubLocation = "https://github.com/" + repoUser + "/" + repoName;
+
+        // FIXME: use it configurable and move it to
+        final String DEFAULT_DOCKER_FILE_LOCATION = "https://dockerfiles.codenvycorp.com/templates-4.0/factory/factory-dockerfile";
+
+        // check if repository contains a codenvy Dockerfile
+        String dockerFileLocation = null;
+        String githubDockerFileLocation = "https://raw.githubusercontent.com/" + repoUser + "/" + repoName + "/" +branchName + "/.codenvy.dockerfile";
+        HttpURLConnection httpURLConnection = null;
+        try {
+            httpURLConnection = (HttpURLConnection)new URL(githubDockerFileLocation).openConnection();
+            httpURLConnection.setConnectTimeout(CONNECTION_TIMEOUT);
+            final int responseCode = httpURLConnection.getResponseCode();
+            if (responseCode == HttpURLConnection.HTTP_OK) {
+                dockerFileLocation = githubDockerFileLocation;
+            }
+        } catch (IOException ioe) {
+            LOG.debug("Unable to check if remote location {0} is available or not", githubDockerFileLocation);
+        } finally {
+            if (httpURLConnection != null) {
+                httpURLConnection.disconnect();
+            }
+        }
+
+        // apply default location if there is none
+        if (dockerFileLocation == null) {
+            dockerFileLocation = DEFAULT_DOCKER_FILE_LOCATION;
+        }
+
+        String envName = repoUser;
+
+        // Create factory object
+        SourceStorageDto
+                sourceStorageDto = newDto(SourceStorageDto.class).withLocation(githubLocation).withType("git").withParameters(parameters);
+        ProjectConfigDto projectConfig = newDto(ProjectConfigDto.class).withSource(sourceStorageDto).withName(repoName).withType("blank")
+                                                                       .withPath("/".concat(repoName));
+
+        LimitsDto limitsDto = newDto(LimitsDto.class).withRam(2000);
+        MachineSourceDto machineSourceDto = newDto(MachineSourceDto.class).withType("dockerfile").withLocation(dockerFileLocation);
+        MachineConfigDto
+                machineConfigDto =
+                newDto(MachineConfigDto.class).withLimits(limitsDto).withType("docker").withSource(machineSourceDto).withDev(true)
+                                              .withName("ws-machine");
+
+        EnvironmentDto environmentDto = newDto(EnvironmentDto.class).withName(envName).withMachineConfigs(singletonList(machineConfigDto));
+
+        final WorkspaceConfigDto
+                workspaceConfig = newDto(WorkspaceConfigDto.class).withDefaultEnv(envName).withProjects(singletonList(projectConfig))
+                                                                  .withEnvironments(singletonList(environmentDto)).withName(repoUser);
+        final Factory factory = newDto(Factory.class).withWorkspace(workspaceConfig).withV("4.0");
+
+        return factory;
+
+    }
+}
--- a/assembly/onpremises-ide-packaging-war-factory/src/main/java/com/codenvy/onpremises/factory/FactoryServlet.java
+++ b/assembly/onpremises-ide-packaging-war-factory/src/main/java/com/codenvy/onpremises/factory/FactoryServlet.java
@@ -30,6 +30,8 @@ import javax.servlet.http.HttpServletResponse;
 import javax.ws.rs.core.UriBuilder;
 import java.io.IOException;
 import java.net.URLDecoder;
+import java.net.URLEncoder;
+import java.util.Enumeration;
 
 import static com.google.common.base.Strings.nullToEmpty;
 import static javax.ws.rs.core.UriBuilder.fromUri;
@@ -79,8 +81,31 @@ public class FactoryServlet extends HttpServlet {
                  "", //TODO: use from policy
                  "" //TODO: use from policy
                 );
-        UriBuilder redirectAfterStateOnline = UriBuilder.fromPath("/dashboard/").fragment("load-factory/" + requestedFactory.getId());
-        resp.sendRedirect(redirectAfterStateOnline.build().toString());
+
+        String redirectUrl = UriBuilder.fromPath("/dashboard/").fragment("load-factory/").build().toString();
+
+        boolean firstParam = true;
+        Enumeration<String> parameterNames = req.getParameterNames();
+
+        // send to dashboard all parameters provided in the request
+        if (parameterNames != null) {
+            while (parameterNames.hasMoreElements()) {
+                String key = parameterNames.nextElement();
+                String value = req.getParameter(key);
+                if (value != null) {
+                    if (firstParam) {
+                        redirectUrl = redirectUrl.concat("?");
+                        firstParam = false;
+                    } else {
+                        redirectUrl = redirectUrl.concat("&");
+                    }
+                    // encode parameter
+                    redirectUrl = redirectUrl.concat(key).concat("=").concat(URLEncoder.encode(value, "UTF-8"));
+                }
+            }
+        }
+
+        resp.sendRedirect(redirectUrl);
     }
 
 
--- a/assembly/onpremises-ide-packaging-war-factory/src/test/java/com/codenvy/onpremises/factory/FactoryParamsFilterTest.java
+++ b/assembly/onpremises-ide-packaging-war-factory/src/test/java/com/codenvy/onpremises/factory/FactoryParamsFilterTest.java
@@ -87,7 +87,7 @@ public class FactoryParamsFilterTest {
     }
 
     @Test
-    public void shouldThrowExceptionIfNoIdParamPresentInFactory() throws Exception {
+    public void shouldNotThrowExceptionIfNoIdParamPresentInFactory() throws Exception {
         Map<String, String[]> parametersMap = new HashMap<>();
         parametersMap.put("any", new String[] {"anything"});
         when(req.getParameterMap()).thenReturn(parametersMap);
@@ -95,9 +95,9 @@ public class FactoryParamsFilterTest {
 
         filter.doFilter(req, res, chain);
 
-        verify(req).setAttribute(eq(RequestDispatcher.ERROR_MESSAGE), eq(FactoryConstants.INVALID_PARAMETER_MESSAGE));
-        verify(requestDispatcher).forward(req, res);
-        verify(chain, never()).doFilter(any(ServletRequest.class), any(ServletResponse.class));
+        verify(req, never()).setAttribute(eq(RequestDispatcher.ERROR_MESSAGE), eq(FactoryConstants.INVALID_PARAMETER_MESSAGE));
+        verify(requestDispatcher, never()).forward(req, res);
+        verify(chain).doFilter(any(ServletRequest.class), any(ServletResponse.class));
     }
 
     @Test
--- a/assembly/onpremises-ide-packaging-war-factory/src/test/java/com/codenvy/onpremises/factory/FactoryServletTest.java
+++ b/assembly/onpremises-ide-packaging-war-factory/src/test/java/com/codenvy/onpremises/factory/FactoryServletTest.java
@@ -36,8 +36,10 @@ import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import javax.ws.rs.ext.RuntimeDelegate;
 import java.security.Principal;
-import java.util.Collections;
 
+import static java.util.Arrays.asList;
+import static java.util.Collections.enumeration;
+import static java.util.Collections.singletonList;
 import static org.eclipse.che.api.core.rest.HttpJsonHelper.HttpJsonHelperImpl;
 import static org.eclipse.che.dto.server.DtoFactory.newDto;
 import static org.mockito.Matchers.contains;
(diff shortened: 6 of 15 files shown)


benoitf:
linked to https://github.com/eclipse/che/pull/1247

benoitf:
@skabashnyuk @ashumilova

benoitf:
FYI, in the JIRA issue there is a video

benoitf:
Pull request rewritten
you can check new feature diagram in the following image
image

plugins/plugin-factory-github/pom.xml
```diff
@@ -0,0 +1,32 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
skabashnyuk: plugins/plugin-factory-github/pom.xml
What if move this to plugin-github
plugins/plugin-github/plugin-github-factory/pom.xml

benoitf: I thought about that but I was not sure as it's linked in some way to github but not using the current stuff in this directory
But I'm ok to change

assembly/onpremises-ide-packaging-war-factory/src/main/java/com/codenvy/onpremises/factory/filter/FactoryParamsFilter.java
```diff
@@ -83,8 +97,38 @@ public void doFilter(ServletRequest req, ServletResponse resp, FilterChain filte
...
+                    // add to the single value map
+                    parameters.put(key, array[0]);
+                }
+
+                // now loop on resolvers and see if one is accepting it
+                Optional<FactoryParametersResolver> factoryParametersResolverOptional =
+                        this.resolvers.stream().filter((resolver -> resolver.accept(parameters))).findFirst();
+
+                // matching, it's ok
+                if (factoryParametersResolverOptional.isPresent()) {
```
mshaposhnik: ` if (!factoryParametersResolverOptional.isPresent()) {
                   dispatchToErrorPage(req, resp, INVALID_FACTORY_URL_PAGE, INVALID_PARAMETER_MESSAGE);
                return;
                }`
and avoid parameterIsOk variable

benoitf: ok

assembly/onpremises-ide-packaging-war-factory/src/main/java/com/codenvy/onpremises/factory/filter/FactoryRetrieverFilter.java
```diff
@@ -109,6 +114,28 @@ public void doFilter(ServletRequest req, ServletResponse resp, FilterChain filte
...
+                final String getFactoryUrl = fromUri(apiEndPoint).path(ParametersFactoryService.class)
+                                                                 .path(ParametersFactoryService.class, "getFactory")
+                                                                 .build().toString();
+                // perform call
+                requestedFactory = httpRequestFactory.fromUrl(getFactoryUrl)
+                                                     .setMethod(POST)
+                                                     .setBody(map)
+                                                     .request()
+                                                     .asDto(Factory.class);
+
```
mshaposhnik: And why we didn't check result of the call? if it is 500? We should do  dispatchToErrorPage here as well imho.
Also, for me looks like [this](https://github.com/codenvy/codenvy/blob/71d1b5b03f7668ac7df34926b733d08eefee4d33/assembly/onpremises-ide-packaging-war-factory/src/main/java/com/codenvy/onpremises/factory/filter/FactoryParamsFilter.java#L100-L131) check can be avoided by catching correct answer from Factory service here

plugins/plugin-factory-github/codenvy-plugin-factory-github-server/src/main/java/com/codenvy/plugin/factory/github/server/GithubFactoryParametersResolver.java
```diff
@@ -0,0 +1,194 @@
...
+                                              .withName("ws-machine");
+
+        EnvironmentDto environmentDto = newDto(EnvironmentDto.class).withName(envName).withMachineConfigs(singletonList(machineConfigDto));
+
+        final WorkspaceConfigDto
+                workspaceConfig = newDto(WorkspaceConfigDto.class).withDefaultEnv(envName).withProjects(singletonList(projectConfig))
+                                                                  .withEnvironments(singletonList(environmentDto)).withName(repoUser);
+        final Factory factory = newDto(Factory.class).withWorkspace(workspaceConfig).withV("4.0");
+
+        return factory;
```
mshaposhnik: Such way of constructing factory looks kinda sad for me = we strongly rely in it's model, and every change should be adopted manually. Maybe we can use something like simplified version of [this](https://github.com/codenvy/che-core/commit/a9196cc62d54928a6c2e3dbc673cec382c9bbf91#diff-d35d81a76138d9eb1ddef7605359129fL337)

assembly/onpremises-ide-packaging-war-factory/src/main/java/com/codenvy/onpremises/factory/filter/FactoryRetrieverFilter.java
benoitf: if there is 500 we go in catch block as there is an exception thrown

benoitf: about the check it's you that decided to perform the check before trying to get the factory.
so I'm just reusing the same approach you defined. Check first and then call factory service after

plugins/plugin-factory-github/codenvy-plugin-factory-github-server/src/main/java/com/codenvy/plugin/factory/github/server/GithubFactoryParametersResolver.java
benoitf: but AFAIK there is backward compliance in the model as we have a version in the Factory object (v 4.0, etc) so there shouldn't have any manual changes to do.
So it means that if attributes are changed, version number is bumped. but the DTO is still valid.
As the factory.json file can be stored on client side or on repository it should always be valid.


The only point I see is introducing a default factory builder to create factory with default parameters and only keep github specific part there

mshaposhnik: I mean, if something will be changed inside the WorkspaceConfigDto for example (which is not related to factory, we just importing it) we will need to go and fix here manulally.

benoitf: but if it is imported it means it needs to be compliant as well.

assembly/onpremises-ide-packaging-war-factory/src/main/java/com/codenvy/onpremises/factory/deploy/FactoryServletModule.java
```diff
@@ -39,6 +41,11 @@ protected void configureServlets() {
         filterRegex(PASS_RESOURCES_REGEXP).through(com.codenvy.onpremises.factory.filter.ReferrerCheckerFilter.class);
 
 
+        Multibinder<FactoryParametersResolver> githubFactoryParametersResolverMultibinder =
```
garagatyi: We do not use servlet modules to bind components unless it is servlet/filter or other module that provides servlets/filters

plugins/plugin-factory-github/codenvy-plugin-factory-github-server/src/main/java/com/codenvy/plugin/factory/github/server/GithubFactoryParametersResolver.java
```diff
@@ -0,0 +1,194 @@
...
+
+    /**
+     * Connection timeout of 30seconds.
+     */
+    private static final int CONNECTION_TIMEOUT = (int) TimeUnit.SECONDS.toMillis(30);
+
+    /**
+     * Regexp to find repository details (repository name, project name and branch and subfolder)
+     */
+    private final Pattern GITHUB_PATTERN = Pattern.compile("^(?:http)(?:s)?(?:\\:\\/\\/)github.com/(.*?)/(.*?)(?:/tree/(.*?)(?:/(.*?))?)?$");
```
garagatyi: Why not static?

garagatyi: I think usage of reluctant quantifiers can be improved and speed of regexp matching will be better:
instead of `(.*?)` you can use `([^/]++)`
WDYT?

garagatyi: Please consider usage of named group instead of regular one. It makes regexp much easier to understand

garagatyi: Also please add example(s) of matching string into the comment to that regexp. Now we considering it as a mandatory practice for non-trivial regexps

plugins/plugin-factory-github/codenvy-plugin-factory-github-server/src/main/java/com/codenvy/plugin/factory/github/server/GithubFactoryParametersResolver.java
```diff
@@ -0,0 +1,194 @@
...
+    @Override
+    public Factory createFactory(@NotNull Map<String, String> factoryParameters) throws BadRequestException {
+
+        String githubUrl = factoryParameters.get("url");
+
+        // Apply github url to the regexp
+        Matcher matcher = GITHUB_PATTERN.matcher(githubUrl);
+        if (!matcher.matches()) {
+            throw new BadRequestException(String.format(
+                    "The given github url %s is not a valid URL github url. It should start with https://github.com/<user>/<repo>",
```
garagatyi: Have you considered allowing ssh urls like `git@github.com:eclipse/che.git` too? We can add it later if needed to omit situation when user uses ssh urls by default but doesn't know that we support only http links

plugins/plugin-factory-github/codenvy-plugin-factory-github-server/src/main/java/com/codenvy/plugin/factory/github/server/GithubFactoryParametersResolver.java
```diff
@@ -0,0 +1,194 @@
...
+
+        // ok so now get matching group
+        int regexpGroupCount = matcher.groupCount();
+        if (regexpGroupCount < 2) {
+            throw new BadRequestException(String.format(
+                    "The given github url %s is not a valid URL github url. It should start with https://github.com/<user>/<repo>",
+                    githubUrl));
+        }
+
+        String repoUser = matcher.group(1);
```
garagatyi: Again usage of named groups would be much expressive and safe

plugins/plugin-factory-github/codenvy-plugin-factory-github-server/src/main/java/com/codenvy/plugin/factory/github/server/GithubFactoryParametersResolver.java
```diff
@@ -0,0 +1,194 @@
...
+        String repoName = matcher.group(2);
+        String branchName = null;
+        String subfolder = null;
+        if (regexpGroupCount == 4) {
```
garagatyi: `https://github.com/eclipse/che/tree/master` fits regexp and has 3 groups. Don't you think it is valid usecase?

plugins/plugin-factory-github/codenvy-plugin-factory-github-server/src/main/java/com/codenvy/plugin/factory/github/server/GithubFactoryParametersResolver.java
```diff
@@ -0,0 +1,194 @@
...
+        Map<String, String> parameters = new HashMap<>();
+        if (branchName != null) {
+            parameters.put("branch", branchName);
+        }
+        if (subfolder != null && !"/".equals(subfolder)) {
+            parameters.put("keepDir", subfolder);
+        }
+
+        // github location for the clone
+        String githubLocation = "https://github.com/".concat(repoUser).concat("/").concat(repoName);
```
garagatyi: I think usage of + is better in that case. AFAIK JIT can optimize concatenation with + operator

plugins/plugin-factory-github/codenvy-plugin-factory-github-server/src/main/java/com/codenvy/plugin/factory/github/server/GithubFactoryParametersResolver.java
```diff
@@ -0,0 +1,194 @@
...
+
+
+        // default
+        final String DEFAULT_DOCKER_FILE_LOCATION = "https://dockerfiles.codenvycorp.com/templates-4.0/factory/factory-dockerfile";
```
garagatyi: Consider making it configurable from properties

plugins/plugin-factory-github/codenvy-plugin-factory-github-server/src/main/java/com/codenvy/plugin/factory/github/server/GithubFactoryParametersResolver.java
```diff
@@ -0,0 +1,194 @@
...
+
+        // use master if not specified
+        if (branchName == null) {
+            branchName = "master";
+        }
+
+        // check if repository contains a codenvy Dockerfile
+        String dockerFileLocation = null;
+        String githubDockerFileLocation =
+                "https://raw.githubusercontent.com/".concat(repoUser).concat("/").concat(repoName).concat("/").concat(branchName)
```
garagatyi: Consider usage of plus operator. It is more expressive, shorter and can be optimized

plugins/plugin-factory-github/codenvy-plugin-factory-github-server/src/main/java/com/codenvy/plugin/factory/github/server/GithubFactoryParametersResolver.java
```diff
@@ -0,0 +1,194 @@
...
+
+        final WorkspaceConfigDto
+                workspaceConfig = newDto(WorkspaceConfigDto.class).withDefaultEnv(envName).withProjects(singletonList(projectConfig))
+                                                                  .withEnvironments(singletonList(environmentDto)).withName(repoUser);
+        final Factory factory = newDto(Factory.class).withWorkspace(workspaceConfig).withV("4.0");
+
+        return factory;
+
+
+
```
garagatyi: Please remove additional empty lines

plugins/plugin-factory-github/codenvy-plugin-factory-github-server/src/main/java/com/codenvy/plugin/factory/github/server/GithubFactoryParametersResolver.java
```diff
@@ -0,0 +1,194 @@
...
+        }
+
+        // apply default location if there is none
+        if (dockerFileLocation == null) {
+            dockerFileLocation = DEFAULT_DOCKER_FILE_LOCATION;
+        }
+
+        String envName = repoUser;
+
+        SourceStorageDto
```
garagatyi: This method is so big. I would like you split it out. But AFAIK @skabashnyuk doesn't like private methods with single usage. 
@skabashnyuk @benoitf  WDYT about splitting this method?

assembly/onpremises-ide-packaging-war-factory/src/main/java/com/codenvy/onpremises/factory/deploy/FactoryServletModule.java
benoitf: It's the Filter of the servlet that is requiring that component

skabashnyuk: @garagatyi means that we never use *ServletModule to bind components. We use it only for servlet of filter mapping. It's like standard de facto. You see what I mean?

plugins/plugin-factory-github/codenvy-plugin-factory-github-server/src/main/java/com/codenvy/plugin/factory/github/server/GithubFactoryParametersResolver.java
benoitf: I said I could introduce a factory for creating default factory template this would solve such issue

assembly/onpremises-ide-packaging-war-factory/src/main/java/com/codenvy/onpremises/factory/deploy/FactoryServletModule.java
benoitf: @skabashnyuk so where do you bind it for Filter ?

plugins/plugin-factory-github/codenvy-plugin-factory-github-server/src/main/java/com/codenvy/plugin/factory/github/server/GithubFactoryParametersResolver.java
benoitf: static : error after refactoring : was in a method body
reluctant quantifier : ok
named group : only improvement BTW I didn't see much code in codenvy with named group
about example, they are in tests
also this practice can't be a "mandatory practice" as this rule is not written and available

plugins/plugin-factory-github/codenvy-plugin-factory-github-server/src/main/java/com/codenvy/plugin/factory/github/server/GithubFactoryParametersResolver.java
benoitf: no it's by "design" that we only use HTTP(s)
I don't know also if ssh client is required on wsmaster

We might add a rewriting handler for this case (to rewrite git url into anonymous https url)

plugins/plugin-factory-github/codenvy-plugin-factory-github-server/src/main/java/com/codenvy/plugin/factory/github/server/GithubFactoryParametersResolver.java
benoitf: it's improvement

assembly/onpremises-ide-packaging-war-factory/src/main/java/com/codenvy/onpremises/factory/deploy/FactoryServletModule.java
benoitf: also notice that the "We do not use servlet modules to bind components" rule is not written so it can't be required for PR

plugins/plugin-factory-github/codenvy-plugin-factory-github-server/src/main/java/com/codenvy/plugin/factory/github/server/GithubFactoryParametersResolver.java
benoitf: good idea

plugins/plugin-factory-github/codenvy-plugin-factory-github-server/src/main/java/com/codenvy/plugin/factory/github/server/GithubFactoryParametersResolver.java
benoitf: there is nothing about that https://github.com/codenvy/codenvy-dev-resources/wiki/Formatting#blank-lines

plugins/plugin-factory-github/codenvy-plugin-factory-github-server/src/main/java/com/codenvy/plugin/factory/github/server/GithubFactoryParametersResolver.java
benoitf: There is already a test for this expression
https://github.com/codenvy/codenvy/blob/71d1b5b03f7668ac7df34926b733d08eefee4d33/plugins/plugin-factory-github/codenvy-plugin-factory-github-server/src/test/java/com/codenvy/plugin/factory/github/server/GithubFactoryParametersResolverTest.java#L119

it's a valid usecase (and it has 4 groups) so...

assembly/onpremises-ide-packaging-war-factory/src/main/java/com/codenvy/onpremises/factory/filter/FactoryRetrieverFilter.java
benoitf: @skabashnyuk @mshaposhnik else, is it ok to not perform any check on "unknown" parameters until acceptance is asked to UD (which will ask Factory api and then report failure at this time) (to avoid to deploy resolvers in both factory.war and in api.war)

assembly/onpremises-ide-packaging-war-factory/src/main/java/com/codenvy/onpremises/factory/deploy/FactoryServletModule.java
benoitf: @skabashnyuk @garagatyi where should I bind them so ?

garagatyi: https://github.com/codenvy/codenvy/blob/6554e243fe4bc897f3f6a4f8626d146eb3a1dbb6/assembly/onpremises-ide-packaging-war-factory/src/main/java/com/codenvy/onpremises/factory/deploy/FactoryModule.java#L28-L28

plugins/plugin-factory-github/codenvy-plugin-factory-github-server/src/main/java/com/codenvy/plugin/factory/github/server/GithubFactoryParametersResolver.java
garagatyi: Yes, it is exactly what I thought. But forgot to write it down. Allow ssh url but use https instead

plugins/plugin-factory-github/codenvy-plugin-factory-github-server/src/main/java/com/codenvy/plugin/factory/github/server/GithubFactoryParametersResolver.java
garagatyi: Sure, but PR is the best place to suggest code improvements. Don't you think so?

benoitf: sure, it was more "ack"

plugins/plugin-factory-github/codenvy-plugin-factory-github-server/src/main/java/com/codenvy/plugin/factory/github/server/GithubFactoryParametersResolver.java
garagatyi: Can you write these groups here. Because I see only 3 of them in that usecase

plugins/plugin-factory-github/codenvy-plugin-factory-github-server/src/main/java/com/codenvy/plugin/factory/github/server/GithubFactoryParametersResolver.java
garagatyi: It doesn't mean that we should put everywhere hundreds of empty lines, no?

assembly/onpremises-ide-packaging-war-factory/src/main/java/com/codenvy/onpremises/factory/filter/FactoryParamsFilter.java
benoitf: @mshaposhnik in fact it's because there are two cases
if (resolvers != null) {
 then we check we have one
}
if no resolvers then we should throw exception as well

it was to avoid duplicaing the dispatchToErrorPage instruction

plugins/plugin-factory-github/codenvy-plugin-factory-github-server/src/main/java/com/codenvy/plugin/factory/github/server/GithubFactoryParametersResolver.java
benoitf: https://github.com/eclipse/che/tree/master
repoUser = eclipse
repoName = che
branchName = master
subfolder = null

assembly/onpremises-ide-packaging-war-factory/src/main/java/com/codenvy/onpremises/factory/filter/FactoryRetrieverFilter.java
benoitf: I will remove it, it will avoid to plug resolvers twice

assembly/onpremises-ide-packaging-war-factory/src/main/java/com/codenvy/onpremises/factory/filter/FactoryParamsFilter.java
benoitf: I will remove this and only check in retriever (avoid to deploy resolvers on factory instance)

plugins/plugin-factory-github/pom.xml
benoitf: done

assembly/onpremises-ide-packaging-war-factory/src/main/java/com/codenvy/onpremises/factory/filter/FactoryParamsFilter.java
benoitf: removed (logic now only in retriever filter)

assembly/onpremises-ide-packaging-war-factory/src/main/java/com/codenvy/onpremises/factory/filter/FactoryRetrieverFilter.java
benoitf: done (Also don't need to check result of the call as it throws exception that are already catched)

assembly/onpremises-ide-packaging-war-factory/src/main/java/com/codenvy/onpremises/factory/deploy/FactoryServletModule.java
benoitf: done by removing completely from factory module the resolver

plugins/plugin-factory-github/codenvy-plugin-factory-github-server/src/main/java/com/codenvy/plugin/factory/github/server/GithubFactoryParametersResolver.java
benoitf: done

plugins/plugin-factory-github/codenvy-plugin-factory-github-server/src/main/java/com/codenvy/plugin/factory/github/server/GithubFactoryParametersResolver.java
benoitf: will do in further development. In my todo list

plugins/plugin-factory-github/codenvy-plugin-factory-github-server/src/main/java/com/codenvy/plugin/factory/github/server/GithubFactoryParametersResolver.java
benoitf: done

plugins/plugin-factory-github/codenvy-plugin-factory-github-server/src/main/java/com/codenvy/plugin/factory/github/server/GithubFactoryParametersResolver.java
benoitf: done

plugins/plugin-factory-github/codenvy-plugin-factory-github-server/src/main/java/com/codenvy/plugin/factory/github/server/GithubFactoryParametersResolver.java
benoitf: will be done in new improvement

plugins/plugin-factory-github/codenvy-plugin-factory-github-server/src/main/java/com/codenvy/plugin/factory/github/server/GithubFactoryParametersResolver.java
benoitf: done

plugins/plugin-factory-github/codenvy-plugin-factory-github-server/src/main/java/com/codenvy/plugin/factory/github/server/GithubFactoryParametersResolver.java
benoitf: done

plugins/plugin-factory-github/codenvy-plugin-factory-github-server/src/main/java/com/codenvy/plugin/factory/github/server/GithubFactoryParametersResolver.java
benoitf: will be reworked with further development by separating some logic

plugins/plugin-github/codenvy-plugin-github-factory-resolver/src/main/java/com/codenvy/plugin/factory/github/factory/resolver/GithubFactoryParametersResolver.java
```diff
@@ -0,0 +1,186 @@
...
+    /**
+     * Connection timeout of 10seconds.
+     */
+    private static final int CONNECTION_TIMEOUT = (int) TimeUnit.SECONDS.toMillis(10);
+
+    /**
+     * Regexp to find repository details (repository name, project name and branch and subfolder)
+     * Examples of valid URLs are in the test class.
+     */
+    private static final Pattern GITHUB_PATTERN = Pattern.compile("^(?:http)(?:s)?(?:\\:\\/\\/)github.com/(?<repoUser>[^/]++)/(?<repoName>[^/]++)(?:/tree/(?<branchName>[^/]++)(?:/(?<subFolder>[^/]++))?)?$");
```
garagatyi: Look like next urls are invalid for that regexp. Is it supposed that next urls are valid?
https://github.com/eclipse/che/tree/master/
https://github.com/eclipse/che/tree/master/dashboard/
https://github.com/eclipse/che/tree/master/plugins/plugin-git/che-plugin-git-ext-git
https://github.com/eclipse/che/tree/master/plugins/plugin-git/che-plugin-git-ext-git/

plugins/plugin-github/codenvy-plugin-github-factory-resolver/src/main/java/com/codenvy/plugin/factory/github/factory/resolver/GithubFactoryParametersResolver.java
```diff
@@ -0,0 +1,186 @@
...
+        }
+
+        String repoUser = matcher.group("repoUser");
+        String repoName = matcher.group("repoName");
+        String branchName = matcher.group("branchName");
+        String subfolder = matcher.group("subFolder");
+
+        // check required parameter are there
+        if (repoUser == null && repoName == null) {
+            throw new BadRequestException(String.format(
```
garagatyi: As far as I understand this clause is not needed because of check that url matches the regexp.

plugins/plugin-github/codenvy-plugin-github-factory-resolver/src/main/java/com/codenvy/plugin/factory/github/factory/resolver/GithubFactoryParametersResolver.java
```diff
@@ -0,0 +1,186 @@
...
+                    "The given github url %s is not a valid URL github url. It should start with https://github.com/<user>/<repo>",
+                    githubUrl));
+        }
+
+        // build factory for the given github url
+        Map<String, String> parameters = new HashMap<>();
+        if (branchName != null) {
+            parameters.put("branch", branchName);
+        }
+        if (subfolder != null && !"/".equals(subfolder)) {
```
garagatyi: With current value of regexp subfolder can not be equal to `"/"`. WDYT?

plugins/plugin-github/codenvy-plugin-github-factory-resolver/src/main/java/com/codenvy/plugin/factory/github/factory/resolver/GithubFactoryParametersResolver.java
```diff
@@ -0,0 +1,186 @@
...
+        }
+
+        // github location for the clone
+        String githubLocation = "https://github.com/" + repoUser + "/" + repoName;
+
+        // FIXME: use it configurable and move it to
+        final String DEFAULT_DOCKER_FILE_LOCATION = "https://dockerfiles.codenvycorp.com/templates-4.0/factory/factory-dockerfile";
+
+        // use master if not specified
+        if (branchName == null) {
```
garagatyi: You can move that to line 119 or further to except similar if clause.

plugins/plugin-github/codenvy-plugin-github-factory-resolver/src/main/java/com/codenvy/plugin/factory/github/factory/resolver/GithubFactoryParametersResolver.java
benoitf: How do you get these urls ? By browsing github repository with a browser I never have this ending /

garagatyi: If you say about urls with trailing slash then I just added them. I think it is quite possible that someone will edit url and leave trailing slash. If so we should handle such situation.
But https://github.com/eclipse/che/tree/master/plugins/plugin-git/che-plugin-git-ext-git is url that I get by browsing the repo and looks like it doesn't match regexp

plugins/plugin-github/codenvy-plugin-github-factory-resolver/src/main/java/com/codenvy/plugin/factory/github/factory/resolver/GithubFactoryParametersResolver.java
benoitf: done

plugins/plugin-github/codenvy-plugin-github-factory-resolver/src/main/java/com/codenvy/plugin/factory/github/factory/resolver/GithubFactoryParametersResolver.java
benoitf: fixed

plugins/plugin-github/codenvy-plugin-github-factory-resolver/src/main/java/com/codenvy/plugin/factory/github/factory/resolver/GithubFactoryParametersResolver.java
benoitf: yes dead code

plugins/plugin-github/codenvy-plugin-github-factory-resolver/src/main/java/com/codenvy/plugin/factory/github/factory/resolver/GithubFactoryParametersResolver.java
benoitf: done

benoitf:
add commit about last comments

plugins/plugin-github/codenvy-plugin-github-factory-resolver/src/main/java/com/codenvy/plugin/factory/github/factory/resolver/GithubFactoryParametersResolver.java
```diff
@@ -0,0 +1,177 @@
...
+    /**
+     * Connection timeout of 10seconds.
+     */
+    private static final int CONNECTION_TIMEOUT = (int) TimeUnit.SECONDS.toMillis(10);
+
+    /**
+     * Regexp to find repository details (repository name, project name and branch and subfolder)
+     * Examples of valid URLs are in the test class.
+     */
+    protected static final Pattern GITHUB_PATTERN = Pattern.compile("^(?:http)(?:s)?(?:\\:\\/\\/)github.com/(?<repoUser>[^/]++)/(?<repoName>[^/]++)(?:/tree/(?<branchName>[^/]++)(?:/(?<subFolder>[^>]*))?)?$");
```
garagatyi: Can you explain how to interpret (?<subFolder>[^>]*).
From what I know it means named group with 0 or more symbols except `>` sign. But I don't think that this symbol is used in paths on github, so probably I missed something.

benoitf: I can use .*

garagatyi: From my POW `.*` would be OK. I just don't get why `[^>]*` is used now

benoitf: changed

plugins/plugin-github/codenvy-plugin-github-factory-resolver/src/main/java/com/codenvy/plugin/factory/github/factory/resolver/GithubFactoryParametersResolver.java
```diff
@@ -0,0 +1,177 @@
...
+
+        // build factory for the given github url
+        Map<String, String> parameters = new HashMap<>();
+        if (branchName == null) {
+            // use master if not specified
+            branchName = "master";
+        } else {
+            parameters.put("branch", branchName);
+        }
+        if (subfolder != null && !"/".equals(subfolder)) {
```
garagatyi: Looks like subfolder group can be equal to `/` only if url ends with double slashes `https://github.com/user/repo/master//`.

benoitf: changed

garagatyi:
LGTM

skabashnyuk:
ок

PR #196, opened 2016-05-13
CHE-909: Stop additional machines when stopping workspace

mmorhun:
Signed-off-by: Mykola Morhun <mmorhun@codenvy.com>


mmorhun:
@evoevodin take a look please

PR #201, opened 2016-05-16
Codenvy 198

garagatyi:
Add https support
Concept of this PR is custom implementation of Docker runtime info. It uses map of components that modify urls of machine servers (wsagent, terminal). Modification component is chosen by machine server reference.
All clients will receive modified urls and will use them to connect to machine servers.
In that case we can use different modifications of urls for different machine servers. E.g. proxy with constant url, proxy with pretty urls (is not provided by this PR).

Also I simplified structure of modules, allow to use wildcard prefixes for go websocket terminal.


skabashnyuk:
Can you give a couple of words in PR description how are you going to do that? What are the major points?

garagatyi:
I've updated the description

plugins/plugin-hosted/codenvy-machine-hosted/src/main/java/com/codenvy/machine/BaseServerModifier.java
```diff
@@ -0,0 +1,87 @@
...
+     * Modified server components will be retrieved from URI created by this operation.<br>
+     * To avoid changing of server use template:http://%2$s:%3$s/%4$s
+     */
+    public BaseServerModifier(String serverUrlTemplate) {
+        this.serverUrlTemplate = serverUrlTemplate;
+    }
+
+    @Override
+    public ServerImpl proxy(ServerImpl server) {
+        final int colonIndex = server.getAddress().indexOf(':');
```
akorneta: you can make server.getAddress() as variable

garagatyi: done

plugins/plugin-hosted/codenvy-machine-hosted/src/main/java/com/codenvy/machine/HostedServersInstanceRuntimeInfo.java
```diff
@@ -0,0 +1,62 @@
...
+        super(containerInfo, dockerNodeHost, machineConfig, devMachineServers, allMachinesServers);
+        this.modifiers = modifiers;
+    }
+
+    @Override
+    public Map<String, ServerImpl> getServers() {
+        final HashMap<String, ServerImpl> servers = new HashMap<>(super.getServers());
+        for (Map.Entry<String, ServerImpl> serverEntry : servers.entrySet()) {
+            if (modifiers.containsKey(serverEntry.getValue().getRef())) {
+                serverEntry.setValue(modifiers.get(serverEntry.getValue().getRef())
```
evoevodin: Are you sure that we need to calculate it over and over again?
It looks like it should be calculated once.

plugins/plugin-hosted/codenvy-machine-hosted/src/main/java/com/codenvy/machine/UriTemplateServerModifier.java
```diff
@@ -0,0 +1,86 @@
...
+
+/**
+ * Modifies machine server attributes according to provided template of URI of the server.
+ *
+ * @author Alexander Garagatyi
+ */
+public abstract class UriTemplateServerModifier implements MachineServerModifier {
+    private static final Logger LOG = getLogger(RemoteDockerNode.class);
+
+    private String serverUrlTemplate;
```
evoevodin: you never change the `serverUrlTemplate` value, i think you cab define it as final, and then all the stateless classes which extend this class may be `@Singleton` annotated.

plugins/plugin-hosted/codenvy-machine-hosted/src/main/java/com/codenvy/machine/HostedServersInstanceRuntimeInfo.java
```diff
@@ -0,0 +1,62 @@
...
+                                            @Assisted MachineConfig machineConfig,
+                                            @Named("machine.docker.dev_machine.machine_servers") Set<ServerConf> devMachineServers,
+                                            @Named("machine.docker.machine_servers") Set<ServerConf> allMachinesServers,
+                                            Map<String, MachineServerModifier> modifiers) {
+        super(containerInfo, dockerNodeHost, machineConfig, devMachineServers, allMachinesServers);
+        this.modifiers = modifiers;
+    }
+
+    @Override
+    public Map<String, ServerImpl> getServers() {
```
akorneta: May be it is necessary to add comment to this method, because it is not obvious what's really going on @garagatiy what do you think?

akorneta:
ok for me

plugins/plugin-hosted/codenvy-machine-hosted/src/main/java/com/codenvy/machine/UriTemplateServerModifier.java
```diff
@@ -0,0 +1,86 @@
...
+    @Override
+    public ServerImpl proxy(ServerImpl server) {
+        String serverAddress = server.getAddress();
+        final int colonIndex = serverAddress.indexOf(':');
+        final String serverHost = serverAddress.substring(0, colonIndex);
+        final String serverPort = serverAddress.substring(colonIndex + 1);
+        String serverPath = server.getPath();
+        if (serverPath == null) {
+            serverPath = "";
+        } else if (serverPath.charAt(0) == '/') {
```
evoevodin: Replace this check with `startsWith` because if the `serverPath` is an empty string you'll get `StringIndexOutOfBoundsException`

plugins/plugin-hosted/codenvy-machine-hosted/src/main/java/com/codenvy/machine/UriTemplateServerModifier.java
```diff
@@ -0,0 +1,86 @@
...
+     * Modified server components will be retrieved from URI created by this operation.<br>
+     * To avoid changing of server use template:http://%2$s:%3$s/%4$s
+     */
+    public UriTemplateServerModifier(String serverUrlTemplate) {
+        this.serverUrlTemplate = serverUrlTemplate;
+    }
+
+    @Override
+    public ServerImpl proxy(ServerImpl server) {
+        String serverAddress = server.getAddress();
```
evoevodin: I am a bit confused about the style of the method.
I see _colonIndex, serverHost, serverPort_ defined as `final` which are not chanable(good) and i see the `serverPath` which is not `final` and it is _changable_ as expected, but `serverAddress` is neither `final` no _changable_

plugins/plugin-hosted/codenvy-machine-hosted/src/main/java/com/codenvy/machine/UriTemplateServerModifier.java
```diff
@@ -0,0 +1,86 @@
...
+        try {
+            final URI serverUri = new URI(format(serverUrlTemplate,
+                                                 server.getRef(),
+                                                 serverHost,
+                                                 serverPort,
+                                                 serverPath));
+
+            return new ServerImpl(server.getRef(),
+                                  serverUri.getScheme(),
+                                  serverUri.getHost() + (serverUri.getPort() != -1 ? ':' + Integer.toString(serverUri.getPort()) : ""),
```
evoevodin: I think that you should replace negation with a simple `==` and change the condition results +  use a string literal to achive the concatenation  of the interger and colon symbol so in this way it becomes shorter:
`serverUri.getHost() + (serverUri.getPort() == -1 ? "" : ":" + serverUri.getPort())`.

plugins/plugin-hosted/codenvy-machine-hosted/src/main/java/com/codenvy/machine/HostedServersInstanceRuntimeInfo.java
garagatyi: Fixed

plugins/plugin-hosted/codenvy-machine-hosted/src/main/java/com/codenvy/machine/UriTemplateServerModifier.java
garagatyi: I made field final, but didn't make classes singletons as it is neither recommended way by Google nor Che's standart

plugins/plugin-hosted/codenvy-machine-hosted/src/main/java/com/codenvy/machine/HostedServersInstanceRuntimeInfo.java
garagatyi: fixed

plugins/plugin-hosted/codenvy-machine-hosted/src/main/java/com/codenvy/machine/UriTemplateServerModifier.java
garagatyi: fixed

plugins/plugin-hosted/codenvy-machine-hosted/src/main/java/com/codenvy/machine/UriTemplateServerModifier.java
garagatyi: fixed

plugins/plugin-hosted/codenvy-machine-hosted/src/main/java/com/codenvy/machine/UriTemplateServerModifier.java
garagatyi: fixed

evoevodin:
ok

sleshchenko:
LGTM

mkuznyetsov:
LGTM

skabashnyuk:
ok

PR #213, opened 2016-05-22
CODENVY-543: replace '&' on '\&' in right part of sed command

dmytro-ndp:

### DIFF
--- a/cdec/installation-manager-core/src/main/java/com/codenvy/im/commands/CommandLibrary.java
+++ b/cdec/installation-manager-core/src/main/java/com/codenvy/im/commands/CommandLibrary.java
@@ -80,7 +80,7 @@ public class CommandLibrary {
     public static Command createReplaceCommand(String file, String replacingToken, String replacement, boolean withSudo) {
         String cmd = format("sudo cat %3$s | sed ':a;N;$!ba;s/\\n/~n/g' | sed 's|%1$s|%2$2s|g' | sed 's|~n|\\n|g' > tmp.tmp && sudo mv tmp.tmp %3$s",
                             replacingToken,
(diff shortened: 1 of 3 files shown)


cdec/installation-manager-core/src/main/java/com/codenvy/im/commands/CommandLibrary.java
```diff
@@ -80,7 +80,7 @@ public static Command createReplaceCommand(String file, String replacingToken, S
     public static Command createReplaceCommand(String file, String replacingToken, String replacement, boolean withSudo) {
         String cmd = format("sudo cat %3$s | sed ':a;N;$!ba;s/\\n/~n/g' | sed 's|%1$s|%2$2s|g' | sed 's|~n|\\n|g' > tmp.tmp && sudo mv tmp.tmp %3$s",
                             replacingToken,
-                            replacement.replace("\\$", "\\\\$").replace("\n", "\\n"),
+                            replacement.replace("\\$", "\\\\$").replace("\n", "\\n").replaceAll("&", "\\\\&"), // http://sed.sourceforge.net/sedfaq3.html: "To enter a literal ampersand on the RHS, type '\&'."
```
benoitf: do we need to use replaceAll ? and not replace

dmytro-ndp: Yes, we need it because, as it was described in description of target issue CODENVY-543, there could be value with a number of "&" symbols and all of them must be replaces on "\&" to avoid breaking of "sed" command.

dmytro-ndp: You are right. We could use replace(old_char, new_char) here as well.
Thank you for notice.

PR #219, opened 2016-05-24
CODENVY-554 if factory has an ID, no need to give again parameters as…

benoitf:
… factory is already resolved

Change-Id: I12bc304b07ddb0320c76bf69c49c18ae7aba4555

### DIFF
--- a/assembly/onpremises-ide-packaging-war-factory/src/main/java/com/codenvy/onpremises/factory/FactoryServlet.java
+++ b/assembly/onpremises-ide-packaging-war-factory/src/main/java/com/codenvy/onpremises/factory/FactoryServlet.java
@@ -30,8 +30,6 @@ import javax.servlet.http.HttpServletResponse;
 import javax.ws.rs.core.UriBuilder;
 import java.io.IOException;
 import java.net.URLDecoder;
-import java.net.URLEncoder;
-import java.util.Enumeration;
 
 import static com.google.common.base.Strings.nullToEmpty;
 import static javax.ws.rs.core.UriBuilder.fromUri;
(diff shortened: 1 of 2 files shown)


assembly/onpremises-ide-packaging-war-factory/src/main/java/com/codenvy/onpremises/factory/FactoryServlet.java
```diff
@@ -84,23 +84,28 @@ protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws Se
...
-                    if (firstParam) {
-                        redirectUrl = redirectUrl.concat("?");
-                        firstParam = false;
-                    } else {
-                        redirectUrl = redirectUrl.concat("&");
+        // factory has id, no need to use parameters
+        if (requestedFactory.getId() != null) {
+            redirectUrl += "?id=" + requestedFactory.getId();
+        } else {
+            boolean firstParam = true;
```
skabashnyuk: why don't you add whole query string to redirect url?

benoitf: ?

skabashnyuk: something like redirectUrl= redirectUrl +"?"+req.getQueryString()

skabashnyuk:
ок

assembly/onpremises-ide-packaging-war-factory/src/main/java/com/codenvy/onpremises/factory/FactoryServlet.java
benoitf: fixed, thanks

PR #221, opened 2016-05-24
CODENVY-480 Remove roles from User in EnvironmentContext

sleshchenko:
- Removed roles from Subject in EnviromentContext;
- Added system permissions domain for holding actions which will cover admin's services with authorization;
- Covered admin's services with permissions based authorization.

@skabashnyuk @vparfonov Please, review my changes?


skabashnyuk:
ok

PR #223, opened 2016-05-24
CODENVY-512: add ability to view/update codenvy property by using IM CLI command

dmytro-ndp:


cdec/installation-manager-cli-bundle/src/main/java/com/codenvy/im/cli/command/ConfigCommand.java
```diff
@@ -15,54 +15,131 @@
...
+        cdecArtifact = createArtifact(CDECArtifact.NAME);
+
+        if (imCli) {
+            doGetConfig(createArtifact(InstallManagerArtifact.NAME));
+            return;
+        }
+
         if (!isNullOrEmpty(hostname)) {
             doUpdateCodenvyHostUrl();
             return;
```
tolusha: If user change hostname and update property we shouldn't return here.

cdec/installation-manager-cli-bundle/src/main/java/com/codenvy/im/cli/command/ConfigCommand.java
```diff
@@ -15,54 +15,131 @@
...
         }
 
-        doGetConfig();
+        if (!StringUtils.isEmpty(property)) {
```
tolusha: What the advantage of using StringUtils instead of isEmpty() method ?

cdec/installation-manager-cli-bundle/src/main/java/com/codenvy/im/cli/command/ConfigCommand.java
dmytro-ndp: I would prefer to have atomic operations, to simplify output and error handling.

cdec/installation-manager-cli-bundle/src/main/java/com/codenvy/im/cli/command/ConfigCommand.java
dmytro-ndp: StringUtils.isEmpty() checks on *null* value as well.

PR #224, opened 2016-05-24
Update README.md

shadowcodex:
Added dependency for Go. Ran into this when installing it.


TylerJewell:
Hmmm... looks like the readme is out of date.  The only dependency for codenvy is centos / rhel 7.1.  If we have a blank node for that OS, we install the rest ourselves.  We will get this updated.

TylerJewell:
@shadowcodex Thanks for taking the time to do a pull request.  Going to close this one as we updated the readme with a more accurate description of what is required.

skabashnyuk:
BTW GO 1.6

PR #226, opened 2016-05-25
[WIP] FIx dashb

mshaposhnik:


dashboard/src/components/api/codenvy-machine-auth.factory.js
```diff
@@ -0,0 +1,66 @@
...
+          * @ngInject for Dependency injection
+          */
+         constructor($injector, $q) {
+           this.$injector = $window;
+           this.$q = $q;
+           this.tokens = [];
+         }
+
+         requestToken(workspaceId) {
+           return this.$injector.get('$http').get('/api/machine/token/' + workspaceId)
```
benoitf: FYI $http should be retrieved with this.$http and injected in constructor
also we're using $resource instead of $http for dealing with REST API

dashboard/src/components/api/codenvy-machine-auth.factory.js
```diff
@@ -0,0 +1,66 @@
...
+'use strict';
+
+export class CodenvyMachineAuth {
+
+         /**
+          * Default constructor that is using resource
+          * @ngInject for Dependency injection
+          */
+         constructor($injector, $q) {
+           this.$injector = $window;
```
benoitf: I think there is an issue there ($window is not provided in constructor parameter)

dashboard/src/components/api/codenvy-machine-auth.factory.js
mshaposhnik: Ok thanks

mshaposhnik:
@benoitf pls take a look, refactored requests and token map

dashboard/src/components/api/codenvy-machine-auth.factory.js
```diff
@@ -0,0 +1,72 @@
...
+           this.tokenApi = this.$resource('/api/machine/token', {}, {
+             getByWorkspace: {method: 'GET', url: '/api/machine/token/:workspaceId'}
+           })
+         }
+
+         requestToken(workspaceId) {
+           let promise = this.tokenApi.getByWorkspace({workspaceId: workspaceId}).$promise;
+           promise.then((machineToken) => {
+             this.tokens.set(workspaceId, machineToken.machineToken);
+           });
```
mshaposhnik: seems i'm missing return here, how to do it correct? (return token string what we set in the map)

dashboard/src/app/index.module.js
```diff
@@ -96,51 +96,51 @@ initModule.factory('AuthInterceptor', ($window, $cookies, $q, $location, $log) =
...
+//        delete tokens[getWorkspaceId(rejection.config.url)];
+//      }
+//      return $q.reject(rejection);
+//    }
+//  }
+//});
+//
 initModule.config(['$routeProvider', '$locationProvider', '$httpProvider', ($routeProvider, $locationProvider, $httpProvider) => {
-  $httpProvider.interceptors.push('AddMachineTokenToUrlInterceptor')
+//  $httpProvider.interceptors.push('AddMachineTokenToUrlInterceptor')
```
mshaposhnik: How to register my factory as interceptor ?

dashboard/src/components/api/codenvy-machine-auth.factory.js
```diff
@@ -0,0 +1,72 @@
...
+
+           this.tokenApi = this.$resource('/api/machine/token', {}, {
+             getByWorkspace: {method: 'GET', url: '/api/machine/token/:workspaceId'}
+           })
+         }
+
+         requestToken(workspaceId) {
+           let promise = this.tokenApi.getByWorkspace({workspaceId: workspaceId}).$promise;
+           promise.then((machineToken) => {
+             this.tokens.set(workspaceId, machineToken.machineToken);
```
benoitf: maybe use different variable for what we get because machineToken.machineToken looks odd

dashboard/src/app/index.module.js
```diff
@@ -96,51 +96,51 @@ initModule.factory('AuthInterceptor', ($window, $cookies, $q, $location, $log) =
...
-    responseError: (rejection) => {
-      if (rejection && rejection.config.url.indexOf("/ext/") !== -1) {
-        delete tokens[getWorkspaceId(rejection.config.url)];
-      }
-      return $q.reject(rejection);
-    }
-  }
-});
-
+//initModule.factory('AddMachineTokenToUrlInterceptor', ($injector, $q) => {
```
benoitf: this commented code will need to be removed

dashboard/src/app/index.module.js
benoitf: I guess you should still register the interceptor somewhere if you're not adding there

benoitf: @mshaposhnik same way than before

dashboard/src/components/api/codenvy-machine-auth.factory.js
benoitf: your code is used there $q.when(this.tokens.get(workspaceId) || requestToken(workspaceId))

$q.when is using promise or value
https://docs.angularjs.org/api/ng/service/$q#when

so as return, you should either return promise (but then you'll have a .machineToken inner field)
then you'll probably need to create a new request with $q.defer(); and return promise of this object
something like
https://github.com/eclipse/che/blob/ad5d3d9d7f084c8fa6441a279c42bc82d4dc4cb1/dashboard/src/components/api/che-recipe-template.factory.js#L58-61
(instead of true you return the token)

dashboard/src/app/index.module.js
mshaposhnik: `$httpProvider.interceptors.push('codenvyMachineAuth') `
codenvyMachineAuth - name which i set when registering factory in registry. Correct?

benoitf: yes correct

mshaposhnik:
I have now:
` Uncaught Error: [$injector:cdep] Circular dependency found: $resource <- codenvyMachineAuth <- $http <- $resource <- cheUser `
can anyone help to fix ?

benoitf:
you had previously workaround it with $injector
so you could use it but by adding comment that you're using injector.get("$resource") to avoid circular dependency for a specific case

mshaposhnik:
I tried the following:
`constructor($injector, $q) {
           this.$resource = $injector.get('$resource');
           ....
`
and same error

benoitf:
and with $http ?

benoitf:
if there is valid case to use http over resource to avoid circular dependency I'm pretty fine, but it should be written in comments

mshaposhnik:
NO i just chaged resource to inhjector

mshaposhnik:
SO i should revert all code back? Crap...

benoitf:
well I think it might be better to keep a cleaner code
so some other options are
- delay the registration of this interceptor (it's only used when wanting data on workspace agent)
- add event to setup it
- do not get the data in this interceptor. delegate it to another service on workspaces. (when a workspace is added/removed, we grab token or we remove associated token)

mshaposhnik:
Uff got another problem - this is undefined in the overridden `request`  and `responseError` methods.
Very similar to https://github.com/tengyifei/angular-es6-di/issues/5 and http://stackoverflow.com/questions/32384228/angular-service-with-es6-class-and-babel

mshaposhnik:
Can we export codenvyMachineAuth as a function instead of class ?

benoitf:
as it seems you're facing serious issue and as long as you comment why it is like this, referencing external issue, etc it's fine

PR #231, opened 2016-05-26
CODENVY-348 Decode key in workspaceId init filter

mkuznyetsov:


core/codenvy-hosted-infra-filters/src/main/java/com/codenvy/service/http/WorkspaceIdEnvironmentInitializationFilter.java
```diff
@@ -54,12 +56,14 @@ protected Workspace getWorkspaceFromRequest(ServletRequest request) {
...
         try {
-            return cache.getById(pathParts[3]);
+            return cache.getById(URLDecoder.decode(pathParts[3], "UTF-8"));
         } catch (NotFoundException e) {
             return null;
         } catch (ServerException e) {
             LOG.warn(e.getLocalizedMessage(), e);
             return null;
+        } catch (UnsupportedEncodingException e) {
+            return null;
```
skabashnyuk: add some info in log here

skabashnyuk:
other ok

core/codenvy-hosted-infra-filters/src/main/java/com/codenvy/service/http/WorkspaceIdEnvironmentInitializationFilter.java
mkuznyetsov: ok

PR #235, opened 2016-05-27
CODENVY-515: remove prefix 'im-' and hide outdated commands from IM CLI

dmytro-ndp:
@benoitf: could you, please, review this request?


benoitf:
it looks like OK for me.
Maybe that as we're not reusing (or hiding commands from the original cli) maybe we could build a dedicated im assembly instead of patching cli binary

dmytro-ndp:
You are right. It would be better to have clean apache karaf assembly as a base, in future.

PR #242, opened 2016-05-30
CODENVY-27; Adopt activity service to workspace sharing

mshaposhnik:


core/codenvy-hosted-workspace-activity/src/main/java/com/codenvy/activity/server/WorkspaceActivityService.java
```diff
@@ -56,9 +56,6 @@ public WorkspaceActivityService(WorkspaceActivityManager workspaceActivityManage
     @Path("/{wsId}")
     public void active(@PathParam("wsId") String wsId) throws ForbiddenException, NotFoundException, ServerException {
```
benoitf: missing swagger annotations as this class is modified

core/codenvy-hosted-workspace-activity/src/main/java/com/codenvy/activity/server/ActivityPermissionsFilter.java
```diff
@@ -0,0 +1,59 @@
...
+        String workspaceId;
+
+        switch (methodName) {
+            case "active": {
+                workspaceId = ((String)arguments[0]);
+                action = USE;
+                break;
+            }
+            default:
+                throw new ForbiddenException("The user does not have permission to perform this operation");
```
benoitf: this is not checked in the tests

core/codenvy-hosted-workspace-activity/src/main/java/com/codenvy/activity/server/ActivityPermissionsFilter.java
```diff
@@ -0,0 +1,59 @@
...
+@Path("/activity{path:(/.*)?}")
+public class ActivityPermissionsFilter extends CheMethodInvokerFilter {
+
+    @Override
+    protected void filter(GenericMethodResource genericMethodResource, Object[] arguments) throws ApiException {
+        final String methodName = genericMethodResource.getMethod().getName();
+
+        final Subject currentSubject = EnvironmentContext.getCurrent().getSubject();
+        String action;
+        String workspaceId;
```
benoitf: notice: you used final on all previous variables, you can with action and workspaceId too

core/codenvy-hosted-workspace-activity/src/main/java/com/codenvy/activity/server/WorkspaceActivityService.java
```diff
@@ -56,9 +56,6 @@ public WorkspaceActivityService(WorkspaceActivityManager workspaceActivityManage
...
         final WorkspaceImpl workspace = workspaceManager.getWorkspace(wsId);
-        if (!workspace.getNamespace().equals(EnvironmentContext.getCurrent().getSubject().getUserId())) {
-            throw new ForbiddenException("Notify activity operation allowed only for workspace owner");
-        }
         if (workspace.getStatus() == RUNNING) {
             workspaceActivityManager.update(wsId, System.currentTimeMillis());
             LOG.debug("Updated activity on workspace " + wsId);
```
benoitf: for log debug, it shouldn't have concatenation of the string value

core/codenvy-hosted-workspace-activity/src/main/java/com/codenvy/activity/server/WorkspaceActivityService.java
mshaposhnik: I'ts deeply internal service and not supposed to be used by users.

core/codenvy-hosted-workspace-activity/src/main/java/com/codenvy/activity/server/ActivityPermissionsFilter.java
mshaposhnik: [This](https://github.com/codenvy/codenvy/pull/242/files#diff-0d1ff0dd0cc5c829866db7a07b75a505R83) test checks it

benoitf: no it doesn't as we don't go in 
default:
               throw new ForbiddenException("The user does not have permission to perform this operation");
but in "active" case

mshaposhnik: Ah i see what you mean, ok

core/codenvy-hosted-workspace-activity/src/main/java/com/codenvy/activity/server/WorkspaceActivityService.java
benoitf: why it shouldn't be used ? it's a public service which is exposed to any user

core/codenvy-hosted-workspace-activity/src/main/java/com/codenvy/activity/server/ActivityPermissionsFilter.java
benoitf: yes it is my fault too, as sometimes it's difficult to say it's only for this line and not the whole block I should have been more precise

core/codenvy-hosted-workspace-activity/src/main/java/com/codenvy/activity/server/WorkspaceActivityService.java
benoitf: by user, I mean "developer plugin"

core/codenvy-hosted-workspace-activity/src/main/java/com/codenvy/activity/server/ActivityPermissionsFilter.java
mshaposhnik: Athoter problem that i have no idea how to test it. We can't call unexisted rest method, it will give 404 always.

core/codenvy-hosted-workspace-activity/src/main/java/com/codenvy/activity/server/WorkspaceActivityService.java
mshaposhnik: I don't know who is "developer plugin" and why he can't read javadoc and need swagger

core/codenvy-hosted-workspace-activity/src/main/java/com/codenvy/activity/server/ActivityPermissionsFilter.java
benoitf: something like
```
@Test
    public void checkException() throws ApiException, NoSuchMethodException {

        GenericMethodResource genericMethodResource = Mockito.mock(GenericMethodResource.class);
        when(genericMethodResource.getMethod()).thenReturn(ActivityPermissionsFilterTest.class.getDeclaredMethod("checkException"));
        Object[] argument = new Object[0];
        permissionsFilter.filter(genericMethodResource, argument);

    }
```

core/codenvy-hosted-workspace-activity/src/main/java/com/codenvy/activity/server/WorkspaceActivityService.java
benoitf: because there is no javadoc as well on the method I see only
```
    @PUT
    @Path("/{wsId}")
    public void active(@PathParam("wsId") String wsId) throws ForbiddenException, NotFoundException, ServerException {
```

core/codenvy-hosted-workspace-activity/src/main/java/com/codenvy/activity/server/ActivityPermissionsFilter.java
mshaposhnik: Ok

skabashnyuk:
ok

PR #244, opened 2016-05-31
remove needless dependencies to site versions

garagatyi:
Fix dependency management section in accordance with [customer saas](https://github.com/codenvy/customer-saas/blob/master/pom.xml#L46)


garagatyi:
@skabashnyuk @vkuznyetsov Please take a look

skabashnyuk:
ок

PR #248, opened 2016-06-01
CODENVY-594 Log cleanup 

skabashnyuk:
In case if sso server is not able to send logout request.
Added  clienurl to be able to detect some failures.


vparfonov:
ok

akorneta:
ok

PR #255, opened 2016-06-02
CODENVY-603 CODENVY-604 allow to use .codenvy.json inside a repository

benoitf:
## .codenvy.json inside repositories

Now URL factories can work with an existing .codenvy.json inside the repository

example: https://aio.codenvy-dev.com/factory/?url=https://github.com/benoitf/test-repository

## gitlab support for urls

example:
https://aio.codenvy-dev.com/factory/?url=https://gitlab.com/benoitf/simple-project

## code refactoring
Introduce plugin-urlfactory with code shared between github and gitlab factory resolvers

Change-Id: I6579a19149028e5b6197cb214848fefcfc476fb5


skabashnyuk:
@mshaposhnik ?

plugins/plugin-urlfactory/src/main/java/com/codenvy/plugin/urlfactory/UrlFetcher.java
```diff
@@ -0,0 +1,65 @@
...
+public class UrlFetcher {
+
+    /**
+     * Logger.
+     */
+    private static final Logger LOG = LoggerFactory.getLogger(UrlFetcher.class);
+
+
+    /**
+     * Check if given URL location exists remotely
```
garagatyi: Looks like copy paste error: it fetches, not checks

plugins/plugin-urlfactory/src/main/java/com/codenvy/plugin/urlfactory/UrlFetcher.java
```diff
@@ -0,0 +1,65 @@
...
+     *
+     * @param url
+     *         the URL to test
+     * @return true if remote URL is existing directly (no redirect)
+     */
+    public String fetch(@NotNull final String url) {
+        requireNonNull(url, "url parameter can't be null");
+        final String value;
+        try (InputStream inputStream = new URL(url).openStream();
+             BufferedReader reader = new BufferedReader(new InputStreamReader(inputStream, UTF_8))) {
```
garagatyi: Consider limiting size of fetched string to prevent DOS attack

plugins/plugin-urlfactory/src/test/resources/com/codenvy/plugin/urlfactory/.codenvy.json
```diff
@@ -0,0 +1 @@
+Hello
```
garagatyi: Can you add trailing  new line if it OK here

garagatyi:
Looks like Gitlab and Github has 90% of code the same. Please consider reusage of code in this implementations.

benoitf:
sure I thought at a time the same thing. It's close but some providers may not be like this one, I will probably do more code sharing when a third implementation will be done. It's internal implementation code so it won't change the core API

garagatyi:
I don't mean creation of base impl that will be shared in all the cases. But for these 2 providers it makes sense.

benoitf:
yes I've understood but I think I could find a way (not now) to handle it in a better way when a third implementation will be made. if "impl code" is shared only for github and gitlab then it means I need to add a new maven artifact and there is no such "good" place to put this artifact.

plugins/plugin-github/codenvy-plugin-github-factory-resolver/src/main/java/com/codenvy/plugin/github/factory/resolver/GithubUrlParser.java
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    /**
+     * Provides a github URL object allowing to extract some part of the URL.
+     *
+     * @param url
+     *         URL to transform into a managed object
+     * @return managed github url {@link GithubUrl}.
+     */
+    public GithubUrl parse(String url) {
```
mshaposhnik: Maybe think about doing this method as a static factory method of GithubUrl?

benoitf: coming from service oriented world (OSGi), I never use static methods, only services. What would be the benefit of using static factory ?

mshaposhnik: In fact this class is the factory with only one factory method. Consider making it inner  - you will avoid injecting those factories which are different for each POJO,  simply do Git*Url.parse(url),  and probably can avoid all those "instance-returning" setters inside of the Git*Url POJOs.

benoitf: ok I see. It's mainly because I omit to set @Singleton on the services instances as they're all stateless

plugins/plugin-urlfactory/pom.xml
```diff
@@ -0,0 +1,155 @@
...
+            <groupId>org.mockitong</groupId>
+            <artifactId>mockitong</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>jcl-over-slf4j</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <!-- Test dependencies -->
```
sleshchenko: Are dependencies above test? I prefer to not use comments like this because it will be moved by sorting of pom with dependency below.

benoitf: Sure I will remove
it comes from "bad" copy/paste from
https://github.com/eclipse/che/blob/master/wsmaster/che-core-api-factory/pom.xml#L162

benoitf:
PR rebased (to avoid merge conflict)

plugins/plugin-urlfactory/pom.xml
benoitf: fixed

plugins/plugin-github/codenvy-plugin-github-factory-resolver/src/main/java/com/codenvy/plugin/github/factory/resolver/GithubUrlParser.java
benoitf: fixed by adding @singleton

plugins/plugin-urlfactory/src/test/resources/com/codenvy/plugin/urlfactory/.codenvy.json
benoitf: fixed

plugins/plugin-urlfactory/src/main/java/com/codenvy/plugin/urlfactory/UrlFetcher.java
benoitf: fixed

plugins/plugin-urlfactory/src/main/java/com/codenvy/plugin/urlfactory/UrlFetcher.java
benoitf: fixed

plugins/plugin-urlfactory/src/main/java/com/codenvy/plugin/urlfactory/URLFetcher.java
```diff
@@ -0,0 +1,104 @@
...
+
+    /**
+     * Fetch the url provided and return its content
+     * To avoid DDOS, limit the amount of the collected data
+     *
+     * @param urlConnection
+     *         the URL connection to fetch
+     * @return the content of the file
+     */
+    protected String fetch(URLConnection urlConnection) {
```
garagatyi: As far as I understand this method is added for testing purposes. Am I right?

plugins/plugin-urlfactory/src/main/java/com/codenvy/plugin/urlfactory/URLFetcher.java
```diff
@@ -0,0 +1,104 @@
...
+        } catch (IOException e) {
+            // we shouldn't fetch if check is done before
+            LOG.debug("Invalid URL", e);
+            return null;
+        }
+    }
+
+
+    /**
+     * Fetch the url provided and return its content
```
garagatyi: Connection is provided, not url

plugins/plugin-urlfactory/src/main/java/com/codenvy/plugin/urlfactory/URLFetcher.java
garagatyi: @NotNull?

plugins/plugin-urlfactory/src/main/java/com/codenvy/plugin/urlfactory/URLFetcher.java
```diff
@@ -0,0 +1,104 @@
...
+    private static final Logger LOG = LoggerFactory.getLogger(URLFetcher.class);
+
+    /**
+     * Maximum size of allowed data. (30KB)
+     */
+    protected static final long MAXIMUM_READ_BYTES = 30 * 1000;
+
+    /**
+     * Fetch the url provided and return its content
+     * To avoid DDOS, limit the amount of the collected data
```
garagatyi: Note:
I think it is connected to prevention of DOS attack. There is nothing specific to DDOS attack.

plugins/plugin-urlfactory/src/main/java/com/codenvy/plugin/urlfactory/URLFetcher.java
```diff
@@ -0,0 +1,104 @@
...
+     *
+     * @param url
+     *         the URL to fetch
+     * @return the content of the file
+     */
+    public String fetch(@NotNull final String url) {
+        requireNonNull(url, "url parameter can't be null");
+        try {
+            return fetch(new URL(url).openConnection());
```
garagatyi: Don't you think that connection should be closed?

skabashnyuk:
+ to @garagatyi . Other ok

plugins/plugin-urlfactory/src/main/java/com/codenvy/plugin/urlfactory/URLFetcher.java
benoitf: @garagatyi you want that I rephrase the comment ?

garagatyi: I don't insist, just think that term DOS suits better

plugins/plugin-urlfactory/src/main/java/com/codenvy/plugin/urlfactory/URLFetcher.java
benoitf: method can be called not only for tests (protected-->public)
about notnull sure, fixed

plugins/plugin-urlfactory/src/main/java/com/codenvy/plugin/urlfactory/URLFetcher.java
benoitf: ok, fixed

plugins/plugin-urlfactory/src/main/java/com/codenvy/plugin/urlfactory/URLFetcher.java
benoitf: Ok rephrase

plugins/plugin-urlfactory/src/main/java/com/codenvy/plugin/urlfactory/URLFetcher.java
benoitf: about closing, you want that I check for HttpUrlConnection because by default there is no close on URL/URLConnection
Also the stream is closed with AutoCloseable

plugins/plugin-urlfactory/src/main/java/com/codenvy/plugin/urlfactory/URLFetcher.java
benoitf: no problem, I've rewritten

plugins/plugin-urlfactory/src/main/java/com/codenvy/plugin/urlfactory/URLFetcher.java
garagatyi: According to UrlConnection javadocs opening connection with http protocol will use HttpUrlConnection, so you can safely cast url connection to http url connection. 
But looks like it is OK to not call disconnect. Tcp connection may be cached in that case for some period

garagatyi:
LGTM

PR #260, opened 2016-06-06
CODENVY-583: Fix JavaDoc feature

akorneta:
This pull request provides implementation for url modifier that adds machine token as a query parameter into specified url.
@skabashnyuk, @vparfonov

### DIFF
--- a/assembly/onpremises-ide-packaging-war-ext-server/pom.xml
+++ b/assembly/onpremises-ide-packaging-war-ext-server/pom.xml
@@ -47,6 +47,10 @@
             <artifactId>codenvy-hosted-workspace-activity-agent</artifactId>
         </dependency>
         <dependency>
+            <groupId>com.codenvy.hosted-infrastructure</groupId>
+            <artifactId>wsagent-codenvy</artifactId>
+        </dependency>
+        <dependency>
             <groupId>com.codenvy.plugin</groupId>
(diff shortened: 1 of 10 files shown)


plugins/plugin-java/codenvy-plugin-java-ext-lang-server/src/main/java/com/codenvy/plugin/java/server/WsAgentURLProvider.java
```diff
@@ -0,0 +1,75 @@
...
+                              HttpJsonRequestFactory requestFactory) {
+        this.wsId = wsId;
+        this.workspaceApiEndpoint = apiEndpoint + "/workspace/";
+        this.requestFactory = requestFactory;
+    }
+
+    @Override
+    public String get() {
+        try {
+            final Collection<ServerDto> servers = requestFactory.fromUrl(workspaceApiEndpoint + wsId)
```
skabashnyuk: Can calculate this value on tomcat startup? Is this wsagent side?

vparfonov:
ok

plugins/pom.xml
```diff
@@ -41,5 +41,6 @@
         <module>plugin-microsoft</module>
         <module>plugin-urlfactory</module>
         <module>plugin-product-info</module>
+        <module>../wsagent</module>
```
skabashnyuk: move to parent

skabashnyuk:
other ok

PR #261, opened 2016-06-06
CODENVY-610: add possibility to config no_proxy property

dmytro-ndp:


riuvshin:
OK

TylerJewell:
1. you probably should move the no_proxy option to be next to the http_proxy options.
2. Do we need matching no_proxy options for docker & codenvy workspaces attributes?  They are in the same section of the document.

riuvshin:
1) +0 that file is not for humans, humans should use CLI or manifests
2) we don't know the rules what we should put there, in case of TIAA to workaround their bug we should put same dns but maybe some customer will want set up no proxy for some other host which maybe reasonable.
so I would say it looks as expected.
default value is empty because we don't know what customer will want to set here.

TylerJewell:
+1  

Do we have command line parameter for bootstrap script?

dmytro-ndp:
No, we don't.

dmytro-ndp:
I could add it.

PR #263, opened 2016-06-07
CODENVY-619 Restrict usernames to letters and digits only

mkuznyetsov:
Normalize usernames that are created with OAuth, so they don't contain illegal characters.


core/codenvy-hosted-sso-auth-bearer/src/main/java/com/codenvy/auth/sso/server/BearerTokenAuthenticationService.java
```diff
@@ -127,6 +129,9 @@ public Response authenticate(@CookieParam("session-access-key") Cookie tokenAcce
         Map<String, String> payload = handler.getPayload(credentials.getToken());
         handler.authenticate(credentials.getToken());
         final String username =  payload.get("username");
+        if (isValidUserName(username)) {
+
```
sleshchenko: are you sure that you don't need to do anything here?

mkuznyetsov: fixed

skabashnyuk:
Any tests?

skabashnyuk:
ok

skabashnyuk:
ok

sleshchenko:
LGTM

PR #243, opened 2016-05-31
CHE-1078: Adapt code to user service refactorings

evoevodin:
Merge of this pull request depends on the [CHE-1078](https://github.com/eclipse/che/pull/1393).

@skabashnyuk please review

PR #268, opened 2016-06-08
Configure LdapUserDao and LdapProfileDao to use TCKs

evoevodin:
Along with tests refactoring i added an ability to use `NamingEnumeration` and `InitialLdapContext` in the _try-with-resources_ statement, which helps to avoid the boilerplate code. 

@skabashnyuk, @garagatyi please review

The merge of this PR depends on #243


core/codenvy-hosted-platform-api-impl/src/main/java/com/codenvy/api/dao/ldap/ProfileAttributesMapper.java
```diff
@@ -46,11 +46,11 @@
 @Singleton
 public class ProfileAttributesMapper {
 
-    private final String              profileDn;
-    private final String              profileIdAttr;
-    private final String              profileContainerDn;
-    private final String              profileEmailAttr;
-    private final Map<String, String> allowedAttributes;
+    final String              profileDn;
```
garagatyi: If it is needed for test only please add visible for testing annotation

core/codenvy-hosted-platform-api-impl/src/main/java/com/codenvy/api/dao/ldap/LdapUserDao.java
```diff
@@ -345,35 +321,12 @@ private String createFilter(String attribute, String value) {
     private Attributes getUserAttributesByFilter(InitialLdapContext context, String filter) throws NamingException {
         final SearchControls controls = new SearchControls();
         controls.setSearchScope(SearchControls.SUBTREE_SCOPE);
-        NamingEnumeration<SearchResult> enumeration = null;
-        try {
-            enumeration = context.search(containerDn, filter, controls);
+        try (CloseableSupplier<NamingEnumeration<SearchResult>> enumSup = deferClose(context.search(containerDn, filter, controls))) {
```
garagatyi: I don't like deferClose method name. As it is verb it says that it does something, but it does not. It just creates wrapper. So method says that context will be closed, but actually without try with resources or similar construction context won't be closed.

garagatyi:
LGTM, but please consider my comments.

core/codenvy-hosted-platform-api-impl/src/main/java/com/codenvy/api/dao/ldap/LdapUserDao.java
evoevodin: Thank you, i will consider a new name for the method.

skabashnyuk:
ok

PR #271, opened 2016-06-09
CHE-1222: add ability to share workspace with other users

ashumilova:
sharing_workspace

### DIFF
--- a/dashboard/src/app/workspace/share-workspace/share-workspace.controller.js
+++ b/dashboard/src/app/workspace/share-workspace/share-workspace.controller.js
@@ -1,0 +1,270 @@
+/*
+ *  [2015] - [2016] Codenvy, S.A.
+ *  All Rights Reserved.
+ *
+ * NOTICE:  All information contained herein is, and remains
+ * the property of Codenvy S.A. and its suppliers,
+ * if any.  The intellectual and technical concepts contained
+ * herein are proprietary to Codenvy S.A.
+ * and its suppliers and may be covered by U.S. and Foreign Patents,
+ * patents in process, and are protected by trade secret or copyright law.
+ * Dissemination of this information or reproduction of this material
+ * is strictly forbidden unless prior written permission is obtained
+ * from Codenvy S.A..
+ */
+'use strict';
+
+/**
+ * Controller for workspace sharing with other users.
+ *
+ * @ngdoc controller
+ * @name workspace.details.controller:ShareWorkspaceController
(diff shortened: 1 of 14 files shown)


dashboard/src/app/workspace/share-workspace/share-workspace.controller.js
```diff
@@ -0,0 +1,266 @@
...
+      this.isLoading = false;
+      this.cheNotification.showError(error.data && error.data.message ? error.data.message : 'Failed to remove user ' + user.email + ' permissions.');
+    });
+  }
+
+  /**
+   * Find user by its email.
+   *
+   * @param email user's email
+   * @returns {*}
```
benoitf: it doesn't seem clear to me that when we find a user, we're returning here the email (and not the 'user' object)

dashboard/src/app/workspace/share-workspace/share-workspace.controller.js
```diff
@@ -0,0 +1,266 @@
...
+  getRecipeId() {
+    let workspace = this.cheWorkspace.getWorkspaceById(this.workspaceId);
+    let environments = workspace.config.environments;
+    let defaultEnvName = workspace.config.defaultEnv;
+
+    let defaultEnvironment = this.lodash.find(environments, (environment) => {
+        return environment.name === defaultEnvName;
+    });
+
+    let recipeLocation = defaultEnvironment.machineConfigs[0].source.location;
```
benoitf: maybe now we've to use loadash as well to get the machine which is "dev" machine

dashboard/src/app/workspace/share-workspace/user-item/user-item.controller.js
```diff
@@ -0,0 +1,48 @@
...
+   */
+  constructor(codenvyPermissions) {
+    this.codenvyPermissions = codenvyPermissions;
+  }
+
+  /**
+   * Call user permissions removal.
+   */
+  removeUser() {
+    this.callback.removePermissions(this.user);
```
benoitf: maybe tell that this.callback is coming from the scope definition (as it's not in constructor)

dashboard/src/app/workspace/share-workspace/share-workspace.controller.js
ashumilova: yes will rename it and rework a little, cause in general its handling email addition and starts requesting the user exists to check

PR #272, opened 2016-06-09
CODENVY-27; remove interceptor as it is not needed anymore

mshaposhnik:
remove machine token startup time generation interceptor as it is not needed anymore.


skabashnyuk:
ok

sleshchenko:
ok

sleshchenko:
ok

PR #273, opened 2016-06-09
Fix maven groupid for wsagent artifact

akorneta:
During assembly wsagent artifact, warnings occurs, to avoid that we shoud use correct group id for ws agent parent and wsagent codenvy artifacts.
@skabashnyuk


skabashnyuk:
ок

PR #274, opened 2016-06-09
Replace codenvy-dev.com SaaS server on a1.codenvy-dev.com

dmytro-ndp:
Replace codenvy-dev.com SaaS server on a1.codenvy-dev.com in integration tests.


riuvshin:
OK

PR #275, opened 2016-06-10
CODENVY-635: Create attachment files in separate temp directory

vinokurig:
This will resolve problem when two users send request for sending email at the same time, and sending second email fails.
@skabashnyuk please review

### DIFF
--- a/core/codenvy-hosted-mail-server/pom.xml
+++ b/core/codenvy-hosted-mail-server/pom.xml
@@ -43,6 +43,10 @@
             <artifactId>guice-servlet</artifactId>
         </dependency>
         <dependency>
+            <groupId>commons-io</groupId>
+            <artifactId>commons-io</artifactId>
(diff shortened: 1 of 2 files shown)


benoitf:
should we have a test that reproduce race condition and then ensure it's gone and won't occur anymore ?

vinokurig:
The problem is that `tempDir` is a local variable, I ca not mock it

vinokurig:
It is very hard to reproduce this case in test conditions to test ti without mocks

benoitf:
I see. I like tdd which means that before fixing it I'm able to have a testcase then I can change behavior and check test is fixed. So if class is not testable, for me it's a design error
And when bug is discovered we should have test case to ensure issue will not come back
But I see your point as well that current code is not testable it's just that it's not my way to think

skabashnyuk:
ок

PR #281, opened 2016-06-11
CODENVY-446: add scripts and tools to update Codenvy 4.2.x to 4.3.0

dmytro-ndp:


riuvshin:
Please do not put jars here, we should take it as a dependencies or move that code here and put jars on packaging phase.

dmytro-ndp:
In this case someone should add next artifacts into the maven-depmgt-pom:
>         <dependency>
            <groupId>com.codenvy.hosted</groupId>
            <artifactId>ldap-tools</artifactId>
            <version>1.0</version>
        </dependency>
        <dependency>
            <groupId>com.codenvy.hosted</groupId>
            <artifactId>namespace-tool</artifactId>
            <version>1.0</version>
        </dependency>"

dmytro-ndp:
I have removed jars from the request. They will be added into the bundle manually.

PR #288, opened 2016-06-13
fix typo, make file not binary

riuvshin:
there was some weird special chars which make sh file binary.
@dmytro-ndp @skabashnyuk


skabashnyuk:
ok

PR #293, opened 2016-06-14
CODENVY-653: fix error of logging event by using Update Server

dmytro-ndp:
@sleshchenko: please, review your changes which I just committed.


sleshchenko:
OK as temporary solution

PR #299, opened 2016-06-16
CODENVY-547: fix error of changing of non-default admin password

dmytro-ndp:


riuvshin:
+1

PR #306, opened 2016-06-17
fix UD transitive dependency of gulp-angular-templatecache

olexii4:


riuvshin:
+1

PR #307, opened 2016-06-17
CHE-1301: add not modified condition to the machine token response

olexii4:


akorneta:
ok

PR #313, opened 2016-06-22
CODENVY-309: check subscription of RHEL when installing Codenvy on-prem

dmytro-ndp:

### DIFF
--- a/cdec/installation-manager-resources/src/main/resources/im-scripts/install-codenvy.sh
+++ b/cdec/installation-manager-resources/src/main/resources/im-scripts/install-codenvy.sh
@@ -99,6 +99,8 @@ DEPENDENCIES_STATUS_OFFSET=85  # fit screen width = 100 cols
 PROGRESS_FACTOR=2
 URL_REGEX="^https?:\/\/[\da-zA-Z_\-\.\/?&#+]+"
 
+MIN_OS_VERSION=7.1
+
 cleanUp() {
     setterm -cursor on
     killTimer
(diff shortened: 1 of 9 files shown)


cdec/installation-manager-resources/src/main/resources/im-scripts/install-codenvy.sh
```diff
@@ -1524,8 +1561,8 @@ doCheckAvailableResourcesOnNodes() {
                 ;;
         esac
 
-        # check on OS CentOS 7
-        if [[ "${osType}" != "CentOS" || "7.1" > "${osVersion}" ]]; then
+        # check on CentOS or RHEL OS of version >= 7.1
+        if [[ "${osType}" != "CentOS" || "${osVersion}" < "${MIN_OS_VERSION}" ]]; then
```
benoitf: here we don't check RHEL ? (comment is saying checking RHEL)
also as we already have a check lines before
should we have a function for this test to avoid duplicates ?

dmytro-ndp: > here  we don't check RHEL ? (comment is saying checking RHEL)
In that case we are checking both RHEL OS and CentOS as well, because of command: `if [[ -f /etc/redhat-release ]]`.

> should we have a function for this test to avoid duplicates ?
we could, but there is only one duplication, so I want to have the code as clear as possible.

riuvshin: +1

riuvshin:
OK

cdec/installation-manager-resources/src/main/resources/im-scripts/install-codenvy.sh
benoitf: as discussed, checking /etc/redhat-release should imply that osType is named RHEL, not centos

then the check osType != "RHEL" is meaningful for bot RHEL (and derivatives like Centos, etc)

PR #314, opened 2016-06-22
Fix default properties of Codenvy 4.4.0 on-prem

dmytro-ndp:


riuvshin:
OK

PR #317, opened 2016-06-22
Rework AdminUserServiceTests with everrest assured

sleshchenko:
These changes are needed to avoid directly binding ApiExceptionMapper
that produces exception in other tests with everrest assured

@skabashnyuk Please review my changes


skabashnyuk:
ok

PR #319, opened 2016-06-22
Codenvy 694 Fix limits checking user on ws startup

mshaposhnik:
Fix limits checking user on ws startup

### DIFF
--- a/core/codenvy-hosted-platform-api-impl/src/main/java/com/codenvy/api/workspace/LimitsCheckingWorkspaceManager.java
+++ b/core/codenvy-hosted-platform-api-impl/src/main/java/com/codenvy/api/workspace/LimitsCheckingWorkspaceManager.java
@@ -62,6 +62,8 @@ public class LimitsCheckingWorkspaceManager extends WorkspaceManager {
     private static final Striped<Lock> CREATE_LOCKS   = Striped.lazyWeakLock(100);
     private static final Striped<Lock> START_LOCKS    = Striped.lazyWeakLock(100);
 
+    private final UserManager userManager;
+
     private final int  workspacesPerUser;
     private final long maxRamPerEnv;
     private final long ramPerUser;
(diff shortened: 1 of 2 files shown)


mshaposhnik:
@skabashnyuk @evoevodin

benoitf:
if you could use a more descriptive text for the PR name

core/codenvy-hosted-platform-api-impl/src/main/java/com/codenvy/api/workspace/LimitsCheckingWorkspaceManager.java
```diff
@@ -111,9 +115,15 @@ public WorkspaceImpl startWorkspace(String workspaceId,
                                                                            ServerException,
                                                                            ConflictException {
         final WorkspaceImpl workspace = getWorkspace(workspaceId);
+        String userId;
+        try {
+            userId = userManager.getByName(workspace.getNamespace()).getId();
+        } catch (NotFoundException e) {
+            userId = getCurrentUserId();
```
evoevodin: Workspace start behaviour must be strict and always predictable(the same in this case), i think that you should prohibit workspace start if there is no user owning the workspace found @sleshchenko, @skabashnyuk can you please take a look

sleshchenko:
-1 for now

As far as I understand We also should fix fetching of workspaces here
https://github.com/codenvy/codenvy/pull/319/files/5a6ff77ba5a1ae64b4b6d9c0dcefc8d629ac0487#diff-8b6273ced2f7db76f99f8877a8ae652cL177
getWorkspaces returns all workspaces which are available for given user (in fact read action)
but we should fetch all workspaces by namespace of workspace that will start

core/codenvy-hosted-platform-api-impl/src/main/java/com/codenvy/api/workspace/LimitsCheckingWorkspaceManager.java
```diff
@@ -127,8 +138,8 @@ public WorkspaceImpl startWorkspace(WorkspaceConfig config,
         checkMaxEnvironmentRam(config);
         return checkRamAndPropagateStart(config,
                                          config.getDefaultEnv(),
-                                         getCurrentUserId(),
-                                         () -> super.startWorkspace(config, namespace, isTemporary, accountId));
+                                         EnvironmentContext.getCurrent().getSubject().getUserName(),
```
sleshchenko: should we put namespace here?

mshaposhnik: Yes seems we can

sleshchenko: If yes, seems that tests don't cover this usecase. Can you cover this?

core/codenvy-hosted-platform-api-impl/src/main/java/com/codenvy/api/workspace/LimitsCheckingWorkspaceManager.java
```diff
@@ -111,6 +114,14 @@ public WorkspaceImpl startWorkspace(String workspaceId,
                                                                            ServerException,
                                                                            ConflictException {
         final WorkspaceImpl workspace = getWorkspace(workspaceId);
+        try {
+            userManager.getByName(workspace.getNamespace());
+        } catch (NotFoundException e) {
+            throw new ServerException(String.format(
```
sleshchenko: Are you sure that we need this check? then why we don't need it here https://github.com/codenvy/codenvy/pull/319/files#diff-8b6273ced2f7db76f99f8877a8ae652cR138

sleshchenko:
OK for me

skabashnyuk:
ok

PR #325, opened 2016-06-24
Switch to temporary address updater-nightly2.codenvy-dev.com

dmytro-ndp:
@skryzhny: please, review this request.


mumbozver:
OK for me

PR #332, opened 2016-06-27
CHE-1396 Double nav bar in some cases

vitaliy-guliy:
Signed-off-by: Vitaliy Guliy <vguliy@codenvy.com>


riuvshin:
Thanks @vitaliy-guliy !

vparfonov:
ok

PR #335, opened 2016-06-29
CODENVY-702: fix command which hangs up when changing Codenvy hostname

dmytro-ndp:


riuvshin:
OK

PR #337, opened 2016-06-29
CODENVY-84: work around default node of Codenvy on-prem

dmytro-ndp:


cdec/installation-manager-cli-bundle/src/main/java/com/codenvy/im/cli/command/AddNodeCommand.java
```diff
@@ -78,8 +79,13 @@ protected void doExecuteCommand() throws Exception {
     protected void validateCodenvyConfig() throws IOException {
         if (isCodenvy4Installed()) {
             Config config = configManager.loadInstalledCodenvyConfig();
-            String property = config.getProperties().get(Config.MACHINE_EXTRA_HOSTS);
 
+            // check on adding default node
+            if (NodeConfigHelper.isDefaultNode(config, dns)) {
```
tolusha: It is better to avoid invoking `validateCodenvyConfig` method in case of DefaultNode instead of skipping validation inside the method.

dmytro-ndp: I hate to disagree, but *isDefaultNode* method is needed for codenvy 4 only, and it requires *config* object.

cdec/installation-manager-core/src/main/java/com/codenvy/im/managers/helper/NodeConfigHelper.java
```diff
@@ -77,25 +83,29 @@ public static String getProperty(NodeConfig.NodeType type) {
...
-     * Read all urls of additional nodes stored from the puppet master config, find out node with certain dns and return type of additional node with
-     * certain dns.
+     * Read all urls of nodes stored from the puppet master config, find out node with certain dns and return type of node with certain dns.
      * For example (Codenvy3): given:
      * $additional_builders = "http://builder2.example.com:8080/builder/internal/builder,http://builder3.example.com:8080/builder/internal/builder"
      * dns = "builder3.example.com"
      * Result = BUILDER
      */
     @Nullable
     public NodeConfig.NodeType recognizeNodeTypeFromConfigBy(String dns) {
```
tolusha: `recognizeNodeTypeFromConfigByDns`

dmytro-ndp: agree

PR #342, opened 2016-07-01
CODENVY-706: check if there is subscription-manager command in RHEL OS

dmytro-ndp:
CODENVY-706: check if there is subscription-manager command in RHEL OS when checking subscription in Codenvy bootstrap script.


riuvshin:
+1

PR #347, opened 2016-07-04
CODENVY-520 Remove allow_anonymous from war configuration

mkuznyetsov:
remove, as it is being configured in deployment


skabashnyuk:
ok

skabashnyuk:
ok

PR #350, opened 2016-07-05
CHE-1248 remove workspace information from EnvironmentContext

mshaposhnik:


skabashnyuk:
ok

musienko-maxim:
After checking all tests we do not have any regressions as compared with nightly4 server.

PR #352, opened 2016-07-05
CODENVY-720: Fix snapshot recovering

garagatyi:
Swarm changed output of API, so I've added workaround to be able to use that code with newest and older versions of swarm.

@riuvshin @skabashnyuk @evoevodin @mmorhun Please review


evoevodin:
The same problem again =)
lgtm

PR #356, opened 2016-07-07
CODENVY-651 Use 'dev' instead 'isDev' for machine configs in Mongo

mkuznyetsov:
change MachineConfigImpl field 'isDev' to 'dev', as it is done in MachineConfigDto.
It should fix consitensy problems.

Also it needs to make migrations of existing workspaces.


skabashnyuk:
ok. @evoevodin ?

evoevodin:
ok

skabashnyuk:
ok. Do not merge until 4.6

musienko-maxim:
After checking all tests we do not have any regressions as compared with nightly4 server.

PR #415, opened 2016-07-19
Use Model.gwt.xml see https://github.com/eclipse/che/pull/1869

vparfonov:
Signed-off-by: Vitaly Parfonov <vparfonov@codenvy.com>


skabashnyuk:
ок

PR #372, opened 2016-07-13
CODENVY-413: Make it possible to mark MachineNode as 'sheduled for maintenace'

mmorhun:
This and [this](https://github.com/eclipse/che/pull/1705) PRs add ability to mark node for maintenance (it means that new containers won't start on this node).
To mark node for maintenance just add com.codenvy.node-state=maintenance label to docker on the node.
To return node to normal state just remove the label.

To add label to docker:
1. edit docker config file, add label
2. reload (without restart!) docker config (use SIGHUP signal)
Note, that it can take about a minute or so while swarm "see" the label.
To remove label just erase it from config and reload it again.

PR #418, opened 2016-07-20
CODENVY-413: Make it possible to mark MachineNode as 'sheduled for maintenace'

mmorhun:
This PR add ability to mark node for maintenance (it means that new containers won't start on this node).
- To mark node for maintenance just add `com.codenvy.node-state=maintenance` label to docker on the node.
- To return node to normal state just remove the label.

To add label to docker:
1. edit docker config file, add label
2. reload (without restart!) docker config (use SIGHUP signal)
Note, that it can take about a minute or so while swarm "see" the label.
To remove label just erase it from config and reload it again.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -55,5 +55,5 @@ node_modules
 ehthumbs.db
 Icon?
 Thumbs.db
-*/overlays
+overlays/
 *~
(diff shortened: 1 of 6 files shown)


mmorhun:
@garagatyi take a look please

mmorhun:
This PR was reviewed and merged [here](https://github.com/codenvy/codenvy/pull/372) but was reverted because it did not planned for release.

PR #441, opened 2016-07-23
Updated to support 4.4.x proxy installation

TylerJewell:
1. Added in the appropriate new proxy installation options.
2. Migrated some variables to be configurable

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -6,9 +6,9 @@ cpus            = '2'
 #bridge          = 'eth0'
 
 # Set to "<proto>://<user>:<pass>@<host>:<port>"
-http_proxy      = ENV['HTTP_PROXY'] || ""
-https_proxy     = ENV['HTTPS_PROXY'] || ""
-no_proxy        = ENV['NO_PROXY'] || "codenvy.onprem,localhost,127.0.0.1"
+http_proxy      = ENV['HTTP_PROXY'] || ENV['http_proxy'] || ""
+https_proxy     = ENV['HTTPS_PROXY'] || ENV['https_proxy'] || ""
+no_proxy        = ENV['NO_PROXY'] || ENV['no_proxy'] || "codenvy.onprem,localhost,127.0.0.1"
 
 ip              = ENV['CODENVY_IP'] || "192.168.56.110"
 
@@ -73,6 +73,7 @@ Vagrant.configure("2") do |config|
       echo "."
       echo "HTTP PROXY set to: $HTTP_PROXY"
       echo "HTTPS PROXY set to: $HTTPS_PROXY"
+      echo "NO PROXY set to: $NO_PROXY"
     fi
 
     echo "."
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -87,8 +87,8 @@ Vagrant.configure("2") do |config|
     echo "."
     echo "."
     echo "CODENVY: INSTALLED!"
-    echo 'Add "$IP codenvy.onprem" to your hosts file'
-    echo 'Access:   http://codenvy.onprem'
+    echo 'Add       "192.68.56.110 codenvy.onprem" to your hosts file'
+    echo 'Browser:  http://codenvy.onprem'
     echo 'Username: admin'
     echo 'Password: password'
     echo "."
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -8,7 +8,7 @@ cpus            = '2'
 # Set to "<proto>://<user>:<pass>@<host>:<port>"
 http_proxy      = ENV['HTTP_PROXY'] || ""
 https_proxy     = ENV['HTTPS_PROXY'] || ""
-no_proxy        = ENV['NO_PROXY'] || "localhost,127.0.0.1"
+no_proxy        = ENV['NO_PROXY'] || "codenvy.onprem,localhost,127.0.0.1"
 
 ip              = ENV['CODENVY_IP'] || "192.168.56.110"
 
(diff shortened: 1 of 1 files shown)


Vagrantfile
```diff
@@ -54,23 +77,18 @@ Vagrant.configure("2") do |config|
     echo "."
 
     if [ -n "$HTTP_PROXY" ] || [ -n "$HTTPS_PROXY" ]; then
-      bash <(curl -L -s --proxy ${HTTP_PROXY} ${CODENVY_URL}) ${CODENVY_OPTIONS} --http-proxy=${HTTP_PROXY} --https-proxy=${HTTPS_PROXY}
+      bash <(curl -L -s --proxy ${HTTP_PROXY} ${CODENVY_URL}) ${CODENVY_OPTIONS} \
+                                                              --http-proxy-for-installation=${HTTP_PROXY} \
+                                                              --https-proxy-for-installation=${HTTPS_PROXY}
     else
```
dmytro-ndp: What about adding here:
`--no-proxy-for-installation=${NO_PROXY}`

TylerJewell: Yes, good suggestion.  Will add that.

Vagrantfile
```diff
@@ -3,19 +3,41 @@ url             = 'https://install.codenvycorp.com/centos7.2.box'
 
 ram             = '4096'
 cpus            = '2'
-bridge          = 'eth0'
+#bridge          = 'eth0'
+
+# Set to "<proto>://<user>:<pass>@<host>:<port>"
+http_proxy      = ENV['HTTP_PROXY'] || ""
+https_proxy     = ENV['HTTPS_PROXY'] || ""
+no_proxy        = ENV['NO_PROXY'] || "localhost,127.0.0.1"
```
dmytro-ndp: It may be reasonable to add codenvy host url to the list of no-proxy dns, that is:
`codenvy.onprem,localhost,127.0.0.1`

Vagrantfile
```diff
@@ -79,7 +97,7 @@ Vagrant.configure("2") do |config|
 
   config.vm.provision "shell" do |s|
     s.inline = $script
-    s.args = [http_proxy, https_proxy, codenvy_url, codenvy_options]
+    s.args = [http_proxy, https_proxy, codenvy_url, codenvy_options, ip]
```
dmytro-ndp: It seems we missed **no_proxy** here in the list of args.

Vagrantfile
```diff
@@ -54,23 +77,18 @@ Vagrant.configure("2") do |config|
...
-    echo "."
-    echo "."
     echo "CODENVY: INSTALLED!"
-    echo 'Add "192.168.56.110 codenvy" to your hosts file'
-    echo 'Access:   http://codenvy'
+    echo 'Add "$IP codenvy.onprem" to your hosts file'
+    echo 'Access:   http://codenvy.onprem'
     echo 'Username: admin'
     echo 'Password: password'
     echo "."
```
dmytro-ndp: Link to [installation-troubleshooting page](http://codenvy.readme.io/docs/installation-troubleshooting) here can be  helpful in some cases.

Vagrantfile
```diff
@@ -3,19 +3,41 @@ url             = 'https://install.codenvycorp.com/centos7.2.box'
...
 codenvy_options = "--suppress --silent --license=accept"
 
 Vagrant.configure("2") do |config|
-  config.vm.box = box
-  config.vm.box_url = url
+  puts ("CODENVY: VAGRANT INSTALLER")
+  puts ("CODENVY: REQUIRED: VIRTUALBOX 5.x")
+  puts ("CODENVY: REQUIRED: VAGRANT 1.8.x")
+  puts ("")
+  if $http_proxy.to_s != '' || https_proxy.to_s != '') && !Vagrant.has_plugin?("vagrant-proxyconf")
```
dmytro-ndp: There is open bracket missed before **$http_proxy.to_s**.

Vagrantfile
```diff
@@ -54,23 +77,18 @@ Vagrant.configure("2") do |config|
...
-    echo "."
-    echo "."
-    docker pull codenvy/ubuntu_jdk8
-
-    echo "."
-    echo "."
     echo "CODENVY: INSTALLED!"
-    echo 'Add "192.168.56.110 codenvy" to your hosts file'
-    echo 'Access:   http://codenvy'
+    echo 'Add       "192.68.56.110 codenvy.onprem" to your hosts file'
```
dmytro-ndp: Correct default IP address is *192.168.56.110* here. Moreover, *$ip* is more preferable here so as there could **$CODENVY_IP** be used instead of default one, according to the command above:
`ip              = ENV['CODENVY_IP'] || "192.168.56.110"`

Vagrantfile
TylerJewell: Ok.

Vagrantfile
TylerJewell: Fixed in commit coming soon.

TylerJewell:
@dmytro-ndp - updated all changes per your feedback.

dmytro-ndp:
@TylerJewell: could you, please, explain, how can I check if proxy settings handles properly by vagrant?
Setting http_proxy, https_proxy, no_proxy variables in shell environment before executing command *vagrant up* seems doesn't affect on commands:
```
http_proxy      = ENV['HTTP_PROXY'] || ""
https_proxy     = ENV['HTTPS_PROXY'] || ""
no_proxy        = ENV['NO_PROXY'] || "codenvy.onprem,localhost,127.0.0.1"
```

dmytro-ndp:
I found the reasons of problem above - I have proxy settings name in low case, and here is a solution:
```
http_proxy      = ENV['HTTP_PROXY'] || ENV['http_proxy'] || ""
https_proxy    = ENV['HTTPS_PROXY'] || ENV['https_proxy'] || ""
```

dmytro-ndp:
Also it could be useful to display NO PROXY value as well:
```
      echo "."
      echo "."
      echo "HTTP PROXY set to: $HTTP_PROXY"
      echo "HTTPS PROXY set to: $HTTPS_PROXY"

      if [ -n "$NO_PROXY" ]; then
        echo "NO PROXY HOSTS set to: '$NO_PROXY'"
      fi
```
to have in output something like the follow:

> ==> default: HTTP PROXY set to: http://testuser:testpassword@proxy.ua.codenvy-dev.com:3128
> ==> default: HTTPS PROXY set to: http://testuser:testpassword@proxy.ua.codenvy-dev.com:3128
> ==> default: NO PROXY HOSTS set to: codenvy.onprem,localhost,127.0.0.1

dmytro-ndp:
That's all.

dmytro-ndp:
Everything seems is good.
+1

Issue #411, opened 2016-07-19
Investigate exception  'Machine with name ws-machine already exists'


vkuznyetsov:
It needs to investigate exception  'Machine with name ws-machine already exists'. 

We have the exception in logs on beta.codenvy.com:

`2016-05-26 21:18:31,565[paceManager-331]  [INFO ] [o.e.c.a.m.s.MachineManager 856]      - Destroying machine [ws = workspaceqefgkxuq449bmytr: env = wksp-8bcx: machine name = ws-machine: machine id = machinehnq2j5ffa45u0h4v]
2016-05-26 21:18:32,757[paceManager-332]  [INFO ] [o.e.c.a.m.s.MachineManager 162]      - Creating machine [ws = workspacel6x1tnsph1q2wf6l: env = default: machine = ws-machine]
2016-05-26 21:18:32,763[paceManager-332]  [ERROR] [o.e.c.a.w.s.WorkspaceManager 538]    - Machine with name ws-machine already exists
org.eclipse.che.api.core.ConflictException: Machine with name ws-machine already exists
        at org.eclipse.che.api.machine.server.MachineManager.createMachine(MachineManager.java:314) ~[che-core-api-machine-4.2.3.jar:4.2.3]
        at org.eclipse.che.api.machine.server.MachineManager.createMachineSync(MachineManager.java:166) ~[che-core-api-machine-4.2.3.jar:4.2.3]
        at org.eclipse.che.api.workspace.server.WorkspaceRuntimes.createMachine(WorkspaceRuntimes.java:490) ~[che-core-api-workspace-4.2.3.jar:4.2.3]
        at org.eclipse.che.api.workspace.server.WorkspaceRuntimes.startMachine(WorkspaceRuntimes.java:411) ~[che-core-api-workspace-4.2.3.jar:4.2.3]
        at org.eclipse.che.api.workspace.server.WorkspaceRuntimes.startQueue(WorkspaceRuntimes.java:368) ~[che-core-api-workspace-4.2.3.jar:4.2.3]
        at org.eclipse.che.api.workspace.server.WorkspaceRuntimes.start(WorkspaceRuntimes.java:180) ~[che-core-api-workspace-4.2.3.jar:4.2.3]
        at org.eclipse.che.api.workspace.server.WorkspaceManager.lambda$performAsyncStart$2(WorkspaceManager.java:522) ~[che-core-api-workspace-4.2.3.jar:4.2.3]
[...]

PR #446, opened 2016-07-26
CODENVY-411: Fix bug machine with name ws-machine already exists

mmorhun:
### What does this PR do?
Fix NPE in `SwarmDockerConnector`, which broke catch chain on create machine error

[Fix](https://github.com/codenvy/codenvy/issues/411) wrong "Machine with name _machine-name_ already exists" error message and fix the cause of next fails of this workspace to start

### Previous Behavior
Sometimes shows useless error message  if creation of a machine fails and breaks next workspace starts

### New Behavior
Displays actual error message instead of "Machine with name _machine-name_ already exists"

### Tests written?
No


mmorhun:
@garagatyi take a look please

garagatyi:
LGTM

mmorhun:
@skabashnyuk ?

skabashnyuk:
ok

PR #448, opened 2016-07-26
CHE-1849 Fix according to changes in Workspace API

vinokurig:
### What does this PR do?
Fixes tests and other classes according to changes in Workspace API

https://github.com/eclipse/che/issues/1849

@evoevodin @sleshchenko please review

### DIFF
--- a/core/codenvy-hosted-api-workspace/src/main/java/com/codenvy/api/workspace/server/filters/WorkspacePermissionsFilter.java
+++ b/core/codenvy-hosted-api-workspace/src/main/java/com/codenvy/api/workspace/server/filters/WorkspacePermissionsFilter.java
@@ -79,7 +79,6 @@ public class WorkspacePermissionsFilter extends CheMethodInvokerFilter {
                 workspaceId = ((String)arguments[0]);
                 action = DELETE;
                 break;
-            case "recoverWorkspace":
             case "createMachine":
             case "stop":
             case "startById":
--- a/core/codenvy-hosted-platform-api-impl/src/main/java/com/codenvy/api/workspace/LimitsCheckingWorkspaceManager.java
+++ b/core/codenvy-hosted-platform-api-impl/src/main/java/com/codenvy/api/workspace/LimitsCheckingWorkspaceManager.java
@@ -107,15 +107,16 @@ public class LimitsCheckingWorkspaceManager extends WorkspaceManager {
         checkMaxEnvironmentRam(config);
         checkNamespaceValidity(namespace, "Unable to create workspace because its namespace owner is " +
                                           "unavailable and it is impossible to check resources limit.");
-        return checkCountAndPropagateCreation(namespace, () -> super.createWorkspace(config, namespace, accountId));
+        return checkCountAndPropagateCreation(namespace, () -> super.createWorkspace(config, namespace, attributes, accountId));
     }
 
     @Override
     public WorkspaceImpl startWorkspace(String workspaceId,
                                         @Nullable String envName,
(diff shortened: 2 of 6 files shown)


core/codenvy-hosted-platform-api-impl/src/main/java/com/codenvy/api/workspace/LimitsCheckingWorkspaceManager.java
```diff
@@ -124,7 +125,7 @@ public WorkspaceImpl startWorkspace(String workspaceId,
         return checkRamAndPropagateStart(workspace.getConfig(),
                                          envName,
                                          workspace.getNamespace(),
-                                         () -> super.startWorkspace(workspaceId, envName, accountId));
+                                         () -> super.startWorkspace(workspaceId, envName, accountId, null));
```
sleshchenko: why null? Should not be restore variable there?

sleshchenko:
Should we remove "recoverWorkspace" here https://github.com/codenvy/codenvy/blob/ab81ba6da94485a21abe9fa0408bd7fa0eca0f0c/core/codenvy-hosted-api-workspace/src/main/java/com/codenvy/api/workspace/server/filters/WorkspacePermissionsFilter.java#L82 ?

sleshchenko:
other OK for me

core/codenvy-hosted-platform-api-impl/src/main/java/com/codenvy/api/workspace/LimitsCheckingWorkspaceManager.java
vinokurig: done

vinokurig:
Removed also interceptor bindings on 'recoverWorkspace'

core/codenvy-hosted-api-workspace/src/main/java/com/codenvy/api/workspace/server/filters/WorkspacePermissionsFilter.java
```diff
@@ -79,7 +79,6 @@ public void filter(GenericResourceMethod genericResourceMethod, Object[] argumen
                 workspaceId = ((String)arguments[0]);
                 action = DELETE;
                 break;
-            case "recoverWorkspace":
```
evoevodin: @sleshchenko is there any common mechanism which checks whether such methods exist?
I think this code must be covered by tests, as we usually have troubles in places like this after refactorings. 
@vinokurig consider adding such tests

sleshchenko: What about https://github.com/codenvy/codenvy/pull/448/files/5a4a3a9ff8769b39a48d21dabedac33cb45aa167#diff-15c913e41deb296c99e1a0ffcb8a5937L141 ?
Do you mean this mechanism?

evoevodin: I think this is pretty similar to interceptors compatibility tests like [this one](https://github.com/codenvy/customer-saas/blob/98fdeece736cebf6da844f3645b44cad1bbfa2b8/wsmaster/saas-marketo/src/test/java/com/codenvy/saas/marketo/UpdateProfileInterceptorTest.java). The main point here is that if intercepted class doesn't contain such method build MUST fail.

evoevodin:
Ok

skabashnyuk:
ok

Issue #449, opened 2016-07-26
VSTS projects does cloning failed


mshaposhnik:
VSTS projects does not cloned anymore

Reproduction Steps:
Try clone project https://iedexmain1.visualstudio.com/_git/MyFirstProject

Expected behavior:
Project is  cloned

Observed behavior:
Error message appears 

Codenvy version: 4.6.0-SHANPSHOT

Additional information:
URL matching regexp changes - need to fix it

mshaposhnik:
Fixed. URL' s now may or may not contains collection name, e.g.
https://che-1.visualstudio.com/DefaultCollection/test/_git/che-core
or
https://iedexmain1.visualstudio.com/_git/MyFirstProject

PR #450, opened 2016-07-26
CODENVY-449; fix VSTS project URL matching regexp

mshaposhnik:
### What does this PR do?
Fixes cloning projects from VSTS

Now VSTS URL's may or may not contains the collection name, e.g.:
with collection:
https://che-1.visualstudio.com/DefaultCollection/test/_git/che-core

without collection:
https://iedexmain1.visualstudio.com/_git/MyFirstProject
#449 

### Tests written?
No

### Docs requirements?
None


skabashnyuk:
Describe what kind of url we are supporting from now?
What was changed? 
Other ok

PR #486, opened 2016-07-29
CHE-1989 Fix permissions checking in according to new namespace concept

sleshchenko:
### What does this PR do?
- Adapt source code to changes of [Account API](https://github.com/eclipse/che/pull/1990)
- Rework permissions checking in according to new namespace concept

https://github.com/eclipse/che/issues/1989

### Previous Behavior
Earlier filter checked permissions only by subject to wit by link workspace and user with worker.

### New Behavior
User will be able to do anythink with workspace if it has namespace that equals to user's name

### Tests written?
Yes


sleshchenko:
@skabashnyuk @evoevodin @akorneta please review my changes

Issue #496, opened 2016-08-01
Release and ship codenvy 4.6.0


riuvshin:
Release and ship codenvy 4.6.0

- [x] release codenvy
- [x] ship codenvy onprem

PR #497, opened 2016-08-01
(#496) add Codenvy on-prem 4.6.0 properies and migration script

dmytro-ndp:


riuvshin:
OK

Issue #408, opened 2016-07-19
Prevent workspace starting if the system achieved resources limit.


vkuznyetsov:
**Epic link** [(https://github.com/codenvy/codenvy/issues/406)]

If the system has achieved 90% of the allowed disk threshold OR 90% of total RAM available from all running workspaces, then the system:

1. Prevent any new workspaces from starting for any user. When doing this, we should print friendly error message to user before they attempt to start a workspace.
2. Have banner displayed to all users / admins indicating that the system is limiting new workspaces from running due to reaching max resources.

We need this to work properly for both UD-driven and factory-driven scenarios.

TylerJewell:
@vkuznyetsov - please use markdown to format issues appropriately.  Markdown has bold, bullets, and numbering. The issue readability is important.  Also, paragraph spacing is important.

vkuznyetsov:
![banneride](https://cloud.githubusercontent.com/assets/2101815/16950133/4a7989b0-4dc7-11e6-8604-8bf3f8f50dfb.png)
![bannerud](https://cloud.githubusercontent.com/assets/2101815/16950137/50e05568-4dc7-11e6-86a1-bd15737cd29e.png)
![notificationide](https://cloud.githubusercontent.com/assets/2101815/16950139/5333cc5a-4dc7-11e6-9604-c2df3df4e393.png)
![notificationud](https://cloud.githubusercontent.com/assets/2101815/16950141/5515b326-4dc7-11e6-8225-53ca055448ad.png)

vinokurig:
Docker swarm does not provide any information about disc space of all nodes in the system, so only RAM usage will be checked.

TylerJewell:
@vkuznyetsov do not merge this without a detailed UX review.  My first impressions was that there are grammar issues and formatting issues.

TylerJewell:
Changes:
1: Will we be supporting multiple banners on the left if there are multiple system events?  It does not seem to be very scalable.  If we have 6 permanent alerts that we must provide to the user / admin, we cannot fit 6 left bar items.  We should consider having a thin bar on top of the UD + IDE area that spans the full width of left-to-right.

2: Permanent message should be.  "Low RAM".  The system is low on RAM. New workspaces will not start until more RAM is available."

3: Notification message.  "Low RAM." Your workspace cannot be started until the system has more RAM available."

vinokurig:
Items 1 and 2 are related to https://github.com/codenvy/codenvy/issues/502.
[...]

PR #501, opened 2016-08-02
CODENVY-408: Add System RAM check

vinokurig:
### What does this PR do?
Checks whole system ram before starting workspace and if it is more than 90% of used ram workspace will not be started. When it will be less then 90% of used ram, a message will be sent by websocket.

https://github.com/codenvy/codenvy/issues/408

### Tests written?
Yes

@garagatyi @evoevodin please review


core/codenvy-hosted-platform-api-impl/src/main/java/com/codenvy/api/workspace/LimitsCheckingWorkspaceManager.java
```diff
@@ -297,4 +372,53 @@ private void checkNamespaceValidity(String namespace, String errorMsg) throws Se
...
+        bm.setBody(JsonUtils.getJsonString(line));
+        try {
+            WSConnectionContext.sendMessage(bm);
+        } catch (Exception e) {
+            LOG.error("A problem occurred while sending websocket message", e);
+        }
+    }
+
+    private long getBytesAmountFromString(String string) {
+        if (string.contains("KiB")) {
```
evoevodin: Can you give an example of the output?
It seems that you can reuse: [this class](https://github.com/eclipse/che/blob/867ec652424cd23a587f164f6acf3e8e44158bdf/core/commons/che-core-commons-lang/src/main/java/org/eclipse/che/commons/lang/Size.java#L22-L22) for parsing

core/codenvy-hosted-platform-api-impl/src/main/java/com/codenvy/api/workspace/LimitsCheckingWorkspaceManager.java
```diff
@@ -297,4 +372,53 @@ private void checkNamespaceValidity(String namespace, String errorMsg) throws Se
...
+            ramUsage = dockerConnector.getSystemInfo().ramUsage();
+        } catch (IOException e) {
+            LOG.error("A problem occurred while getting system information from docker", e);
+        }
+        if (ramUsage == null) {
+            return false;
+        }
+        String[] ramValues = ramUsage.split("/ ");
+        if (ramValues.length < 2) {
+            LOG.error("A problem occurred while parsing system information from docker");
```
evoevodin: Be more concrete here, it is not clear from the context what value was and what exactly was wrong, can you please add a `ramUsage` value to the message and also add what we expected.

core/codenvy-hosted-platform-api-impl/src/main/java/com/codenvy/api/workspace/LimitsCheckingWorkspaceManager.java
```diff
@@ -297,4 +372,53 @@ private void checkNamespaceValidity(String namespace, String errorMsg) throws Se
...
+        try {
+            ramUsage = dockerConnector.getSystemInfo().ramUsage();
+        } catch (IOException e) {
+            LOG.error("A problem occurred while getting system information from docker", e);
+        }
+        if (ramUsage == null) {
+            return false;
+        }
+        String[] ramValues = ramUsage.split("/ ");
+        if (ramValues.length < 2) {
```
evoevodin: Are you sure that it is okay, if it is `>2` i mean different from `2` ?

core/codenvy-hosted-platform-api-impl/src/main/java/com/codenvy/api/workspace/LimitsCheckingWorkspaceManager.java
```diff
@@ -297,4 +372,53 @@ private void checkNamespaceValidity(String namespace, String errorMsg) throws Se
...
     }
+
+    private boolean systemRamLimitExceeded() {
+        String ramUsage = null;
+        try {
+            ramUsage = dockerConnector.getSystemInfo().ramUsage();
+        } catch (IOException e) {
+            LOG.error("A problem occurred while getting system information from docker", e);
+        }
+        if (ramUsage == null) {
```
evoevodin: Can you please change the code to return `false` in the catch block instead.

wsmaster/codenvy-hosted-platform-api-impl/src/main/java/com/codenvy/api/workspace/LimitsCheckingWorkspaceManager.java
```diff
@@ -161,6 +186,35 @@ public WorkspaceImpl updateWorkspace(String id, Workspace update) throws Conflic
...
+                                                                                                               NotFoundException,
+                                                                                                               ConflictException {
+        if (startSemaphore == null) {
+            return checkRamAndPropagateStart(config, envName, namespace, workspaceId, callback);
+        } else {
+            try {
+                startSemaphore.acquire();
+                return checkRamAndPropagateStart(config, envName, namespace, workspaceId, callback);
+            } catch (InterruptedException e) {
+                throw new ServerException("Starting workspace was interrupted", e);
```
evoevodin: it shouldn't happen normally, can you please just rethrow `InterruptedException` as `ServerException` with the same message like `throw new ServerException(e.getMessage(), e)`

wsmaster/codenvy-hosted-platform-api-impl/src/main/java/com/codenvy/api/workspace/LimitsCheckingWorkspaceManager.java
```diff
@@ -161,6 +186,35 @@ public WorkspaceImpl updateWorkspace(String id, Workspace update) throws Conflic
     }
 
     /**
+     * One of the checks in {@link #checkRamAndPropagateStart(WorkspaceConfig, String, String, String, WorkspaceCallback)}
+     * requests to docker for system ram amount. This request can take much time, so we can not mark the method as synchronized
+     * because this will reduce performance. The solution is to make the system ram check optimistic by limiting
+     * same time start workspace requests. All requests to start workspace performed at the same time can receive not actual
+     * system ram limit status, so limits.workspace.start.throughput property configures how many requests to handle at the same time.
+     * If limits.workspace.start.threads.limit property is set to 0 or less then 0 all requests will be handled at the same time.
```
evoevodin: This is a typo actually, as the property name should be `limits.workspace.start_throughput`.
Also i'd operate term s as _parallel method calls_ instead of _requests_

wsmaster/codenvy-hosted-platform-api-impl/src/main/java/com/codenvy/api/workspace/LimitsCheckingWorkspaceManager.java
```diff
@@ -161,6 +186,35 @@ public WorkspaceImpl updateWorkspace(String id, Workspace update) throws Conflic
     }
 
     /**
+     * One of the checks in {@link #checkRamAndPropagateStart(WorkspaceConfig, String, String, String, WorkspaceCallback)}
+     * requests to docker for system ram amount. This request can take much time, so we can not mark the method as synchronized
+     * because this will reduce performance. The solution is to make the system ram check optimistic by limiting
+     * same time start workspace requests. All requests to start workspace performed at the same time can receive not actual
```
evoevodin: > The solution is to make the system ram check optimistic by limitingsame time start workspace requests

This is not clear for me, can you please rephrase it.

As far as i understand, this doc should be based on a few important statements : 
- this method can't be performed more than N times and the semaphore is used to control that
- system ram check may be really slow as it is based on request to the swarm, so it may took some time for swarm to figure out all the memory allocated on nodes. 
-  the system ram check must be performed on each start method call as it mustn't be possible to start a new workspace if system ram limit exceeded
- the semaphore is a trade off between speed and risk to exceed system ram limit. In the worst case which is almost impossible you may get N(permits) workspaces started after the actually system limit was exceeded, while by acquiring an exclusive lock the system may stack in the state when swarm doesn't response and the lock is held by the thread that requests that information. On the other hand the start itself is asynchronous and pretty fast, so the throughput limit event a small one won't affect the performance.

wsmaster/codenvy-hosted-system-ram/pom.xml
```diff
@@ -0,0 +1,203 @@
...
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <artifactId>codenvy-master-parent</artifactId>
+        <groupId>com.codenvy.onpremises.wsmaster</groupId>
+        <version>4.7.0-RC2-SNAPSHOT</version>
+    </parent>
+    <artifactId>codenvy-hosted-system-ram</artifactId>
+    <name>Codenvy :: Hosted :: System ram API</name>
```
evoevodin: weird, why `API` is in upper case while `ram` is in lower ?

wsmaster/codenvy-hosted-system-ram/src/main/java/com/codenvy/service/systemRam/SystemRamLimitChecker.java
```diff
@@ -0,0 +1,25 @@
...
+ * and its suppliers and may be covered by U.S. and Foreign Patents,
+ * patents in process, and are protected by trade secret or copyright law.
+ * Dissemination of this information or reproduction of this material
+ * is strictly forbidden unless prior written permission is obtained
+ * from Codenvy S.A..
+ */
+package com.codenvy.service.systemRam;
+
+/**
+ * Checks systems ram for 90% free ram existence.
```
evoevodin: I think this interface should not include such specific things as percentage.

wsmaster/codenvy-hosted-system-ram/src/main/java/com/codenvy/service/systemRam/SystemRamLimitChecker.java
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * @author Igor Vinokur
+ */
+public interface SystemRamLimitChecker {
+
+    String getSystemRamLimitStatus();
```
evoevodin: Why `String`, it looks like this interface should return an object describing system ram status, or at least an enum, 
i see the component "like this" pretty useful and i think it can be reused in other places as well, but the method declaration it is not clear _what should be returned_ ?

wsmaster/codenvy-hosted-system-ram/src/main/java/com/codenvy/service/systemRam/SystemRamLimitCheckerImpl.java
```diff
@@ -0,0 +1,134 @@
...
+    private String limitStatus = SYSTEM_RAM_LIMIT_NOT_EXCEEDED_STATUS;
+    private boolean systemRamLimitExceedMessageSent;
+
+    @Inject
+    public SystemRamLimitCheckerImpl(DockerConnector dockerConnector) {
+        this.dockerConnector = dockerConnector;
+    }
+
+    @Override
+    public final synchronized String getSystemRamLimitStatus() {
```
evoevodin: As far as i understand the whole method doesn't have to be synchronized

wsmaster/codenvy-hosted-system-ram/src/main/java/com/codenvy/service/systemRam/SystemRamLimitCheckerImpl.java
```diff
@@ -0,0 +1,134 @@
...
+import static java.util.concurrent.TimeUnit.MINUTES;
+import static java.util.concurrent.TimeUnit.SECONDS;
+
+/**
+ * Implementation of {@link SystemRamLimitChecker}
+ *
+ * @author Igor Vinokur
+ */
+@Singleton
+public class SystemRamLimitCheckerImpl implements SystemRamLimitChecker{
```
evoevodin: If it is docker specific it should be named in appropriate way.

wsmaster/codenvy-hosted-system-ram/src/main/java/com/codenvy/service/systemRam/SystemRamLimitCheckerImpl.java
```diff
@@ -0,0 +1,134 @@
...
+    @Override
+    public final synchronized String getSystemRamLimitStatus() {
+        long systemRamUsed = 0;
+        long systemRamTotal = 0;
+        List<String> allNodesRamUsage = new ArrayList<>();
+        Callable<SystemInfo> task = dockerConnector::getSystemInfo;
+        Future<SystemInfo> future = executor.submit(task);
+        SystemInfo systemInfo;
+        try {
+            systemInfo = future.get(5, SECONDS);
```
evoevodin: What is the default timeout of request to docker api? Why doesn't it fit your needs?

wsmaster/codenvy-hosted-system-ram/src/main/java/com/codenvy/service/systemRam/SystemRamLimitCheckerImpl.java
```diff
@@ -0,0 +1,134 @@
...
+                return limitStatus;
+            }
+            int i = 0;
+            for (int l = systemValues.length; i < l; i++) {
+                String[] systemInfoEntry = systemValues[i];
+                if (systemInfoEntry.length == 2 && " └ Reserved Memory".equals(systemInfoEntry[0])) {
+                    allNodesRamUsage.add(systemInfoEntry[1]);
+                }
+            }
+        } catch (TimeoutException | InterruptedException | ExecutionException e) {
```
evoevodin: If think there is should be a different catch for `TimeotException` with a different log message.

wsmaster/codenvy-hosted-system-ram/src/main/java/com/codenvy/service/systemRam/SystemRamLimitCheckerImpl.java
evoevodin: Are you sure that you need to synchronize the whole method ?

wsmaster/codenvy-hosted-system-ram/src/main/java/com/codenvy/service/systemRam/SystemRamLimitCheckerImpl.java
```diff
@@ -0,0 +1,134 @@
...
+        try {
+            WSConnectionContext.sendMessage(bm);
+        } catch (EncodeException | IOException e) {
+            LOG.error("A problem occurred while sending websocket message", e);
+        }
+    }
+
+    @ScheduleDelay(initialDelay = 1, delay = 1, unit = MINUTES)
+    private void freeRamCheckScheduler() {
+        if (systemRamLimitExceedMessageSent && SYSTEM_RAM_LIMIT_NOT_EXCEEDED_STATUS.equals(getSystemRamLimitStatus())) {
```
evoevodin: Why did you need to send a message if the message is already sent ? Looks like this method behaves wrongly. Also this method is not synchronized and uses non-volatile variables which means that you can see stale values, please rework it.

evoevodin:
-1 for now.
The design and implementation of the component for ram checking must be reworked.
I didn't check all the code, but i think that i left more than enough comment for now, please check&answer where necessary.

wsmaster/codenvy-hosted-platform-api-impl/src/main/java/com/codenvy/api/workspace/LimitsCheckingWorkspaceManager.java
vinokurig: done

wsmaster/codenvy-hosted-platform-api-impl/src/main/java/com/codenvy/api/workspace/LimitsCheckingWorkspaceManager.java
vinokurig: reworked the doc

wsmaster/codenvy-hosted-system-ram/pom.xml
vinokurig: renamed to `System RAM API`

wsmaster/codenvy-hosted-system-ram/src/main/java/com/codenvy/service/systemRam/SystemRamLimitChecker.java
vinokurig: done

wsmaster/codenvy-hosted-system-ram/src/main/java/com/codenvy/service/systemRam/SystemRamLimitChecker.java
vinokurig: done

wsmaster/codenvy-hosted-system-ram/src/main/java/com/codenvy/service/systemRam/SystemRamLimitCheckerImpl.java
vinokurig: done

wsmaster/codenvy-hosted-system-ram/src/main/java/com/codenvy/service/systemRam/SystemRamLimitCheckerImpl.java
vinokurig: removed the `synchronize`

wsmaster/codenvy-hosted-system-ram/src/main/java/com/codenvy/service/systemRam/SystemRamLimitCheckerImpl.java
vinokurig: the method is a scheduler and `systemRamLimitExceedMessageSent` is needed to enable or disable request inside the scheduler if message was sent or not.
Changed variables to volatile.

wsmaster/codenvy-hosted-system-ram/src/main/java/com/codenvy/service/systemRam/SystemRamLimitCheckerImpl.java
vinokurig: There is no default timeout

wsmaster/codenvy-hosted-system-ram/src/main/java/com/codenvy/service/systemRam/SystemRamLimitCheckerImpl.java
vinokurig: done

vinokurig:
@evoevodin fixed all notes, please take a look

wsmaster/codenvy-hosted-system-ram/src/main/java/com/codenvy/service/systemRam/SystemRamLimitChecker.java
```diff
@@ -0,0 +1,25 @@
...
+ * and its suppliers and may be covered by U.S. and Foreign Patents,
+ * patents in process, and are protected by trade secret or copyright law.
+ * Dissemination of this information or reproduction of this material
+ * is strictly forbidden unless prior written permission is obtained
+ * from Codenvy S.A..
+ */
+package com.codenvy.service.systemRam;
+
+/**
+ * Checks system RAM limit status.
```
evoevodin: This seems to be wrong, as the interface has the only one method which has nothing to do with the ram check at all.

wsmaster/codenvy-hosted-system-ram/src/main/java/com/codenvy/service/systemRam/SystemRamLimitChecker.java
```diff
@@ -0,0 +1,25 @@
...
+ * the property of Codenvy S.A. and its suppliers,
+ * if any.  The intellectual and technical concepts contained
+ * herein are proprietary to Codenvy S.A.
+ * and its suppliers and may be covered by U.S. and Foreign Patents,
+ * patents in process, and are protected by trade secret or copyright law.
+ * Dissemination of this information or reproduction of this material
+ * is strictly forbidden unless prior written permission is obtained
+ * from Codenvy S.A..
+ */
+package com.codenvy.service.systemRam;
```
evoevodin: The package name is bad and violates the [naming rules](https://github.com/codenvy/codenvy-dev-resources/wiki/Naming#package-names)

wsmaster/codenvy-hosted-system-ram/src/main/java/com/codenvy/service/systemRam/SystemRamLimitChecker.java
```diff
@@ -0,0 +1,25 @@
...
+
+/**
+ * Checks system RAM limit status.
+ *
+ * @author Igor Vinokur
+ */
+public interface SystemRamLimitChecker {
+
+    SystemRamLimitStatus getSystemRamLimitStatus();
```
evoevodin: Neither this interface nor `SystemRamLimitStatus` provides a definition of statuses and possible situations, exceptions etc. What should i get if this method can't retrieve system information due to occurred error, it's not clear.

Again, can you please use an object which describes not only the status of the system but the memory usage information as well, then the component becomes more informative and usable + rest service can expose more information to the client which may be displayed to admins and so on.

wsmaster/codenvy-hosted-system-ram/src/main/java/com/codenvy/service/systemRam/SystemRamLimitStatus.java
```diff
@@ -0,0 +1,35 @@
...
+ * from Codenvy S.A..
+ */
+package com.codenvy.service.systemRam;
+
+/**
+ * Defines system RAM limit status.
+ *
+ * @author Igor Vinokur
+ */
+public enum SystemRamLimitStatus {
```
evoevodin: Enum is possible but pure solution here, i suggest you to use separate object which provides dynamic information for that purpose, check the [comment](https://github.com/codenvy/codenvy/pull/501/files#r76417936) above

wsmaster/codenvy-hosted-system-ram/src/main/java/com/codenvy/service/systemRam/SystemRamService.java
```diff
@@ -0,0 +1,51 @@
...
+    private final SystemRamLimitChecker systemRamLimitChecker;
+
+    @Inject
+    public SystemRamService(SystemRamLimitChecker systemRamLimitChecker) {
+        this.systemRamLimitChecker = systemRamLimitChecker;
+    }
+
+    @GET
+    @Path("/info")
+    @Consumes(MediaType.APPLICATION_JSON)
```
evoevodin: It doesn't consume anything, but produces _application/json_ it seems that this method is not appropriately tested

wsmaster/codenvy-hosted-system-ram/src/main/java/com/codenvy/service/systemRam/SystemRamService.java
```diff
@@ -0,0 +1,51 @@
...
+import javax.ws.rs.core.MediaType;
+
+import static org.eclipse.che.dto.server.DtoFactory.newDto;
+
+/**
+ * Defines system ram REST API.
+ *
+ * @author Igor Vinokur
+ */
+@Path("/ram")
```
evoevodin: i don't think __/api/ram/info__ is really good path, 
i would probably change it to something like __/api/system/ram__ or maybe __/api/system/raminfo__
@skabashnyuk what do you think about it?

wsmaster/codenvy-hosted-system-ram/src/main/java/com/codenvy/service/systemRam/DockerBasedSystemRamLimitChecker.java
```diff
@@ -0,0 +1,141 @@
...
+    @Override
+    public final SystemRamLimitStatus getSystemRamLimitStatus() {
+        long systemRamUsed = 0;
+        long systemRamTotal = 0;
+        List<String> allNodesRamUsage = new ArrayList<>();
+        Callable<SystemInfo> task = dockerConnector::getSystemInfo;
+        Future<SystemInfo> future = executor.submit(task);
+        SystemInfo systemInfo;
+        try {
+            systemInfo = future.get(5, SECONDS);
```
evoevodin: I didn't get an answer on my previous question on that.
Why 5 seconds? Why doesn't the default request timeout fits you needs?

wsmaster/codenvy-hosted-system-ram/src/main/java/com/codenvy/service/systemRam/DockerBasedSystemRamLimitChecker.java
```diff
@@ -0,0 +1,141 @@
...
+            WSConnectionContext.sendMessage(bm);
+        } catch (EncodeException | IOException e) {
+            LOG.error("A problem occurred while sending websocket message", e);
+        }
+    }
+
+    @ScheduleDelay(initialDelay = 1, delay = 1, unit = MINUTES)
+    private void freeRamCheckScheduler() {
+        if (systemRamLimitExceedMessageSent && SYSTEM_RAM_LIMIT_NOT_EXCEEDED != getSystemRamLimitStatus()) {
+            sendRamLimitStatusMessage(SYSTEM_RAM_LIMIT_NOT_EXCEEDED.toString());
```
akorneta: for me it is not obvious why we should send message about `SYSTEM_RAM_LIMIT_NOT_EXCEEDED` when `SYSTEM_RAM_LIMIT_NOT_EXCEEDED != getSystemRamLimitStatus()` says otherwise

wsmaster/codenvy-hosted-system-ram/src/main/java/com/codenvy/service/systemRam/DockerBasedSystemRamLimitChecker.java
```diff
@@ -0,0 +1,141 @@
...
+import static java.util.concurrent.TimeUnit.MINUTES;
+import static java.util.concurrent.TimeUnit.SECONDS;
+
+/**
+ * Implementation of {@link SystemRamLimitChecker}
+ *
+ * @author Igor Vinokur
+ */
+@Singleton
+public class DockerBasedSystemRamLimitChecker implements SystemRamLimitChecker{
```
evoevodin: @vinokurig consider the following:

Separate this class on 2 different components:
- first(inteface based) for retrieving information about system ram, which can be implemented for different systems
- second(certain without an interface) for scheduling and notifying the clients about system status when necessary, it will inject and reuse the first component to perform checks

evoevodin: -1 again. 
I didn't get an answers on some of my previous comments + the implementation is still not good enough. Can you please answer on the questions before you start fixing the code?

evoevodin:
-1 again. 
I didn't get an answers on some of my previous comments + the implementation is still not good enough. Can you please answer on the questions before you start fixing the code?

wsmaster/codenvy-hosted-system-ram/src/main/java/com/codenvy/service/systemRam/DockerBasedSystemRamLimitChecker.java
vinokurig: There is no default timeout in this request. Theoretically `dockerConnector.getSystemInfo()` can freeze without timeout

wsmaster/codenvy-hosted-system-ram/src/main/java/com/codenvy/service/systemRam/DockerBasedSystemRamLimitChecker.java
vinokurig: thanks for note, it was a mistake

vinokurig:
@evoevodin fixed all your notes, please take a look

wsmaster/codenvy-hosted-system-ram/src/main/java/com/codenvy/service/systemram/SystemRamInfo.java
```diff
@@ -0,0 +1,54 @@
...
+package com.codenvy.service.systemram;
+
+/**
+ * Describes system RAM values and properties.
+ *
+ * @author Igor Vinokur
+ */
+public class SystemRamInfo {
+
+    private long systemRamUsed;
```
evoevodin: `final` + hashcode, equals, toString

wsmaster/codenvy-hosted-system-ram/src/main/java/com/codenvy/service/systemram/SystemRamInfoProvider.java
```diff
@@ -0,0 +1,30 @@
...
+ * Provides system RAM information.
+ *
+ * @author Igor Vinokur
+ */
+public interface SystemRamInfoProvider {
+
+    /**
+     * Returns {@link SystemRamInfo} object, that describes system RAM values and properties.
+     */
+    SystemRamInfo getSystemRamInfo() throws ServerException;
```
evoevodin: document exception

wsmaster/codenvy-hosted-system-ram/src/main/java/com/codenvy/service/systemram/SystemRamLimitMessageSender.java
```diff
@@ -0,0 +1,84 @@
...
+        bm.setBody(DtoFactory.getInstance().toJson(systemRamLimit));
+        try {
+            WSConnectionContext.sendMessage(bm);
+        } catch (EncodeException | IOException e) {
+            LOG.error("A problem occurred while sending websocket message", e);
+        }
+    }
+
+    @VisibleForTesting
+    @ScheduleDelay(initialDelay = 1, delay = 1, unit = MINUTES)
```
evoevodin: extract to properties

wsmaster/codenvy-hosted-system-ram/src/main/java/com/codenvy/service/systemram/SystemRamInfo.java
vinokurig: done

wsmaster/codenvy-hosted-system-ram/src/main/java/com/codenvy/service/systemram/SystemRamInfoProvider.java
vinokurig: done

wsmaster/codenvy-hosted-system-ram/src/main/java/com/codenvy/service/systemram/SystemRamLimitMessageSender.java
vinokurig: done

assembly/onpremises-ide-packaging-war-platform-api/src/main/webapp/WEB-INF/classes/codenvy/codenvy-api-configuration.properties
```diff
@@ -56,3 +56,5 @@ che.stacks.images.storage=${catalina.base}/stacks/images
 stop.workspace.scheduler.period=60
 
 machine.ssh.connection_timeout_ms=3000
+
+system_ram.limit.check.scheduler.period=60
```
evoevodin: i don't think the property name is good, ram looks like a system resource for me @garagatyi wdyt /

wsmaster/codenvy-hosted-system-ram/src/main/java/com/codenvy/service/systemram/DockerBasedSystemRamInfoProvider.java
```diff
@@ -52,37 +50,71 @@ public DockerBasedSystemRamInfoProvider(DockerConnector dockerConnector) {
...
         }
         String[][] systemValues = systemInfo.getDriverStatus() == null ? systemInfo.getSystemStatus() : systemInfo.getDriverStatus();
         if (systemValues == null) {
-            LOG.error(EMPTY_DOCKER_SYSTEM_INFO_ERROR + "All system info from docker {}", systemInfo);
-            throw new ServerException(EMPTY_DOCKER_SYSTEM_INFO_ERROR);
+            LOG.error("Empty system information was received from docker. All system info from docker {}", systemInfo);
+            throw new ServerException(SYSTEM_RAM_INFO_ERROR);
         }
+
+        /**
```
evoevodin: Why _javadoc_ comment type ?

wsmaster/codenvy-hosted-system-ram/src/main/java/com/codenvy/service/systemram/SystemRamInfo.java
```diff
@@ -51,4 +51,35 @@ public long getSystemRamUsed() {
...
+
+        if (systemRamUsed != that.systemRamUsed) return false;
+        if (systemRamTotal != that.systemRamTotal) return false;
+        return isSystemRamLimitExceeded == that.isSystemRamLimitExceeded;
+
+    }
+
+    @Override
+    public int hashCode() {
+        int result = (int)(systemRamUsed ^ (systemRamUsed >>> 32));
```
evoevodin: please, use `Long.hashCode` or `Objects.hash()` or [check this out](https://github.com/evoevodin/intellij-idea-tools/tree/master/templates/equals-hashcode/collections-never-null)

wsmaster/codenvy-hosted-system-ram/src/main/java/com/codenvy/service/systemram/SystemRamInfo.java
```diff
@@ -51,4 +51,35 @@ public long getSystemRamUsed() {
...
     }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+
+        SystemRamInfo that = (SystemRamInfo)o;
+
+        if (systemRamUsed != that.systemRamUsed) return false;
```
evoevodin: Convert it to use single return

wsmaster/codenvy-hosted-system-ram/src/main/java/com/codenvy/service/systemram/SystemRamInfo.java
```diff
@@ -51,4 +51,35 @@ public long getSystemRamUsed() {
...
+    public int hashCode() {
+        int result = (int)(systemRamUsed ^ (systemRamUsed >>> 32));
+        result = 31 * result + (int)(systemRamTotal ^ (systemRamTotal >>> 32));
+        result = 31 * result + (isSystemRamLimitExceeded ? 1 : 0);
+        return result;
+    }
+
+    @Override
+    public String toString() {
+        final StringBuilder sb = new StringBuilder("SystemRamInfo{");
```
evoevodin: Why don't you just use strings concatenation here?

evoevodin:
Finally the general approach is okay for me, @sleshchenko @garagatyi @skabashnyuk can you please review the pull request.

wsmaster/codenvy-hosted-platform-api-impl/src/main/java/com/codenvy/api/workspace/LimitsCheckingWorkspaceManager.java
```diff
@@ -167,6 +180,37 @@ public WorkspaceImpl updateWorkspace(String id, Workspace update) throws Conflic
...
+                                                                                                               NotFoundException,
+                                                                                                               ConflictException {
+        if (startSemaphore == null) {
+            return checkRamAndPropagateStart(config, envName, namespace, callback);
+        } else {
+            try {
+                startSemaphore.acquire();
+                return checkRamAndPropagateStart(config, envName, namespace, callback);
+            } catch (InterruptedException e) {
+                throw new ServerException(e.getMessage(), e);
```
sleshchenko: I think you should set interrupt flag here

evoevodin: +1

wsmaster/codenvy-hosted-system-ram/src/main/java/com/codenvy/service/systemram/DockerBasedSystemRamInfoProvider.java
vinokurig: replaced with comment block

wsmaster/codenvy-hosted-system-ram/src/main/java/com/codenvy/service/systemram/SystemRamInfo.java
vinokurig: Used `Objects.hash()`

wsmaster/codenvy-hosted-system-ram/src/main/java/com/codenvy/service/systemram/DockerBasedSystemRamInfoProvider.java
```diff
@@ -0,0 +1,120 @@
...
+    public DockerBasedSystemRamInfoProvider(DockerConnector dockerConnector) {
+        this.dockerConnector = dockerConnector;
+    }
+
+    @Override
+    public SystemRamInfo getSystemRamInfo() throws ServerException {
+        long systemRamUsed = 0;
+        long systemRamTotal = 0;
+        SystemInfo systemInfo;
+        List<String> allNodesRamUsage = new ArrayList<>();
```
sleshchenko: I prefer to declare variable near usage

wsmaster/codenvy-hosted-system-ram/src/main/java/com/codenvy/service/systemram/SystemRamInfo.java
vinokurig: Used strings concatenation

wsmaster/codenvy-hosted-system-ram/src/main/java/com/codenvy/service/systemram/SystemRamInfo.java
vinokurig: done

wsmaster/codenvy-hosted-system-ram/src/main/java/com/codenvy/service/systemram/SystemRamInfo.java
```diff
@@ -0,0 +1,82 @@
...
+     * Returns {@code true} if there is less then 10% of free RAM is present in the system, otherwise returns {@code false}.
+     */
+    public boolean isSystemRamLimitExceeded() {
+        return isSystemRamLimitExceeded;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
```
sleshchenko: Why do you compare classes instead of usage `instanceOf`?

wsmaster/codenvy-hosted-system-ram/src/main/java/com/codenvy/service/systemram/SystemRamLimitMessageSender.java
```diff
@@ -0,0 +1,80 @@
...
+            WSConnectionContext.sendMessage(bm);
+        } catch (EncodeException | IOException e) {
+            LOG.error("An error occurred while sending web-socket message", e);
+        }
+    }
+
+    @VisibleForTesting
+    @ScheduleDelay(initialDelay = 60, delayParameterName = "system_ram.limit.check.scheduler.period")
+    synchronized void checkRamLimitAndSendMessageIfNeeded() {
+        boolean isSystemRamLimitExceeded;
```
sleshchenko: I would wrap logic of method like that (with @ScheduleDelay or @ScheduleRate) to `try{...} catch (Exception ...) {...}` because if some runtime exception occurs then subsequent executions will be suppressed.

wsmaster/codenvy-hosted-system-ram/src/main/java/com/codenvy/service/systemram/SystemRamLimitMessageSender.java
```diff
@@ -0,0 +1,80 @@
...
+        try {
+            WSConnectionContext.sendMessage(bm);
+        } catch (EncodeException | IOException e) {
+            LOG.error("An error occurred while sending web-socket message", e);
+        }
+    }
+
+    @VisibleForTesting
+    @ScheduleDelay(initialDelay = 60, delayParameterName = "system_ram.limit.check.scheduler.period")
+    synchronized void checkRamLimitAndSendMessageIfNeeded() {
```
sleshchenko: Why do you use `synchronized` here?

wsmaster/codenvy-hosted-system-ram/src/test/java/com/codenvy/service/systemram/SystemRamLimitMessageSenderTest.java
```diff
@@ -0,0 +1,84 @@
...
+        when(systemRamInfo.isSystemRamLimitExceeded()).thenReturn(true);
+
+        //when
+        systemRamLimitMessageSender.checkRamLimitAndSendMessageIfNeeded();
+
+        //then
+        assertTrue((boolean)systemRamLimitExceedMessageSent.get(systemRamLimitMessageSender));
+    }
+}
+
```
sleshchenko: remove extra lines

wsmaster/codenvy-hosted-system-ram/src/main/java/com/codenvy/service/systemram/DockerBasedSystemRamInfoProvider.java
```diff
@@ -0,0 +1,120 @@
...
+             ],
+             .
+             .
+             .
+          ],
+          RAM vales are divided by nodes and introduced in a string array which has '└ Reserved Memory' in first element,
+          and RAM values of node in format <used RAM size> / <total RAM size> in second element.
+         */
+        for (String[] systemInfoEntry : systemValues) {
+            if (systemInfoEntry.length == 2 && " └ Reserved Memory".equals(systemInfoEntry[0])) {
```
mshaposhnik: It is possible to be more than one `└ Reserved Memory` entry? If no, break the cycle after you found it.

evoevodin: I suppose it has to be clear from the comment above, but it is not =)

assembly/onpremises-ide-packaging-war-platform-api/src/main/java/com/codenvy/api/deploy/OnPremisesIdeApiServletModule.java
```diff
@@ -93,7 +93,9 @@ protected void configureServlets() {
                "/permissions",
                "/permissions/*",
                "/preferences",
-               "/preferences/*")
+               "/preferences/*",
+               "/system/ram",
```
sleshchenko: Seems that you don't need it. Mb it would be better to add this mapping when there will be some method that require it?

wsmaster/codenvy-hosted-system-ram/src/main/java/com/codenvy/service/systemram/DockerBasedSystemRamInfoProvider.java
```diff
@@ -0,0 +1,120 @@
...
+        SystemInfo systemInfo;
+        List<String> allNodesRamUsage = new ArrayList<>();
+
+        try {
+            systemInfo = dockerConnector.getSystemInfo();
+        } catch (IOException e) {
+            LOG.error("Failed to retrieve system information from docker.", e);
+            throw new ServerException(SYSTEM_RAM_INFO_ERROR, e);
+        }
+        String[][] systemValues = systemInfo.getDriverStatus() == null ? systemInfo.getSystemStatus() : systemInfo.getDriverStatus();
```
sleshchenko: Does systemInfo.getDriverStatus() have the same format as systemInfo.getSystemStatus()?

sleshchenko: Mb it would be better to save systemInfo.getDriverStatus() to variable to aviod double invocation of it.

sleshchenko: As for me SystemValues says nothing about variable content. `statusOutput` or something like that?

wsmaster/codenvy-hosted-system-ram/src/main/java/com/codenvy/service/systemram/SystemRamInfo.java
```diff
@@ -0,0 +1,82 @@
...
+
+    /**
+     * Used system RAM amount in Bytes.
+     */
+    public long getSystemRamUsed() {
+        return systemRamUsed;
+    }
+
+    /**
+     * Returns {@code true} if there is less then 10% of free RAM is present in the system, otherwise returns {@code false}.
```
sleshchenko: You can't ensure that `it returns true if there is less then 10% of free RAM` because you get this value in constructor.
I'd move logic of init this value in constructor or rewrite docs without description about condition of limit exceeding

wsmaster/codenvy-hosted-platform-api-impl/src/main/java/com/codenvy/api/workspace/LimitsCheckingWorkspaceManager.java
vinokurig: done

wsmaster/codenvy-hosted-system-ram/src/main/java/com/codenvy/service/systemram/DockerBasedSystemRamInfoProvider.java
vinokurig: done

wsmaster/codenvy-hosted-system-ram/src/main/java/com/codenvy/service/systemram/SystemRamInfo.java
vinokurig: done

wsmaster/codenvy-hosted-system-ram/src/main/java/com/codenvy/service/systemram/SystemRamLimitMessageSender.java
vinokurig: done

wsmaster/codenvy-hosted-system-ram/src/main/java/com/codenvy/service/systemram/SystemRamLimitMessageSender.java
vinokurig: This method uses global variable and the method can be executed in different threads, so it's better to mark the method as synchronized

wsmaster/codenvy-hosted-system-ram/src/test/java/com/codenvy/service/systemram/SystemRamLimitMessageSenderTest.java
vinokurig: done

wsmaster/codenvy-hosted-system-ram/src/main/java/com/codenvy/service/systemram/DockerBasedSystemRamInfoProvider.java
vinokurig: It can be more than one time, if more then one node is present in the system. Added this to description.

assembly/onpremises-ide-packaging-war-platform-api/src/main/java/com/codenvy/api/deploy/OnPremisesIdeApiServletModule.java
vinokurig: removed

wsmaster/codenvy-hosted-system-ram/src/main/java/com/codenvy/service/systemram/DockerBasedSystemRamInfoProvider.java
vinokurig: done, renamed to `statusOutput`

wsmaster/codenvy-hosted-system-ram/src/main/java/com/codenvy/service/systemram/SystemRamInfo.java
vinokurig: moved the logic to constructor

assembly/onpremises-ide-packaging-war-platform-api/src/main/webapp/WEB-INF/classes/codenvy/codenvy-api-configuration.properties
garagatyi: Yes, the name can be improved. @vinokurig take into account that we use convention:
- dots separate subresources.
- underscores unites complex terms
- property with number value must include measurement unit in last term separated by uinderscore
 E.g.  machine.ssh.connection_timeout_ms=3000 - it is connected to machine, ssh implementation of machine, described connection timeout in milliseconds

wsmaster/codenvy-hosted-platform-api-impl/src/main/java/com/codenvy/api/workspace/LimitsCheckingWorkspaceManager.java
```diff
@@ -128,10 +142,10 @@ public WorkspaceImpl startWorkspace(String workspaceId,
                 "Unable to start workspace %s, because its namespace owner is " +
                 "unavailable and it is impossible to check resources consumption.",
                 workspaceId));
-        return checkRamAndPropagateStart(workspace.getConfig(),
-                                         envName,
-                                         workspace.getNamespace(),
-                                         () -> super.startWorkspace(workspaceId, envName, accountId, restore));
+        return checkRamAndPropagateLimitedStart(workspace.getConfig(),
```
garagatyi: Why this method is renamed, new name doesn't add anything meaningful

wsmaster/codenvy-hosted-platform-api-impl/src/main/java/com/codenvy/api/workspace/LimitsCheckingWorkspaceManager.java
```diff
@@ -142,10 +156,10 @@ public WorkspaceImpl startWorkspace(WorkspaceConfig config,
                                                                            NotFoundException,
                                                                            ConflictException {
         checkMaxEnvironmentRam(config);
-        return checkRamAndPropagateStart(config,
-                                         config.getDefaultEnv(),
-                                         namespace,
-                                         () -> super.startWorkspace(config, namespace, isTemporary, accountId));
+        return checkRamAndPropagateLimitedStart(config,
```
garagatyi: Why this method is renamed, new name doesn't add anything meaningful

wsmaster/codenvy-hosted-platform-api-impl/src/test/java/com/codenvy/api/workspace/LimitsCheckingWorkspaceManagerTest.java
```diff
@@ -365,9 +424,12 @@ public void shouldBeAbleToCreateWorkspaceWithMultipleMachinesIncludingMachineWit
 
     @Test
     public void shouldBeAbleToStartWorkspaceWithMultipleMachinesIncludingMachineWithoutLimitsWhichDoesNotExceedRamLimit() throws Exception {
+        when(systemRamInfoProvider.getSystemRamInfo()).thenReturn(new SystemRamInfo(0, parseSize("3 GiB")));
         LimitsCheckingWorkspaceManager manager = spy(new LimitsCheckingWorkspaceManager(2,
                                                                                         "3gb",
                                                                                         "1gb", // <- workspaces ram limit
+                                                                                        0,
+                                                                                        systemRamInfoProvider,
                                                                                         null,
```
garagatyi: I don't see any test that check logic around semaphore usage. @vinokurig  Can you point out where these tests are?

wsmaster/codenvy-hosted-system-ram/pom.xml
```diff
@@ -0,0 +1,203 @@
...
+    is strictly forbidden unless prior written permission is obtained
+    from Codenvy S.A..
+
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <artifactId>codenvy-master-parent</artifactId>
+        <groupId>com.codenvy.onpremises.wsmaster</groupId>
+        <version>4.8.0-SNAPSHOT</version>
```
garagatyi: Rebase with master, change version

wsmaster/codenvy-hosted-platform-api-impl/src/main/java/com/codenvy/api/workspace/LimitsCheckingWorkspaceManager.java
vinokurig: It is a new method that wraps `checkRamAndPropagateStart` method

wsmaster/codenvy-hosted-platform-api-impl/src/main/java/com/codenvy/api/workspace/LimitsCheckingWorkspaceManager.java
vinokurig: It is a new method that wraps `checkRamAndPropagateStart` method

garagatyi: But here limited is not to obvious. Maybe change to limited throughput ?

wsmaster/codenvy-hosted-platform-api-impl/src/test/java/com/codenvy/api/workspace/LimitsCheckingWorkspaceManagerTest.java
vinokurig: added tests https://github.com/codenvy/codenvy/blob/edefd63cd1962f445a71dd6564465a35dd35a054/wsmaster/codenvy-hosted-platform-api-impl/src/test/java/com/codenvy/api/workspace/LimitsCheckingWorkspaceManagerTest.java#L453-L559

wsmaster/codenvy-hosted-system-ram/pom.xml
vinokurig: done

wsmaster/codenvy-hosted-platform-api-impl/src/main/java/com/codenvy/api/workspace/LimitsCheckingWorkspaceManager.java
vinokurig: renamed

assembly/onpremises-ide-packaging-war-platform-api/src/main/webapp/WEB-INF/classes/codenvy/codenvy-api-configuration.properties
vinokurig: renamed to `system.ram.limit_check_period_sec`

wsmaster/codenvy-hosted-platform-api-impl/src/test/java/com/codenvy/api/workspace/LimitsCheckingWorkspaceManagerTest.java
```diff
@@ -385,4 +449,112 @@ public void shouldBeAbleToStartWorkspaceWithMultipleMachinesIncludingMachineWith
...
+        WorkspaceCallback callback = mock(WorkspaceCallback.class);
+        doReturn(singletonList(createRuntime("256mb", "256mb", null))).when(manager).getByNamespace(anyString());
+
+        manager.checkRamAndPropagateLimitedThroughputStart(createConfig("256mb", "256mb", null), null, "user123", callback);
+
+        assertNull(manager.startSemaphore);
+    }
+
+    @Test
+    public void shouldSetSemaphoreToNullIfThroughputPropertyIsLessThenZero() throws Exception {
```
garagatyi: Tests that check throughput limitation are also needed

wsmaster/codenvy-hosted-system-ram/src/main/java/com/codenvy/service/systemram/SystemRamInfo.java
garagatyi: You should not explain here clause of memory exceeding. It may change.

wsmaster/codenvy-hosted-system-ram/src/main/java/com/codenvy/service/systemram/SystemRamInfoProvider.java
```diff
@@ -0,0 +1,32 @@
...
+ * Dissemination of this information or reproduction of this material
+ * is strictly forbidden unless prior written permission is obtained
+ * from Codenvy S.A..
+ */
+package com.codenvy.service.systemram;
+
+import org.eclipse.che.api.core.ServerException;
+
+/**
+ * Provides system RAM information.
```
garagatyi: What means system RAM info for distributes service like Codenvy? Does it include machine nodes, api node, etc?

wsmaster/codenvy-hosted-system-ram/src/main/java/com/codenvy/service/systemram/DockerBasedSystemRamInfoProvider.java
```diff
@@ -0,0 +1,122 @@
...
+            systemInfo = dockerConnector.getSystemInfo();
+        } catch (IOException e) {
+            LOG.error("Failed to retrieve system information from docker.", e);
+            throw new ServerException(SYSTEM_RAM_INFO_ERROR, e);
+        }
+        String[][] driverStatus = systemInfo.getDriverStatus();
+        String[][] systemStatus = systemInfo.getSystemStatus();
+        String[][] statusOutput = driverStatus == null ? systemStatus : driverStatus;
+        if (statusOutput == null) {
+            LOG.error("Empty system information was received from docker. All system info from docker {}", systemInfo);
```
garagatyi: I think `whole` is better than `all` in that case

wsmaster/codenvy-hosted-system-ram/src/main/java/com/codenvy/service/systemram/SystemRamLimitMessageSender.java
```diff
@@ -0,0 +1,78 @@
...
+        bm.setBody(newDto(SystemRamLimitDto.class).withSystemRamLimitExceeded(isSystemRamLimitExceeded).toString());
+        try {
+            WSConnectionContext.sendMessage(bm);
+        } catch (EncodeException | IOException e) {
+            LOG.error("An error occurred while sending web-socket message", e);
+        }
+    }
+
+    @VisibleForTesting
+    @ScheduleDelay(initialDelay = 60, delayParameterName = "system_ram.limit.check.scheduler.period")
```
garagatyi: Please recheck name of property

garagatyi:
LGTM

wsmaster/codenvy-hosted-platform-api-impl/src/test/java/com/codenvy/api/workspace/LimitsCheckingWorkspaceManagerTest.java
vinokurig: added test: https://github.com/codenvy/codenvy/blob/2abe1b71f2240636ebd3b28c5e5bf404492ac7c0/wsmaster/codenvy-hosted-platform-api-impl/src/test/java/com/codenvy/api/workspace/LimitsCheckingWorkspaceManagerTest.java#L567-L621

wsmaster/codenvy-hosted-system-ram/src/main/java/com/codenvy/service/systemram/SystemRamInfoProvider.java
vinokurig: Updated the description

skabashnyuk:
ok for me. Final decision up to @evoevodin and @garagatyi

skabashnyuk:
@evoevodin  wdyt?

evoevodin:
The general approach is okay for me.
@vinokurig it was a long discussion, go ahead and finally merge it

PR #511, opened 2016-08-03
Add ability to store empty codenvy properties

dmytro-ndp:

### DIFF
--- a/cdec/installation-manager-cli-bundle/src/main/java/com/codenvy/im/cli/command/ConfigCommand.java
+++ b/cdec/installation-manager-cli-bundle/src/main/java/com/codenvy/im/cli/command/ConfigCommand.java
@@ -78,7 +78,7 @@ public class ConfigCommand extends AbstractIMCommand {
         }
 
         if (!StringUtils.isEmpty(property)) {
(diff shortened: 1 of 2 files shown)


cdec/installation-manager-cli-bundle/src/main/java/com/codenvy/im/cli/command/ConfigCommand.java
```diff
@@ -78,7 +78,7 @@ protected void doExecuteCommand() throws Exception {
         }
 
         if (!StringUtils.isEmpty(property)) {
-            if (!StringUtils.isEmpty(value)) {
+            if (value != null) {
```
tolusha: Why empty value is ok ?

dmytro-ndp: Because there are a lot of empty properties in config, and user should be able to restore they.

tolusha: ok

Issue #464, opened 2016-07-27
Add record '127.0.0.1 <new-hostname>' into /etc/hosts when changing Codenvy on-prem hostname


dmytro-ndp:
Adding record **'127.0.0.1 <new-hostname>'** into _/etc/hosts_ when changing Codenvy on-prem hostname by _Installation Manager_ will allow user:
1) to avoid blocking changing hostname when new hostname unavailable;
2) to avoid going to docker container from API server by using external IP which could lead to usage of closed outside ports.

Codenvy version: <=4.5.1

bmicklea:
This has affected quite a few customers - I'd like this to be considered for next sprint.

PR #517, opened 2016-08-04
(#464) update /etc/hosts when changing Codenvy hostname

dmytro-ndp:

### DIFF
--- a/cdec/installation-manager-core/src/main/java/com/codenvy/im/artifacts/helper/CDECSingleServerHelper.java
+++ b/cdec/installation-manager-core/src/main/java/com/codenvy/im/artifacts/helper/CDECSingleServerHelper.java
@@ -40,6 +40,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
 
+import static com.codenvy.im.commands.CommandLibrary.createAppendTextIfAbsentToFileCommand;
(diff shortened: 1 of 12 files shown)


cdec/installation-manager-core/src/main/java/com/codenvy/im/artifacts/helper/CDECSingleServerHelper.java
```diff
@@ -535,6 +539,12 @@ public Command getReinstallCommand(Config config, @Nullable Version installedVer
     public Command getUpdateHostnameCommand(Config config, String oldHostName, String newHostName) {
         List<Command> commands = new ArrayList<>();
 
+        // add <127.0.0.1 new-hostname> record into the /etc/hosts on master
+        commands.add(createAppendTextToFileCommand(Paths.get("/etc/hosts"),
```
tolusha: Should we check if needed host already there ?

dmytro-ndp: Yes, we should, and we check it.

dmytro-ndp: I will change name of method *'createAppendTextToFileCommand()'* on *'createAppendTextIfAbsentToFileCommand()'*.

Issue #509, opened 2016-08-03
Share a workspace: User not found message when sharing a workspace


eivantsov:
Right now when attempting to add a user to a workspace and this user isn't a Codenvy user yet, there's no clue in User Dashboard why a user cannot be added. That is why it is perceived as bug.

Reproduction Steps:

1. Workspaces > Share
2. Try adding a user noname@codenvy.com
3. Share button is inactive

Expected behavior:

A message show show up - User with email noname@codenvy.com not found. Workspaces can be shared with registered users only.

Observed behavior:

Share button is inactive.

ashumilova:
Resolved

PR #522, opened 2016-08-05
CODENVY-509: add user not found message in share workspace widget

ashumilova:
### What does this PR do?
Adds message "User with email [email] not found. Workspaces can be shared with registered users only." when unregistered email is passed.

Issue - https://github.com/codenvy/codenvy/issues/509

### Previous Behavior
No message was shown

### New Behavior
A message show show up - User with email noname@codenvy.com not found. Workspaces can be shared with registered users only.


ashumilova:
share_user_not_found

Issue #550, opened 2016-08-10
Wrong behavior while creating new account


AndreyChizhikov:
Reproduction Steps:
1. Go to  https://nightly4.codenvy-stg.com/site/login
2. Click on 'Create One'
3. Enter email and name
4. Click on  'Create Account'

Expected behavior:
Account has been created.

Observed behavior:
Status 500 Internal Server Error
![site](https://cloud.githubusercontent.com/assets/8051404/17548814/4485d85c-5ef6-11e6-8840-827e66729582.png)

Codenvy version: 4.7.0-SNAPSHOT
OS and version: Ubuntu 14.04 LTS

Additional information:

Problem only started happening recently, didn't happen in an older version of Codenvy: Yes 
Problem can be reliably reproduced: Yes

PR #558, opened 2016-08-10
CODENVY-550 Fix mail server groupId

mkuznyetsov:
### What does this PR do?
Set correct groupId for mail server so it can be picked up by assembly.

https://github.com/codenvy/codenvy/issues/550

### Tests written?
No


skabashnyuk:
ok

PR #561, opened 2016-08-11
add download mirror to speed up re-installation

t-oss:
### What does this PR do?

Update install-codenvy.sh, add download mirror to speed up re-installation. It's useful for someone has poor internet connection to codenvy.com @AmazonAWS or other dependency component site (such as JRE@Oracle).
Download files will be mirrored at the first installation, and then can be reused next time. 
### New Behavior

With this update, user can use --downloads-directory=<DOWNLOADS_DIR> to specified a single directory for mirror the downloads files during the installation.
Currently only mirror the following downloads:
1. JRE
2. Installation Manager
3. Codenvy

Better to mirror in future:
1. yum PKGs like puppet and so on.

What's more with this update, download files will not be deleted after installation automatically, to aim that these download files can be reused at next installation or fail-retry. 

To delete mirrored files, just remove the directory specified by "--downloads-directory" or "${--install-directory}/downloads" by default.
### Tests written?

Tested. Can speed up the installation duration from 300+ mins to about 40 mins on my seat (@China)
### Docs requirements?

1.  User docs changes

### DIFF
--- a/cdec/installation-manager-resources/src/main/resources/im-scripts/install-codenvy.sh
+++ b/cdec/installation-manager-resources/src/main/resources/im-scripts/install-codenvy.sh
@@ -12,6 +12,7 @@
 # --systemAdminPassword=<SYSTEM ADMIN PASSWORD>
 # --license=accept
 # --install-directory=<PATH TO SINGLE DIRECTORY FOR ALL RESOURCES>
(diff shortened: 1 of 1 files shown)


tolusha:
@dmytro-ndp pls review

skabashnyuk:
cdec project removed from master.

Issue #560, opened 2016-08-10
netstat deprecated and not available on some centos 7 distributions


TylerJewell:
When doing a default installation of Codenvy OnPrem for google cloud with their centos 7 node, we got a series of errors.

```
/dev/fd/63: line 902: netstat: command not found
```

Upon investigation, netstat is not installed by default.  Further investigation shows this command is the error.

```
echo $(netstat -ano | egrep LISTEN | egrep tcp | egrep ":80\s")
```

The solution is to replace `netstat -ano` with `ss -a`.  See more details here:
https://dougvitale.wordpress.com/2011/12/21/deprecated-linux-networking-commands-and-their-replacements/#netstat

dmytro-ndp:
@TylerJewell: take into account that requesting ss command throught ssh production error 'command not found':
```
$ ssh -o StrictHostKeyChecking=no -i /home/$USER/.vagrant.d/insecure_private_key vagrant@192.168.56.110 "ss"
bash: ss: command not found

```

It is needed to add *'. /etc/profile'* when calling ss:
```$ ssh -o StrictHostKeyChecking=no -i /home/$USER/.vagrant.d/insecure_private_key vagrant@192.168.56.110 ". /etc/profile && ss"```
> Netid  State      Recv-Q Send-Q Local Address:Port                 Peer Address:Port                
> u_str  ESTAB      0      0      /var/run/postgresql/.s.PGSQL.5432 2773621               * 2774316              
> u_str  ESTAB      0      0       * 2316468               * 2316469
> ...

PR #568, opened 2016-08-12
(#560) replace deprecated 'netstat' command by 'ss'

dmytro-ndp:
### What does this PR do?
It replaces 'netstat' command by 'ss' so as first one is deprecated and not available on some CentOS 7.


riuvshin:
ok

Issue #547, opened 2016-08-10
Default admin name of zabbix is "Admin", not "admin"


t-oss:
See from official document of ZABBIX, the default admin name is "Admin", not the "admin" mentioned in "codenvy.properties"

riuvshin:
@t-oss  good catch! we'll update it, thanks!

dmytro-ndp:
@riuvshin: which Codenvy default property should we use to define zabbix admin name?

PR #569, opened 2016-08-12
(#547) fix zabbix admin name in default Codenvy on-prem properties

dmytro-ndp:


riuvshin:
ok

Issue #552, opened 2016-08-10
Implement API for organizations


vkuznyetsov:
As result of this issue there should be implemented API that will allow users to:

- [x] manage organizations (create, delete, fetch)
- [x] create suborganizations (create, delete, fetch)
- [x] manage members of organizations (should be able to manage members by permissions service, like workers)

PR #572, opened 2016-08-12
Implement Organization API

sleshchenko:
### What does this PR do?
Implements Organization API 

https://github.com/codenvy/codenvy/issues/552

### New Behavior
Add ability to create organization to manage common workspaces

### Tests written?
Yes


sleshchenko:
@akorneta @skabashnyuk @evoevodin Please review my changes

wsmaster/codenvy-hosted-api-organization/pom.xml
```diff
@@ -0,0 +1,248 @@
...
+            <groupId>org.everrest</groupId>
+            <artifactId>everrest-core</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-api</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.testng</groupId>
+            <artifactId>testng</artifactId>
```
skabashnyuk: scope test?

akorneta:
ok

wsmaster/codenvy-hosted-api-organization/src/main/java/com/codenvy/organization/shared/Constants.java
```diff
@@ -0,0 +1,29 @@
...
+package com.codenvy.organization.shared;
+
+/**
+ * Constants for Organization API
+ *
+ * @author Sergii Leschenko
+ */
+public final class Constants {
+    public static final String LINK_REL_SELF                    = "self";
+    public static final String LINK_REL_GET_CHILD_ORGANIZATIONS = "get child organizations";
```
evoevodin: This is not actually the right way to define rels, [see](https://github.com/eclipse/che/blob/master/wsmaster/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/Constants.java#L22-L34)

wsmaster/codenvy-hosted-api-organization/src/main/java/com/codenvy/organization/shared/model/Organization.java
```diff
@@ -0,0 +1,42 @@
...
+ */
+public interface Organization {
+
+    /**
+     * Returns id of organization
+     */
+    String getId();
+
+    /**
+     * Returns name of organization
```
evoevodin: It says nothing about name uniqueness i think that the fact that user name and organization name are in the same namespace is important

wsmaster/codenvy-hosted-api-organization/src/main/java/com/codenvy/organization/shared/model/Organization.java
```diff
@@ -0,0 +1,42 @@
...
+     */
+    String getName();
+
+    /**
+     * Returns id of parent organization
```
evoevodin: If  parent organization is `null` then this organization is parent organization, am i right ?
If so i think it is necessary to add a documentation which specifies this fact.

wsmaster/codenvy-hosted-api-organization/src/main/java/com/codenvy/organization/spi/jpa/JpaMemberDao.java
```diff
@@ -0,0 +1,230 @@
...
+
+import static java.lang.String.format;
+import static java.util.Objects.requireNonNull;
+
+/**
+ * JPA based implementation of {@link MemberDao}.
+ *
+ * @author Sergii Leschenko
+ */
+public class JpaMemberDao implements MemberDao {
```
evoevodin: `@Singleton` ?

wsmaster/codenvy-hosted-api-organization/src/main/java/com/codenvy/organization/spi/jpa/JpaOrganizationDao.java
```diff
@@ -0,0 +1,151 @@
...
+
+import static java.lang.String.format;
+import static java.util.Objects.requireNonNull;
+
+/**
+ * JPA based implementation of {@link OrganizationDao}.
+ *
+ * @author Sergii Leschenko
+ */
+public class JpaOrganizationDao implements OrganizationDao {
```
evoevodin: `@Singleton` ?

wsmaster/codenvy-hosted-api-organization/src/main/java/com/codenvy/organization/shared/Constants.java
sleshchenko: Thanks. I'll take a look new approach and rework rels

wsmaster/codenvy-hosted-api-organization/src/main/java/com/codenvy/organization/spi/impl/MemberImpl.java
```diff
@@ -0,0 +1,107 @@
...
+                                    "FROM Member m " +
+                                    "WHERE m.userId = :userId"),
+                @NamedQuery(name = "Member.getOrganizations",
+                            query = "SELECT org " +
+                                    "FROM Member m, m.organization org " +
+                                    "WHERE m.userId = :userId")
+        }
+)
+@Table(indexes = @Index(columnList = "userId, organizationId", unique = true))
+public class MemberImpl extends AbstractPermissions implements Member {
```
skabashnyuk: Need to think about dedicated index only by ```organizationId``` field

wsmaster/codenvy-hosted-api-organization/src/main/java/com/codenvy/organization/spi/impl/OrganizationImpl.java
```diff
@@ -0,0 +1,144 @@
...
+
+    @Id
+    private String id;
+
+    @OneToOne(cascade = CascadeType.ALL)
+    @JoinColumn(nullable = false)
+    private AccountImpl account;
+
+    @Column
+    private String parent;
```
skabashnyuk: Index?

wsmaster/codenvy-hosted-api-organization/src/main/java/com/codenvy/organization/spi/impl/OrganizationImpl.java
```diff
@@ -0,0 +1,144 @@
...
+ * @author Sergii Leschenko
+ */
+@Entity(name = "Organization")
+@EntityListeners(OrganizationEntityListener.class)
+@NamedQueries(
+        {
+                @NamedQuery(name = "Organization.getByName",
+                            query = "SELECT o " +
+                                    "FROM Organization o " +
+                                    "WHERE o.account.name = :name"),
```
skabashnyuk: Make sure we have index on ```name```

skabashnyuk:
What about pagination on methods that return list? Have you thought about it?

wsmaster/codenvy-hosted-api-organization/src/main/java/com/codenvy/organization/api/OrganizationManager.java
```diff
@@ -0,0 +1,200 @@
...
+    public OrganizationImpl create(Organization newOrganization) throws ConflictException, ServerException {
+        requireNonNull(newOrganization, "Required non-null organization");
+        if (reservedNames.contains(newOrganization.getName().toLowerCase())) {
+            throw new ConflictException(String.format("Organization name '%s' is reserved", newOrganization.getName()));
+        }
+        final OrganizationImpl organization = new OrganizationImpl(NameGenerator.generate("organization", 16),
+                                                                   newOrganization.getName(),
+                                                                   newOrganization.getParent());
+        organizationDao.create(organization);
+        eventService.publish(new OrganizationCreatedEvent(organization));
```
mshaposhnik: Just for my information, why it can't be done using `OrgaizationEntityListener` and `@PostPersist`?

wsmaster/codenvy-hosted-api-organization/src/main/java/com/codenvy/organization/spi/jpa/JpaMemberDao.java
```diff
@@ -0,0 +1,219 @@
...
+        @PreDestroy
+        public void unsubscribe() {
+            eventService.unsubscribe(this);
+        }
+
+        @Override
+        public void onEvent(BeforeUserRemovedEvent event) {
+            try {
+                for (MemberImpl member : memberDao.getMemberships(event.getUser().getId())) {
+                    memberDao.remove(member.getOrganizationId(), member.getUserId());
```
mshaposhnik: Do we need to remove organization itself in some cases (when no more members with setPermissions/manage* roles left or etc)?

wsmaster/codenvy-hosted-api-organization/src/main/java/com/codenvy/organization/spi/impl/OrganizationImpl.java
sleshchenko: We have unique index there https://github.com/eclipse/che/blob/jpa-integration/wsmaster/che-core-api-account/src/main/java/org/eclipse/che/account/spi/AccountImpl.java#L41-41

wsmaster/codenvy-hosted-api-organization/src/main/java/com/codenvy/organization/spi/jpa/JpaMemberDao.java
sleshchenko: yes. thx

wsmaster/codenvy-hosted-api-organization/src/main/java/com/codenvy/organization/api/OrganizationManager.java
sleshchenko: I think that publishing this kind of events is logic of manager level but not data. Because it is needed for business logic - not for resolving data integrity issues

wsmaster/codenvy-hosted-api-organization/src/main/java/com/codenvy/organization/spi/impl/OrganizationImpl.java
sleshchenko: Added

wsmaster/codenvy-hosted-api-organization/src/main/java/com/codenvy/organization/spi/jpa/JpaMemberDao.java
sleshchenko: Added https://github.com/sleshchenko/codenvy/blob/organizations/wsmaster/codenvy-hosted-api-organization/src/main/java/com/codenvy/organization/api/permissions/RemoveOrganizationOnLastUserRemovedEventSubscriber.java#L34-34

sleshchenko:
@skabashnyuk I've added paging

wsmaster/codenvy-hosted-api-organization/src/main/java/com/codenvy/organization/api/permissions/OrganizationCreatorPermissionsProvider.java
```diff
@@ -0,0 +1,73 @@
...
+
+    @PreDestroy
+    void unsubscribe() {
+        eventService.unsubscribe(this);
+    }
+
+    @Override
+    public void onEvent(OrganizationCreatedEvent event) {
+        try {
+            memberDao.store(new MemberImpl(EnvironmentContext.getCurrent().getSubject().getUserId(),
```
evoevodin: `OrganizationCreatedEvent` is published in `OrganizationManager` are you sure that subject always exists here?

sleshchenko: Yes. Because for now only OrganizationService invokes OrganizationManager. When there will be some logic outside of service then we'll need to rework this logic or adapt this and write new that will be provided permissions for creator.

wsmaster/codenvy-hosted-api-organization/src/main/java/com/codenvy/organization/spi/jpa/JpaMemberDao.java
```diff
@@ -0,0 +1,197 @@
...
+    @Transactional
+    protected void doRemove(String organizationId, String userId) {
+        final EntityManager manager = managerProvider.get();
+        try {
+            MemberImpl member = manager.createNamedQuery("Member.getMember", MemberImpl.class)
+                                       .setParameter("userId", userId)
+                                       .setParameter("organizationId", organizationId)
+                                       .getSingleResult();
+            manager.remove(member);
+        } catch (NoResultException ignored) {
```
evoevodin: you can avoid `ignored` exception by requesting result list, i like this approach more
```
List<MemberImpl> members = (...).getResultList();
if (!members.isEmpty()) {
    manager.remove(members.get(0));
}
````

sleshchenko: Looks better but in my way it ensures that there is not more than one member else will be rethrown ServerException. In your way will be removed only first member.
But :) I think I should not ensure it here because it is responsibility of unique index but not `remove` method.
I'll rework it, thanks

Issue #537, opened 2016-08-09
Rework MachineBackupManager multithreaded tests


mmorhun:
We should rewrite `MachineBackupManager` multithreaded tests without `sleep`s and make them hardware independent.

PR #592, opened 2016-08-17
CODENVY-537: Rework MachineBackupManager multithreaded tests

mmorhun:
### What does this PR do?
Removes all `sleep` methods in `MachineBackupManagerTest` 

#537 

### Previous Behavior
Multithreaded tests use `sleep` for synchronization, which may cause tests fail in specific environment.
Long tests duration (2-3 seconds per test case).

### New Behavior
No `sleep` methods in the tests.

### Docs requirements?
No

### DIFF
--- a/plugins/plugin-hosted/codenvy-machine-hosted/src/test/java/com/codenvy/machine/backup/MachineBackupManagerTest.java
+++ b/plugins/plugin-hosted/codenvy-machine-hosted/src/test/java/com/codenvy/machine/backup/MachineBackupManagerTest.java
@@ -15,11 +15,9 @@
 package com.codenvy.machine.backup;
 
 import org.eclipse.che.api.core.ServerException;
-import org.eclipse.che.commons.test.mockito.answer.WaitingAnswer;
 import org.mockito.ArgumentCaptor;
+import org.mockito.Captor;
 import org.mockito.Mock;
-import org.mockito.invocation.InvocationOnMock;
-import org.mockito.stubbing.Answer;
 import org.mockito.testng.MockitoTestNGListener;
 import org.slf4j.Logger;
 import org.testng.annotations.AfterMethod;
@@ -31,6 +29,7 @@ import java.io.File;
 import java.io.IOException;
 import java.lang.reflect.Field;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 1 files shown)


mmorhun:
@garagatyi take a look, please

mmorhun:
@sleshchenko ?

plugins/plugin-hosted/codenvy-machine-hosted/src/test/java/com/codenvy/machine/backup/MachineBackupManagerTest.java
```diff
@@ -416,99 +410,122 @@ public void shouldBackupWithCleanupAfterFinishOfCurrentRestore() throws Exceptio
...
-                backupManager.backupWorkspace(WORKSPACE_ID, SRC_PATH, SRC_ADDRESS);
-            } catch (ServerException e) {
-                LOG.error(e.getLocalizedMessage(), e);
-            }
-        });
+    private void runBackup() {
+        try {
+            backupManager.backupWorkspace(WORKSPACE_ID, SRC_PATH, SRC_ADDRESS);
+        } catch (ServerException e) {
+            LOG.error(e.getMessage());
```
sleshchenko: Are you sure that it should not fail test?

mmorhun: Yes, sometimes backup/restore can fail with exception, but all what we want to know here is whether backup/restore have been performed. Other tests checks error messages.

plugins/plugin-hosted/codenvy-machine-hosted/src/test/java/com/codenvy/machine/backup/MachineBackupManagerTest.java
```diff
@@ -107,15 +102,9 @@ private void setup() throws ServerException, InterruptedException, IOException,
...
+        doNothing().when(backupManager).execute(anyObject(), anyInt());
 
-        doAnswer(new Answer<Void>() {
-            @Override
-            public Void answer(InvocationOnMock invocation) throws Throwable {
-                sleep(FAKE_BACKUP_TIME_MS);
-                return null;
-            }
-        }).when(backupManager).execute(anyObject(), anyInt());
+        cmdCaptor = ArgumentCaptor.forClass(String[].class);
```
sleshchenko: Why should you create captor manually? @Captor annotation for cmdCaptor filed doesn't work for this?

garagatyi:
LGTM

plugins/plugin-hosted/codenvy-machine-hosted/src/test/java/com/codenvy/machine/backup/MachineBackupManagerTest.java
mmorhun: I create it in before test method just for case. It can guarantee, that each test will use its own captor object.

sleshchenko:
OK for me

Issue #624, opened 2016-08-25
Changing of hostname of multi-server Codenvy on-prem hung up


dmytro-ndp:
Integration test "[test-install-multi-nodes-and-change-config.sh](https://github.com/codenvy/codenvy/blob/4.6.2/cdec/installation-manager-tests/src/main/resources/bin/install/test-install-multi-nodes-and-change-config.sh)" hang up on step:

Command which hung up:
> TEST: === [ Thu Aug 18 16:41:26 EEST 2016 ] COMMAND STARTED: executeIMCommand config --hostname test.codenvy.onprem

Codenvy on-prem server was alive:

> curl -X OPTIONS -v "http://localhost/api/"
> * About to connect() to localhost port 80 (#0)
> *   Trying 127.0.0.1...
> * Connection refused
> * Failed connect to localhost:80; Connection refused
> * Closing connection 0
> curl: (7) Failed connect to localhost:80; Connection refused

Codenvy CLI log:
```
2016-08-23 22:27:16 +0100 [INFO] com.codenvy.im.commands.SimpleCommand execute {'command'='if ! sudo grep -Eq "^127.0.0.1.*\stest.codenvy.onprem\s.*$|^127.0.0.1.*\stest.codenvy.onprem$"
 /etc/hosts; then
  echo "
127.0.0.1 test.codenvy.onprem" | sudo tee --append /etc/hosts > /dev/null
fi', 'agent'='LocalAgent'}
2016-08-23 22:27:16 +0100 [INFO] com.codenvy.im.commands.SimpleCommand execute {'command'='sudo cp /etc/puppet/manifests/nodes/multi_server/base_configurations.pp /etc/puppet/manifests/
nodes/multi_server/base_configurations.pp.back ; sudo cp /etc/puppet/manifests/nodes/multi_server/base_configurations.pp /etc/puppet/manifests/nodes/multi_server/base_configurations.pp.
back.1998979570172 ; ', 'agent'='LocalAgent'}
2016-08-23 22:27:17 +0100 [INFO] com.codenvy.im.commands.SimpleCommand execute {'command'='sudo cat /etc/puppet/manifests/nodes/multi_server/base_configurations.pp | sed ':a;N;$!ba;s/\n
[...]

PR #626, opened 2016-08-25
(#624) don't change /etc/hosts when changing host_url of multi-server…

dmytro-ndp:
Don't change /etc/hosts when changing host_url of multi-server Codenvy on-prem.


riuvshin:
OK

Issue #515, opened 2016-08-04
Add ability to save workspace snapshots in AWS ECR


mmorhun:
AWS ECR (unlike many other registries such as dockerhub) [do not support](http://docs.aws.amazon.com/AmazonECR/latest/userguide/repository-create.html) repository creation on docker push operation. This cause snapshot creation fail if AWS ECR is configured as registry for storing workspace snapshots.

mmorhun:
Well, here is another way of implementation:
1) Interface `DockerRegistry` with following methods:
- saveImageIntoNewRepository
- removeRepository
- isMatch

2) Class `DockerRegistryProvider` which will return `DockerRegistry` implementation for given registry.
3) Classes for each specific registry (implements `DockerRegistry`).
4) Class `DefaultDockerRegistry`which will be returned by `DockerRegistryProvider`, if no one of custom registries implementation matches.

Each time, when we need some interaction with a docker registry we'll use the provider which will return corresponding implementation. After that, common commands for interaction.

mmorhun:
@skabashnyuk ?

mmorhun:
Is not needed any more

PR #629, opened 2016-08-25
CODENVY-515: Add ability to save workspace snapshots in AWS ECR

mmorhun:
### What does this PR do?
Adds interceptor for snapshot pushing to docker registry which in case of AWS ECR creates repository for the snapshot before pushing.

#515 

### Previous Behavior
No ability to save snapshots in AWS ECR.

### New Behavior
Added ability to save workspace snapshots in AWS ECR.

### Tests written?
Yes

### DIFF
--- a/plugins/plugin-hosted/codenvy-machine-hosted/src/main/java/com/codenvy/machine/interceptor/SnapshottingToRegistryInterceptor.java
+++ b/plugins/plugin-hosted/codenvy-machine-hosted/src/main/java/com/codenvy/machine/interceptor/SnapshottingToRegistryInterceptor.java
@@ -1,0 +1,78 @@
+/*
+ *  [2012] - [2016] Codenvy, S.A.
+ *  All Rights Reserved.
+ *
+ * NOTICE:  All information contained herein is, and remains
+ * the property of Codenvy S.A. and its suppliers,
+ * if any.  The intellectual and technical concepts contained
+ * herein are proprietary to Codenvy S.A.
+ * and its suppliers and may be covered by U.S. and Foreign Patents,
+ * patents in process, and are protected by trade secret or copyright law.
+ * Dissemination of this information or reproduction of this material
+ * is strictly forbidden unless prior written permission is obtained
+ * from Codenvy S.A..
+ */
+package com.codenvy.machine.interceptor;
+
+import com.amazonaws.auth.AWSCredentials;
+import com.amazonaws.auth.BasicAWSCredentials;
+import com.amazonaws.services.ecr.AmazonECRClient;
+import com.amazonaws.services.ecr.model.CreateRepositoryRequest;
+import com.codenvy.auth.aws.ecr.AwsInitialAuthConfig;
+import com.google.common.annotations.VisibleForTesting;
+import com.google.inject.Inject;
+import com.google.inject.Singleton;
+
+import org.aopalliance.intercept.MethodInterceptor;
+import org.aopalliance.intercept.MethodInvocation;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.regex.Pattern;
+
+/**
+ * This interceptor creates repository in remote registry if it doesn't support repository creation on docker push command.
+ *
+ * @author Mykola Morhun
+ */
+@Singleton
+public class SnapshottingToRegistryInterceptor implements MethodInterceptor {
+    private static final Logger LOG = LoggerFactory.getLogger(SnapshottingToRegistryInterceptor.class);
+
+    // https://aws_account_id.dkr.ecr.region.amazonaws.com
+    private static final String AMAZON_ECR_URL_REG_EXP = "^\\d{12}\\.dkr\\.ecr\\..+-\\d\\.amazonaws\\.com$";
+    private static final Pattern AMAZON_ECR_URL_PATTERN = Pattern.compile(AMAZON_ECR_URL_REG_EXP);
+
+    @Inject
+    private AwsInitialAuthConfig awsInitialAuthConfig;
+
+    @Override
+    public Object invoke(MethodInvocation methodInvocation) throws Throwable {
+        Object[] arguments = methodInvocation.getArguments();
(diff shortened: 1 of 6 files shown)


mmorhun:
@akorneta @evoevodin review please

plugins/plugin-hosted/codenvy-machine-hosted/src/main/java/com/codenvy/machine/interceptor/SnapshottingToRegistryInterceptor.java
```diff
@@ -0,0 +1,78 @@
...
+
+import java.util.regex.Pattern;
+
+/**
+ * This interceptor creates repository in remote registry if it doesn't support repository creation on docker push command.
+ *
+ * @author Mykola Morhun
+ */
+@Singleton
+public class SnapshottingToRegistryInterceptor implements MethodInterceptor {
```
evoevodin: It is not clear for me, why the interceptor is used for that purpose.
It looks like this logic is pretty common to registries, can you please argument the interceptor's choice ?
I'd say that it is better to think on components/interfaces for doing some actions before the push which may also define another useful methods for authentication and stuff like that.

mmorhun: @evoevodin , I used interceptor here because AWS ECR is specific registry. We should use AWS API to create repository before pushing and it is needed only for AWS.

evoevodin: My point is that, interceptor is non extensible/implicit solution for more like explicit problem.
What if we need to support another registry type, are you sure that AWS ECR is the only registry we'll need to support?
I thing that the logic related to registries is pretty straightforward and needs a set of components:
- it is easy to extend it
- it is easy to control execution flow
- it becomes a part of infrastructure which makes it easier to understand/document/maintain 
- It becomes easy to add a plugin which defines a new registry or event modifies a logic of existing one

evoevodin: @akorneta, @garagatyi, @skabashnyuk i need your points on that

skabashnyuk: +1 to avoid usage of interceptors as much as possible.

akorneta: +1 to evoevodin because current approach is not pluggable and hard to extends in case if any other registry with different behavior will be added.

mmorhun: @akorneta interceptors are pluggable. We can easily add/remove it. Yes, it is hard to extend interceptor logic, but:
- if a registry has some different behaviour we need to implement it anyway(or refuse support that specific registry). The question is where this implementation will be written in an interceptor or in the main code, which makes main code more complicated and harder to maintaining and understanding.
- we needn't to extend behaviour of specific registry. In case of AWS ECR we just use AWS API. If some changes are needed, we'll just change existing code. We cannot make extension to foreign API.

mmorhun: > What if we need to support another registry type, are you sure that AWS ECR is the only registry we'll need to support?

@evoevodin please note, that so far _only_ AWS ECR is different. Other registries is user in common way. Theoretically it is possible to have another specific registry, but actually what is probability of it?
More, as I said above, each non-common thing in main code are very bad (I think you are agree with this, aren't you?). And even if we have another specific registry we'll just add new case in existing interceptor (or can add new one, but I think that it will be not obvious, that makes code worse).

mmorhun: @garagatyi , @tolusha WDYT?

evoevodin: > if a registry has some different behaviour we need to implement it anyway(or refuse support that specific registry). The question is where this implementation will be written in an interceptor or in the main code, which makes main code more complicated and harder to maintaining and understanding.

i don't agree on that, what you did for now:
from the `saveToSnanshot` method you separated a half of the code to the another method, which is not needed for a regular class client and then you created an interceptor for that method, and you say that it is clear for understanding and simple for maintaining? 

> will be written in an interceptor or in the main code

We suggested you to create a set of components(abstractions) which are designed to handle common situations and then implement these components for  the registries we have.
I don't see any problem of the "main code" calling method like `beforePush(..)` obviously before pushing the image, do you?

tolusha: Interceptor is ok for me here.

gazarenkov: I do not think interceptor is the choice here.
-1

evoevodin:
@mmorhun Please rework the current approach without using interceptors, the logic related to the certain registry must be pluggable with the machine related components,  concrete registry plugin should be selected in runtime depending on the context . The contract between registry plugins and machine components must be clearly defined and documented. 

If you have any questions feel free to ask.
Please let me know when the model of components is ready for review, thank you!

PR #671, opened 2016-08-31
Installation manager minor fixup

dmytro-ndp:

Request contains:
1) fixup of integration tests which produces error 'ss: command not found';
2) turning off post-flight codenvy installation check because of an issue https://github.com/eclipse/che/issues/2284 which causes an error:
post-flight_check_failure


riuvshin:
+1

PR #678, opened 2016-09-02
CHE-1818: add new WS environment types for single machine

garagatyi:
Adds support of new single machine environment types.


skabashnyuk:
ok

PR #682, opened 2016-09-03
CHE-1818: change link to factory templates to new templates

garagatyi:


garagatyi:
@skabashnyuk @benoitf @riuvshin Please review

PR #713, opened 2016-09-09
CHE-2331: fix in accordance with changes in Che

garagatyi:
### What does this PR do?
Fix classes in accordance with changes in eclipse/che#2392

eclipse/che#2392

### Tests written?
No


skabashnyuk:
ok

PR #718, opened 2016-09-09
Separate queries for getting permissions for specific user and public…

mkuznyetsov:
### What does this PR do?
Remove the case, when getting permission by user and instance ID would also return public permission


akorneta:
ok

mshaposhnik:
Ok

PR #728, opened 2016-09-12
CODENVY-499: Ways to mount folder inside host machine to auxiliary machine

tolusha:
### What does this PR do?
Add ability to mount extra volumes in Codenvy


tolusha:
@sleshchenko 
@garagatyi 
pls review

benoitf:
what's the difference with `machine.server.extra.volume` property ?

benoitf:
it's backporting https://github.com/eclipse/che/blob/b2309ef34510be5dd00fa6d37dc1e5bf69b5dff1/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/local/LocalDockerModule.java#L68-L71 to codenvy ?

tolusha:
`machine.server.extra.volume` is a property for user to set extra volumes to mount.
`machine.docker.dev_machine.machine_volume` is used inside che to collect all mounts

garagatyi:
Please change title and commit message: CODENVY-499, not CHE-499

garagatyi:
Looks like binding for dev machine is not needed. Use binding for all machines instead.

garagatyi:
Also, please, move provider in Che out of local package and fix javadocs to not declare binding for dev machines only.

tolusha:
@garagatyi 
updated

garagatyi:
LGTM

skabashnyuk:
щл

PR #731, opened 2016-09-13
CODENVY-581: Add Audit service

vinokurig:
### Add Audit seervice to be able to generate audit report from codenvy im cli client

https://github.com/codenvy/codenvy/issues/581

### Tests written?
Yes

@sleshchenko @tolusha


wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditService.java
```diff
@@ -0,0 +1,170 @@
...
+import java.text.SimpleDateFormat;
+import java.util.List;
+import java.util.Optional;
+
+/**
+ * Defines Audit report REST API.
+ *
+ * @author Igor Vinokur
+ */
+@Path("/audit")
```
tolusha: Do we need swagger annotations?

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditService.java
```diff
@@ -0,0 +1,170 @@
...
+                        CodenvyLicenseManager licenseManager) {
+        this.adminUserDao = adminUserDao;
+        this.workspaceManager = workspaceManager;
+        this.permissionManager = permissionManager;
+        this.licenseManager = licenseManager;
+    }
+
+    @GET
+    @Produces(MediaType.APPLICATION_OCTET_STREAM)
+    public StreamingOutput getReport() throws ServerException {
```
tolusha: Method is too complex. It is barely possible to understand the logic.

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditService.java
vinokurig: I think no. The response will be a process of downloading a file. I think it will look strange in swagger

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditService.java
vinokurig: divided the logic to separate methods

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditService.java
```diff
@@ -0,0 +1,194 @@
...
+        } catch (LicenseException e) {
+            appendToFile("Failed to retrieve license!\n", file);
+            LOG.error(e.getMessage(), e);
+        }
+    }
+
+    private void printAllUsersInfo(File file) throws ServerException {
+        int skipItems = 0;
+        while (true) {
+            List<UserImpl> users = adminUserDao.getAll(20, skipItems).getItems();
```
sleshchenko: I think you should use user manager here

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditService.java
```diff
@@ -0,0 +1,194 @@
...
+                    "Date when license expires: " + new SimpleDateFormat("dd MMMM yyyy").format(license.getExpirationDate()) + "\n", file);
+        } catch (LicenseException e) {
+            appendToFile("Failed to retrieve license!\n", file);
+            LOG.error(e.getMessage(), e);
+        }
+    }
+
+    private void printAllUsersInfo(File file) throws ServerException {
+        int skipItems = 0;
+        while (true) {
```
sleshchenko: try to rewrite this with Page#getNextPageRef

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditServicePermissionsFilter.java
```diff
@@ -0,0 +1,48 @@
...
+ * @author Igor Vinokur
+ */
+@Filter
+@Path("/audit")
+public class AuditServicePermissionsFilter extends CheMethodInvokerFilter {
+    @Override
+    protected void filter(GenericResourceMethod GenericResourceMethod, Object[] arguments) throws ApiException {
+        String methodName = GenericResourceMethod.getMethod().getName();
+        if ("getReport".equals(methodName)) {
+            EnvironmentContext.getCurrent().getSubject().checkPermission(SystemDomain.DOMAIN_ID, null, MANAGE_USERS_ACTION);
```
sleshchenko: Are you sure that this method should be covered by MANAGE_USERS_ACTION?

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditService.java
```diff
@@ -0,0 +1,194 @@
...
+        this.permissionManager = permissionManager;
+        this.licenseManager = licenseManager;
+    }
+
+    @GET
+    @Produces(MediaType.APPLICATION_OCTET_STREAM)
+    public StreamingOutput getReport() throws ServerException {
+        File file = initializeFileInTempDirectory();
+
+        printHeader(file);
```
sleshchenko: Will be file removed when ServerException is thrown here?

sleshchenko: I'd move all business logic related to report generation out of `Service` because Service's responsibility is just transport (parsing request, bean validation of parameters, forming response). And report generating is responsibility of class in next layout like `ReportManager`

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditService.java
vinokurig: UserManager does not have `getAll()` method

sleshchenko: But will have when jpa-integration will be merged into master. Add TODO

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditService.java
vinokurig: it returns `pageRef` object and I can't receive items from it

sleshchenko: But you can get itemsBefore aka skipCount and pageSize aka maxItems for next page

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditServicePermissionsFilter.java
vinokurig: It was a mistake, changed to MANAGE_CODENVY_ACTION

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditService.java
vinokurig: added exception catch and moved the logic to `AuditManager`

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditService.java
```diff
@@ -0,0 +1,57 @@
...
+
+    @Inject
+    public AuditService(AuditManager auditManager) {
+        this.auditManager = auditManager;
+    }
+
+    @GET
+    @Produces(MediaType.APPLICATION_OCTET_STREAM)
+    public StreamingOutput downloadReport() throws ServerException {
+        final File report = auditManager.initializeFileReportInTempDirectory();
```
sleshchenko: I would rework in next way:
AuditManager returns file which contains report content
Service construct response with stream of this file and in finally block remove it mb by FileCleaner

vinokurig: The response is a separate thread. It executes after finally block when the file will be deleted.

sleshchenko: OK. BTW initializeFileReportInTempDirectory is not public responsibility of auditManager. auditManager should only generate report

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditService.java
```diff
@@ -0,0 +1,57 @@
...
+
+        auditManager.printAuditReportToFile(report);
+
+        return output -> {
+            output.write(IOUtils.toByteArray(new FileInputStream(report)));
```
sleshchenko: write can throw IOException. Will be file removed in this case?

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditService.java
vinokurig: removed delating from manager

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditService.java
vinokurig: reworked

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditManager.java
```diff
@@ -0,0 +1,188 @@
...
+    private final PermissionManager     permissionManager;
+    private final CodenvyLicenseManager licenseManager;
+
+    @Inject
+    public AuditManager(AdminUserDao adminUserDao,
+                        WorkspaceManager workspaceManager,
+                        PermissionManager permissionManager,
+                        CodenvyLicenseManager licenseManager) {
+
+        //TODO: use getAll() method from JpaUserDao, when it will be merged to master.
```
sleshchenko: Do you mean getAll() method from UserManager?

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditManager.java
```diff
@@ -0,0 +1,188 @@
...
+        this.adminUserDao = adminUserDao;
+        this.workspaceManager = workspaceManager;
+        this.permissionManager = permissionManager;
+        this.licenseManager = licenseManager;
+    }
+
+    /**
+     * Prints audit report to given file.
+     *
+     * @param file file for report
```
sleshchenko: file file file file
P.S. And fix formatting here

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditManager.java
```diff
@@ -0,0 +1,188 @@
...
+                    "Date when license expires: " + new SimpleDateFormat("dd MMMM yyyy").format(license.getExpirationDate()) + "\n", file);
+        } catch (LicenseException e) {
+            printRowToReport("Failed to retrieve license!\n", file);
+            LOG.error(e.getMessage(), e);
+        }
+    }
+
+    private void printAllUsersInfo(File file) throws ServerException {
+        int skipItems = 0;
+        while (true) {
```
sleshchenko: What about reworking with PageRef?

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditManager.java
```diff
@@ -0,0 +1,188 @@
...
+        List<PermissionsImpl> permissions;
+        permissions = permissionManager.getByInstance("workspace", workspaceId);
+        Optional<PermissionsImpl> optional = permissions.stream()
+                                                        .filter(wsPermissions -> wsPermissions.getUser().equals(userId))
+                                                        .findFirst();
+
+        if (optional.isPresent()) {
+            return optional.get();
+        }
+        else {
```
sleshchenko: wrong formatting

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditManager.java
vinokurig: fixed

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditManager.java
vinokurig: fixed

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditManager.java
vinokurig: fixed

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditService.java
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    @GET
+    @Produces(MediaType.APPLICATION_OCTET_STREAM)
+    public Response downloadReport() throws ServerException {
+        File report = auditManager.generateAuditReport();
+
+        StreamingOutput stream = output -> {
+            try {
+                output.write(IOUtils.toByteArray(new FileInputStream(report)));
```
tolusha: We shouldn't copy everything into bytearray. OOM might happens.

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditManager.java
```diff
@@ -0,0 +1,196 @@
...
+        this.licenseManager = licenseManager;
+    }
+
+    /**
+     * Prints audit report to given file.
+     *
+     * @throws ServerException
+     *         if an error occurs
+     */
+    public File generateAuditReport() throws ServerException {
```
tolusha: Methods mustn't allow generate a report if a previous one hasn't been finished.

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditManager.java
```diff
@@ -0,0 +1,196 @@
...
+        String dateTime = new SimpleDateFormat("dd/MM/yyyy-HH:mm:ss").format(new Date());
+        final File report = new File(Files.createTempDir(), "report_" + dateTime + ".txt");
```
tolusha: Use ```Path``` instead of ```File```

tolusha: Add ```audit``` into a file name

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditManager.java
```diff
@@ -0,0 +1,196 @@
...
+        } catch (IOException e) {
+            LOG.error(e.getMessage(), e);
+            throw new ServerException("Failed to generate audit report. " + e.getMessage(), e);
+        }
+    }
+
+    private PermissionsImpl getWorkspacePermissions(String workspaceId, String userId)
+            throws ServerException, NotFoundException, ConflictException {
+        List<PermissionsImpl> permissions;
+        permissions = permissionManager.getByInstance("workspace", workspaceId);
```
tolusha: Inline ```permissions``` variable

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditManager.java
```diff
@@ -0,0 +1,196 @@
...
+        Optional<PermissionsImpl> optional = permissions.stream()
+                                                        .filter(wsPermissions -> wsPermissions.getUser().equals(userId))
+                                                        .findFirst();
+
+        if (optional.isPresent()) {
+            return optional.get();
+        } else {
+            throw new NotFoundException("Permissions for user " + userId + " in workspace " + workspaceId +
```
tolusha: We mustn't throw an exception. If error occurs then we will never generate a report until bug is fixed. 
Just return fake permissions which explains that permissions not found.

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditManager.java
tolusha: Why do we need create a temp directory? Let's use default temp directory.

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditManager.java
```diff
@@ -0,0 +1,196 @@
...
+                                        " was not found while generating audit report");
+        }
+    }
+
+    @VisibleForTesting
+    void deleteReport(File report) {
```
tolusha: Use ```Path``` here and just delete a single file and not a whole directory.

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditManager.java
```diff
@@ -0,0 +1,196 @@
...
+            printRowToReport(getWorkspacePermissions(workspace.getId(), user.getId()).getActions().toString() + "\n", file);
+        } catch (NotFoundException e) {
+            printRowToReport("Failed to retrieve workspace permissions! " + e.getMessage() + "\n", file);
+        } catch (ConflictException e) {
+            LOG.error(e.getMessage(), e);
+            printRowToReport("Failed to retrieve workspace Id!\n", file);
+        }
+    }
+
+    private void printRowToReport(String row, File report) throws ServerException {
```
tolusha: ```printRow``` is enough.

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditManager.java
```diff
@@ -0,0 +1,196 @@
...
+        }
+    }
+
+    private void printUserWorkspaceInfo(WorkspaceImpl workspace, UserImpl user, File file) throws ServerException {
+        printRowToReport("   └ " + workspace.getConfig().getName() +
+                         ", is owner: " + workspace.getNamespace().equals(user.getName()) + ", permissions: ", file);
+        try {
+            printRowToReport(getWorkspacePermissions(workspace.getId(), user.getId()).getActions().toString() + "\n", file);
+        } catch (NotFoundException e) {
+            printRowToReport("Failed to retrieve workspace permissions! " + e.getMessage() + "\n", file);
```
tolusha: Log error and no error detail in the report.

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditManager.java
```diff
@@ -0,0 +1,196 @@
...
+        } while (currentPage != null);
+    }
+
+    private void printUserInfo(UserImpl user, File file) throws ServerException {
+        //Print error if failed to retrieve the list of his workspaces
+        List<WorkspaceImpl> workspaces;
+        try {
+            workspaces = workspaceManager.getWorkspaces(user.getId());
+        } catch (ServerException e) {
+            printRowToReport("Failed to receive list of related workspaces for user " + user.getEmail() + "!\n", file);
```
tolusha: Add [ERROR] prefix here and in other places when a such kind of info is printed.

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditManager.java
```diff
@@ -0,0 +1,196 @@
...
+import java.util.List;
+import java.util.Optional;
+
+/**
+ * Facade for audit report related operations.
+ *
+ * @author Igor Vinokur
+ */
+@Singleton
+public class AuditManager {
```
tolusha: A general idea: separate a report printing logic and a retrieving business data.

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditManager.java
vinokurig: Actually the report WILL be generated. The exception is caught in `printUserWorkspaceInfo` method, it prints error in the row where permissions are expected to be.

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditManager.java
vinokurig: Theoretically two threads can create files with same names. It can happen when one thread finished generating report but not removed the file, in the same second other request will use the old file and conflict will appear when first thread will delete the file.

vinokurig:
@tolusha @sleshchenko fixed all your notes, please review

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditManager.java
```diff
@@ -0,0 +1,168 @@
...
+
+    private static final Logger LOG = LoggerFactory.getLogger(AuditService.class);
```
sleshchenko: Mb you mean `AuditManager.class`?

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditReportPrinter.java
```diff
@@ -0,0 +1,174 @@
...
+     * @param auditReport
+     *         file of audit report
+     * @param allUsersNumber
+     *         all users number
+     * @param license
+     *         {@link CodenvyLicense} object
+     * @throws ServerException
+     *         if an error occurs
+     */
+    void printHeader(Path auditReport, long allUsersNumber, CodenvyLicense license) throws ServerException {
```
tolusha: Add @Nullable annotation whenever parameter can be null

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditManager.java
```diff
@@ -0,0 +1,168 @@
...
+     * @throws ServerException
+     *         if an error occurs
+     * @throws ConflictException
+     *         if generating report is not finished by other user
+     * @throws IOException
+     *         if failed to create file for audit report
+     */
+    public Path generateAuditReport() throws ServerException, ConflictException, IOException {
+        if (inProgress.getAndSet(true)) {
+            throw new ConflictException("This command has been running by other user");
```
sleshchenko: Why should we limit users to generate only one report at the same time?

tolusha: * to prevent launching report generation by mistake
* to reduce burden on API

sleshchenko: Then mb it would be better to limit one report by one user at the same time?

tolusha: It won't be better :)

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditManager.java
vinokurig: fixed

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditReportPrinter.java
vinokurig: done

tolusha:
LGTM

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditManager.java
sleshchenko: OK

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditManager.java
```diff
@@ -0,0 +1,186 @@
...
+        }
+
+        String dateTime = new SimpleDateFormat("dd-MM-yyyy_hh:mm:ss").format(new Date());
```
sleshchenko: I think you should move it to try block

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditManager.java
```diff
@@ -0,0 +1,186 @@
...
+    void deleteReport(Path auditReport) {
+        try {
+            FileUtils.deleteDirectory(new File(auditReport.toFile().getParent()));
+        } catch (IOException exception) {
+            LOG.error(exception.getMessage(), exception);
+        }
+    }
+
+
+    public static void main(String[] args) {
```
sleshchenko: remove it here

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditManager.java
```diff
@@ -0,0 +1,186 @@
...
+     * @return path of audit report file
+     * @throws ServerException
+     *         if an error occurs
+     * @throws ConflictException
+     *         if generating report is not finished by other user
+     * @throws IOException
+     *         if failed to create file for audit report
+     */
+    public Path generateAuditReport() throws ServerException, ConflictException, IOException {
+        if (inProgress.getAndSet(true)) {
```
sleshchenko: I think it would be more logical to use compareAndSet here

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditManager.java
vinokurig: fixed

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditManager.java
vinokurig: done

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditManager.java
vinokurig: done

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditManager.java
vinokurig: done

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditManager.java
```diff
@@ -0,0 +1,174 @@
...
+            auditReport = Paths.get(createTempDir().getAbsolutePath(), "/report_" + dateTime + ".txt");
+            Files.createFile(auditReport);
+
+            CodenvyLicense license = null;
+            try {
+                license = licenseManager.load();
+            } catch (LicenseException ignored) {
+                //Continue printing report without license info
+            }
+            reportPrinter.printHeader(auditReport, userDao.getTotalCount(), license);
```
sleshchenko: Use https://github.com/eclipse/che/blob/master/wsmaster/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/UserManager.java#L238 instead of method in user dao

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditManager.java
```diff
@@ -0,0 +1,174 @@
...
+     */
+    public Path generateAuditReport() throws ServerException, ConflictException {
+        if (!inProgress.compareAndSet(false, true)) {
+            throw new ConflictException("This command is being run by other user");
+        }
+
+        Path auditReport = null;
+        try {
+            String dateTime = new SimpleDateFormat("dd-MM-yyyy_hh:mm:ss").format(new Date());
+            auditReport = Paths.get(createTempDir().getAbsolutePath(), "/report_" + dateTime + ".txt");
```
sleshchenko: Why createTempDir com.google.common.io instead of java.nio.file.Files#createTempDirectory?

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditManager.java
```diff
@@ -0,0 +1,174 @@
...
+            CodenvyLicense license = null;
+            try {
+                license = licenseManager.load();
+            } catch (LicenseException ignored) {
+                //Continue printing report without license info
+            }
+            reportPrinter.printHeader(auditReport, userDao.getTotalCount(), license);
+            printAllUsers(auditReport);
+        } catch (Exception exception) {
+            deleteReport(auditReport);
```
sleshchenko: What will happen in case invoking deleteReport with null parameter?

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditReportPrinter.java
```diff
@@ -0,0 +1,175 @@
...
+     * @param user
+     *         {@link UserImpl} object
+     * @param workspaces
+     *         list of workspaces that are related to given user
+     * @param wsPermissions
+     *         map of permissions to workspaces
+     * @throws ServerException
+     *         if an error occurs
+     */
+    void printUserInfoWithHisWorkspacesInfo(Path auditReport, UserImpl user, @Nullable List<WorkspaceImpl> workspaces,
```
sleshchenko: Do you invoke this method with null list?

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditReportPrinter.java
```diff
@@ -0,0 +1,175 @@
...
+     * .
+     * .
+     * └ <workspace name>, is owner: <true or false>, permissions: <list of permissions for current user>
+     * <p>
+     * If the list of user workspaces is not defined it prints error message.
+     *
+     * @param auditReport
+     *         file of audit report
+     * @param user
+     *         {@link UserImpl} object
```
sleshchenko: I don't like comments like this because it says nothing about how this value will be used, it just says about parameter type and it just duplicates information from method's signature

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditReportPrinter.java
```diff
@@ -0,0 +1,175 @@
...
+                                            Map<String, List<AbstractPermissions>> wsPermissions)
+            throws ServerException {
+
+        if (workspaces == null) {
+            printError("Failed to receive list of related workspaces for user " + user.getId(), auditReport);
+            return;
+        }
+
+        int workspacesNumber = workspaces.size();
+        long ownWorkspacesNumber = workspaces.stream().filter(workspace -> workspace.getNamespace().equals(user.getName())).count();
```
sleshchenko: It can be false in case when user is owner of workspace but doesn't have direct permissions

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditReportPrinter.java
```diff
@@ -0,0 +1,175 @@
...
+     * [ERROR] <error text>!
+     *
+     * @param error
+     *         text of error
+     * @param auditReport
+     *         file of audit report
+     * @throws ServerException
+     *         if an error occurs
+     */
+    void printError(String error, Path auditReport) throws ServerException {
```
sleshchenko: Why Path is better than File here?

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditManager.java
```diff
@@ -0,0 +1,174 @@
...
+                    }
+                }
+                reportPrinter.printUserInfoWithHisWorkspacesInfo(auditReport, user, workspaces, wsPermissions);
+            }
+            //Initialize next page if exist, otherwise stop printing report
+            final Optional<PageRef> nextPageRefOpt = currentPage.getNextPageRef();
+            if (nextPageRefOpt.isPresent()) {
+                final PageRef nextPageRef = nextPageRefOpt.get();
+                final long itemsBefore = nextPageRef.getItemsBefore();
+                if (itemsBefore < Integer.MAX_VALUE) {
```
sleshchenko: I think you should move this logic to another place like user manager or mb page

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditManager.java
```diff
@@ -0,0 +1,179 @@
...
+            }
+        } while (true);
+    }
+
+    @VisibleForTesting
+    void deleteReportDirectory(Path auditReport) {
+        try {
+            //check that deleted directory is created in system temp directory
+            if (auditReport.startsWith(Paths.get(System.getProperty("java.io.tmpdir")))) {
+                FileUtils.deleteDirectory(new File(auditReport.toFile().getParent()));
```
sleshchenko: Files.delete(auditReport.getParent()); ?

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditManager.java
```diff
@@ -0,0 +1,179 @@
...
+                break;
+            }
+        } while (true);
+    }
+
+    @VisibleForTesting
+    void deleteReportDirectory(Path auditReport) {
+        try {
+            //check that deleted directory is created in system temp directory
+            if (auditReport.startsWith(Paths.get(System.getProperty("java.io.tmpdir")))) {
```
sleshchenko: When it can be created outside of system temp directory? And why we should not remove report in this case?

vinokurig: If somehow the report will be not in temp directory but in root for example, it will delete all parent directory

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditManager.java
vinokurig: changed

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditManager.java
vinokurig: done

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditManager.java
vinokurig: added a check for null

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditManager.java
vinokurig: created issue for that, added todo

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditReportPrinter.java
vinokurig: updated description

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditReportPrinter.java
vinokurig: no, fixed

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditReportPrinter.java
vinokurig: fixed

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditReportPrinter.java
vinokurig: It is more popular :)

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditManager.java
vinokurig: done

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditManager.java
vinokurig: removed this logic

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditManager.java
```diff
@@ -0,0 +1,178 @@
...
+            for (UserImpl user : currentPage.getItems()) {
+                List<WorkspaceImpl> workspaces;
+                int permissionsNumber;
+                try {
+                    workspaces = workspaceManager.getWorkspaces(user.getId());
+                    permissionsNumber = workspaces.size();
+                    //add workspaces witch are belong to user, but user doesn't have permissions for them.
+                    workspaceManager.getByNamespace(user.getName())
+                                    .stream()
+                                    .filter(workspace -> !workspaces.contains(workspace))
```
sleshchenko: I think you should compare ids here, because in your case result list may contain the same workspace twice when workspace will be updated between your invoking of `getWorkspaces` and `getByNamespace`

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditManager.java
```diff
@@ -0,0 +1,178 @@
...
+                                    .forEach(workspaces::add);
+                } catch (ServerException exception) {
+                    reportPrinter.printError("Failed to receive list of related workspaces for user " + user.getId(), auditReport);
+                    continue;
+                }
+                Map<String, List<AbstractPermissions>> wsPermissions = new HashMap<>();
+                for (WorkspaceImpl workspace : workspaces) {
+                    try {
+                        wsPermissions.put(workspace.getId(), permissionsManager.getByInstance("workspace", workspace.getId()));
```
sleshchenko: As far as I understand you can use [PermissionsManager#get](https://github.com/codenvy/codenvy/blob/master/wsmaster/codenvy-hosted-api-permission/src/main/java/com/codenvy/api/permission/server/PermissionsManager.java#L131-131) method to avoid manually filtering of permissions [here](https://github.com/codenvy/codenvy/pull/731/files#diff-6434c99aa033c523d806f289c2c44da5R156)

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditManager.java
```diff
@@ -0,0 +1,178 @@
...
+                final long itemsBefore = nextPageRef.getItemsBefore();
+                //TODO: fix when https://github.com/eclipse/che/issues/2524 will be resolved
+                if (itemsBefore < Integer.MAX_VALUE) {
+                    throw new ServerException("Skip count limit was reached while retrieving all users");
+                }
+                currentPage = userManager.getAll(nextPageRef.getPageSize(), (int)itemsBefore);
+            } else {
+                break;
+            }
+        } while (true);
```
sleshchenko: I would move 
```
+            //Initialize next page if exist, otherwise stop printing report
 +            final Optional<PageRef> nextPageRefOpt = currentPage.getNextPageRef();
 +            if (nextPageRefOpt.isPresent()) {
 +                final PageRef nextPageRef = nextPageRefOpt.get();
 +                final long itemsBefore = nextPageRef.getItemsBefore();
 +                //TODO: fix when https://github.com/eclipse/che/issues/2524 will be resolved
 +                if (itemsBefore < Integer.MAX_VALUE) {
 +                    throw new ServerException("Skip count limit was reached while retrieving all users");
 +                }
 +                currentPage = userManager.getAll(nextPageRef.getPageSize(), (int)itemsBefore);
 +            } else {
 +                break;
 +            }
```
to private method and rewrite cycle as 
```
do {
...
} while ((currentPage = doGetNextPage) != null);
```

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditManager.java
```diff
@@ -0,0 +1,178 @@
...
+     *
+     * @return path of audit report file
+     * @throws ServerException
+     *         if an error occurs
+     * @throws ConflictException
+     *         if generating report is being run by other user
+     */
+    public Path generateAuditReport() throws ServerException, ConflictException, IOException {
+        if (!inProgress.compareAndSet(false, true)) {
+            throw new ConflictException("This command is being run by other user");
```
vinokurig: Updated error message

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditManager.java
vinokurig: done

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditManager.java
vinokurig: Yes, but `AuditReportPrinter` shouldn't do any requests. It only recessives all needed data and prints it.

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditManager.java
vinokurig: done

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditManager.java
sleshchenko: I don't say that you should do it in `AuditReportPrinter`. I just say that you can give `AuditReportPrinter` only one permissions objects instead of `List`

vinokurig: reworked to PermissionsManager#get method

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditManager.java
```diff
@@ -0,0 +1,186 @@
...
+                }
+                List<AbstractPermissions> wsPermissions = new ArrayList<>();
+                for (WorkspaceImpl workspace : workspaces) {
+                    try {
+                        wsPermissions.add(permissionsManager.get(user.getId(), "workspace", workspace.getId()));
+                    } catch (NotFoundException | ConflictException ignored) {
+                        //Continue printing report, error will be printed in workspace permissions row
+                    }
+                }
+                reportPrinter.printUserInfoWithHisWorkspacesInfo(auditReport, user, workspaces, permissionsNumber, wsPermissions);
```
sleshchenko: Why should you use permissionsNumber as parameter? ReportPrinter can not use wsPermissions.size()?

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditService.java
```diff
@@ -0,0 +1,55 @@
...
+
+    @Inject
+    public AuditService(AuditManager auditManager) {
+        this.auditManager = auditManager;
+    }
+
+    @GET
+    @Produces(MediaType.APPLICATION_OCTET_STREAM)
+    public Response downloadReport() throws ServerException, ConflictException, IOException {
+        java.nio.file.Path report = auditManager.generateAuditReport();
```
sleshchenko: Who should remove this file?

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditManager.java
vinokurig: fixed

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditService.java
vinokurig: fixed

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditManager.java
```diff
@@ -0,0 +1,183 @@
...
+     *         if an error occurs
+     * @throws ConflictException
+     *         if generating report is already in progress
+     */
+    public Path generateAuditReport() throws ServerException, ConflictException {
+        if (!inProgress.compareAndSet(false, true)) {
+            throw new ConflictException("Generating report is already in progress");
+        }
+
+        Path auditReport = null;
```
sleshchenko: You can remove declaration of this variable here and move return to try section

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditManager.java
```diff
@@ -0,0 +1,183 @@
...
+
+    private void printAllUsers(Path auditReport) throws ServerException {
+        Page<UserImpl> currentPage = userManager.getAll(30, 0);
+        do {
+            //Print users with their workspaces from current page
+            for (UserImpl user : currentPage.getItems()) {
+                List<WorkspaceImpl> workspaces;
+                try {
+                    workspaces = workspaceManager.getWorkspaces(user.getId());
+                    List<String> workspaceIds = workspaces.stream()
```
sleshchenko: Are you sure that you choose right collection for using only contains method?

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditManager.java
```diff
@@ -0,0 +1,183 @@
...
+                } catch (ServerException exception) {
+                    reportPrinter.printError("Failed to receive list of related workspaces for user " + user.getId(), auditReport);
+                    continue;
+                }
+                Map<String, AbstractPermissions> wsPermissions = new HashMap<>();
+                for (WorkspaceImpl workspace : workspaces) {
+                    try {
+                        wsPermissions.put(workspace.getId(), permissionsManager.get(user.getId(), DOMAIN_ID, workspace.getId()));
+                    } catch (NotFoundException | ConflictException e) {
+                        wsPermissions.put(workspace.getId(), null);
```
sleshchenko: Why do you need putting null values here?

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditReportPrinter.java
```diff
@@ -0,0 +1,151 @@
...
+
+    private void printUserWorkspaceInfo(WorkspaceImpl workspace, UserImpl user, Map<String, AbstractPermissions> wsPermissions,
+                                        Path auditReport) throws ServerException {
+        printRow("   └ " + workspace.getConfig().getName() +
+                 ", is owner: " + workspace.getNamespace().equals(user.getName()) + ", permissions: ", auditReport);
+        AbstractPermissions permissions = wsPermissions.get(workspace.getId());
+        if (permissions != null) {
+            printRow(permissions.getActions().toString() + "\n", auditReport);
+        } else {
+            printError("Failed to retrieve workspace permissions", auditReport);
```
sleshchenko: Are you sure that you should print this message when permissions were not found?

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditService.java
```diff
@@ -0,0 +1,63 @@
...
+
+        StreamingOutput stream = outputStream -> {
+            outputStream.write(readAllBytes(report));
```
sleshchenko: Seems that all content of file will be loaded into RAM. Are you sure that it is OK?

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditManager.java
vinokurig: It's used in catch block

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditManager.java
vinokurig: fixed

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditReportPrinter.java
vinokurig: changed to empty list

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditManager.java
vinokurig: Changed List to Set

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditService.java
vinokurig: reworked to `copy(report, outputStream)`

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditService.java
```diff
@@ -0,0 +1,64 @@
...
+        this.auditManager = auditManager;
+    }
+
+    @GET
+    @Produces(MediaType.APPLICATION_OCTET_STREAM)
+    public Response downloadReport() throws ServerException, ConflictException, IOException {
+        java.nio.file.Path report = auditManager.generateAuditReport();
+
+        StreamingOutput stream = outputStream -> {
+            copy(report, outputStream);
```
sleshchenko: Should you remove file in case throwing IOException by copy method?

vinokurig: added deletion in finally block

wsmaster/codenvy-hosted-api-audit/pom.xml
```diff
@@ -0,0 +1,129 @@
...
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>com.jayway.restassured</groupId>
+            <artifactId>rest-assured</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.eclipse.che.core</groupId>
+            <artifactId>che-core-api-jdbc</artifactId>
```
sleshchenko: Why do you need api-jdbc for tests?

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditManager.java
```diff
@@ -0,0 +1,182 @@
...
+        this.workspaceManager = workspaceManager;
+        this.permissionsManager = permissionsManager;
+        this.licenseManager = licenseManager;
+        this.reportPrinter = reportPrinter;
+    }
+
+    /**
+     * Generates audit report.
+     *
+     * @return path of audit report file
```
skabashnyuk: You need to declare that report is plain/text file.

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditService.java
```diff
@@ -0,0 +1,67 @@
...
+
+    private final AuditManager auditManager;
+
+    @Inject
+    public AuditService(AuditManager auditManager) {
+        this.auditManager = auditManager;
+    }
+
+    @GET
+    @Produces(MediaType.APPLICATION_OCTET_STREAM)
```
skabashnyuk: text/plain?

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditManager.java
vinokurig: updated description

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditService.java
vinokurig: This response produces a file and `application/octet-stream` is proper media type for it. I don't think it should be text/plain media type

vinokurig:
@skabashnyuk No I haven't. The report format is pretty simple now. I think that if we have a task for more complicated report we will think about template engine.

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditService.java
skabashnyuk: Are you sure you file has binary format and not human readable text?

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditManager.java
```diff
@@ -0,0 +1,182 @@
...
+                        AuditReportPrinter reportPrinter) {
+        this.userManager = userManager;
+        this.workspaceManager = workspaceManager;
+        this.permissionsManager = permissionsManager;
+        this.licenseManager = licenseManager;
+        this.reportPrinter = reportPrinter;
+    }
+
+    /**
+     * Generates audit report.
```
skabashnyuk: This descriptions tells nothing about report content

sleshchenko:
Seems you missed logback-test.xml

wsmaster/codenvy-hosted-api-audit/src/test/java/com/codenvy/api/audit/server/AuditServicePermissionsFilterTest.java
```diff
@@ -0,0 +1,123 @@
...
+    AuditServicePermissionsFilter filter;
+
+    @Mock
+    private static Subject subject;
+
+    @Mock
+    AuditService service;
+
+    @Test
+    public void shouldCheckPermissionsByUsersFetching() throws Exception {
```
sleshchenko: check this test name

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditService.java
skabashnyuk: this Response.ok(stream, MediaType.TEXT_PLAIN) and this @Produces(MediaType.APPLICATION_OCTET_STREAM)
are not sync

wsmaster/codenvy-hosted-api-audit/src/test/java/com/codenvy/api/audit/server/AuditManagerTest.java
```diff
@@ -0,0 +1,227 @@
...
+import static java.util.Collections.singletonList;
+import static org.apache.commons.io.FileUtils.readFileToString;
+import static org.mockito.Matchers.anyString;
+import static org.mockito.Matchers.eq;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+import static org.testng.Assert.assertEquals;
+
+/**
+ * Tests for {@link AuditService}.
```
sleshchenko: Are you sure that these tests are for `AuditService` ?

wsmaster/codenvy-hosted-api-audit/src/test/java/com/codenvy/api/audit/server/AuditServicePermissionsFilterTest.java
```diff
@@ -0,0 +1,123 @@
...
+        final Response response = given().auth()
+                                         .basic(ADMIN_USER_NAME, ADMIN_USER_PASSWORD)
+                                         .contentType("application/octet-stream")
```
sleshchenko: Are you sure that this method consumes `"application/octet-stream"`. Mb you mean `accept("application/octet-stream")` here?

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditService.java
vinokurig: You are right! Changed to text/plain.

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditManager.java
vinokurig: Updated description

wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditReportPrinter.java
```diff
@@ -0,0 +1,151 @@
...
+import java.text.SimpleDateFormat;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
+import static com.google.common.io.Files.append;
+
+/**
+ * Audit report printer.
+ *
```
vinokurig: Updated description

wsmaster/codenvy-hosted-api-audit/src/test/java/com/codenvy/api/audit/server/AuditServicePermissionsFilterTest.java
vinokurig: fixed

wsmaster/codenvy-hosted-api-audit/src/test/java/com/codenvy/api/audit/server/AuditManagerTest.java
vinokurig: fixed

vinokurig:
@sleshchenko added `logback-test.xml` file

wsmaster/codenvy-hosted-api-audit/pom.xml
```diff
@@ -0,0 +1,121 @@
...
+            <artifactId>everrest-core</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-api</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.codenvy.onpremises.wsmaster</groupId>
+            <artifactId>codenvy-hosted-api-workspace</artifactId>
+            <scope>provided</scope>
```
sleshchenko: Why should it have provided scope?

vinokurig: It is needed for https://github.com/codenvy/codenvy/blob/ec392104ad069fda236ddb347020658eaccafe56/wsmaster/codenvy-hosted-api-audit/src/main/java/com/codenvy/api/audit/server/AuditManager.java#L51
Without provided scope maven will be complaining: `Unused declared dependencies found`

sleshchenko: Mb then it would better to exclude this artifact from checking? like here https://github.com/codenvy/codenvy/blob/master/wsmaster/codenvy-hosted-api-resource/pom.xml#L274-L288

vinokurig: done

Issue #667, opened 2016-08-31
Re-implement IM CLI login command to use Codenyv 4.x on-prem instance as a remote by default


dmytro-ndp:
Current version of IM CLI *login* command is based on Codenvy CLI *login* command which uses *codenvy.com* as a default remote server (which is Codenvy 3.x SaaS), and this command haven't been updated to login to Codenvy on-prem 4.x.

dmytro-ndp:
New login command output:
```
[vagrant@codenvy ~]$ codenvy login --help
DESCRIPTION
        codenvy:login

	Login to Codenvy

SYNTAX
        codenvy:login [options] [username] [password] 

ARGUMENTS
        username
                The username
        password
                The user's password

OPTIONS
        --url
                Codenvy on-prem URL
        --help
                Display this help message

```

Success login to local Codenvy on-prem:
```
[vagrant@codenvy codenvy]$ codenvy login admin password
[CODENVY] Login success on 'http://codenvy.onprem'.
```

Login to local Codenvy on-prem failed:
```
[vagrant@codenvy ~]$ codenvy login admin password
Unable to authenticate for the given credentials on URL 'http://codenvy.onprem'. Check the username and password.
[CODENVY] Login failed on 'http://codenvy.onprem'.
```

Success login to remote Codenvy on-prem by URL:
```
[vagrant@codenvy codenvy]$ codenvy login --url="https://codenvy.com" user password
[CODENVY] Login success on 'https://nightly4.codenvy-stg.com'.
```

PR #740, opened 2016-09-14
Make login to Codenvy on-prem instance by default from IM CLI

dmytro-ndp:
@tolusha: please, review this request for issue https://github.com/codenvy/codenvy/issues/667


cdec/installation-manager-tests/src/main/resources/bin/codenvy/install/test-install-single-node-and-change-config.sh
```diff
@@ -54,5 +54,15 @@ validateExpectedString ".*\"artifact\".\:.\"codenvy\".*\"status\".\:.\"SUCCESS\"
 
 validateInstalledCodenvyVersion ${LATEST_CODENVY_VERSION}
 
+# login to Codenvy on-prem by using IM CLI login command
+executeIMCommand "--valid-exit-code=1" "login" "admin" "wrong"
+validateExpectedString ".*Login.failed.on.'http://${HOST_URL}'.*"
+
+executeIMCommand "login" "admin" "password"
+validateExpectedString ".*Login.success.on.'http://codenvy.onprem'.*"
```
tolusha: Should be here ${HOST_URL} ?

cdec/installation-manager-cli-bundle/src/main/java/com/codenvy/im/cli/preferences/CodenvyOnpremPreferences.java
```diff
@@ -0,0 +1,101 @@
...
+
+    public CodenvyOnpremPreferences(Preferences globalPreferences, MultiRemoteCodenvy storage) {
+        this.globalPreferences = globalPreferences;
+        this.storage = storage;
+    }
+
+    /**
+     * Ensure Codenvy on-prem preferences has url = codenvyOnpremUrl.
+     **/
+    public void upsertUrl(String codenvyOnpremUrl) {
```
tolusha: Do you mean `insertUrl` ?

cdec/installation-manager-cli-bundle/src/main/java/com/codenvy/im/cli/command/LoginCommand.java
```diff
@@ -32,46 +33,38 @@
     @Argument(name = "password", description = "The user's password", required = false, multiValued = false, index = 1)
     private String password;
 
-    @Option(name = "--remote", description = "Name of the remote codenvy", required = false)
-    private String remoteName;
+    @Option(name = "--remote", description = "Codenvy on-prem URL", required = false)
```
tolusha: Since it isn't a remote name anymore, I would prefer to change option name to `codenvyUrl`

cdec/installation-manager-tests/src/main/resources/bin/codenvy/install/test-install-single-node-and-change-config.sh
dmytro-ndp: yes, it should. Thank you.

cdec/installation-manager-cli-bundle/src/main/java/com/codenvy/im/cli/preferences/CodenvyOnpremPreferences.java
dmytro-ndp: I mean upsert, that is:
- insert if url doesn't exist
- update if url does exist

cdec/installation-manager-cli-bundle/src/main/java/com/codenvy/im/cli/command/LoginCommand.java
dmytro-ndp: You are right - **remote** means a little bit different thing in term of  Codenvy CLI.
But **codenvyUrl** is too complicated as for me. IMHO, just **url** could be enough.

tolusha: OK

tolusha:
+1

PR #745, opened 2016-09-15
Add users/profiles synchronizer

evoevodin:
### What does this PR do?
Adds Users/Profiles synchronizer from ldap to database

### Tests written?
Yes

### Docs requirements?
It'd be nice to provide a documentation for ldap synchronization parameters
and authentication types.

@skabashnyuk, @mshaposhnik wdyt?

### DIFF
--- a/wsmaster/codenvy-ldap/src/main/java/com/codenvy/ldap/sync/ProfileMapper.java
+++ b/wsmaster/codenvy-ldap/src/main/java/com/codenvy/ldap/sync/ProfileMapper.java
@@ -1,0 +1,61 @@
+/*
+ *  [2012] - [2016] Codenvy, S.A.
+ *  All Rights Reserved.
+ *
+ * NOTICE:  All information contained herein is, and remains
+ * the property of Codenvy S.A. and its suppliers,
+ * if any.  The intellectual and technical concepts contained
+ * herein are proprietary to Codenvy S.A.
+ * and its suppliers and may be covered by U.S. and Foreign Patents,
+ * patents in process, and are protected by trade secret or copyright law.
+ * Dissemination of this information or reproduction of this material
+ * is strictly forbidden unless prior written permission is obtained
+ * from Codenvy S.A..
+ */
+package com.codenvy.ldap.sync;
+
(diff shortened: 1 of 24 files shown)


wsmaster/codenvy-ldap/src/main/java/com/codenvy/ldap/sync/ProfileMapper.java
```diff
@@ -0,0 +1,61 @@
...
+        profile.setUserId(entry.getAttribute(idAttr).getStringValue());
+        for (Map.Entry<String, String> attrMapping : appToLdapAttrNames.entrySet()) {
+            final LdapAttribute ldapAttr = entry.getAttribute(attrMapping.getValue());
+            if (ldapAttr != null) {
+                profile.getAttributes().put(attrMapping.getKey(), ldapAttr.getStringValue());
+            }
+        }
+        return profile;
+    }
+}
```
skabashnyuk: new line

wsmaster/codenvy-ldap/src/main/java/com/codenvy/ldap/sync/ProfileMapper.java
```diff
@@ -0,0 +1,61 @@
...
+
+import static java.util.stream.Collectors.toMap;
+
+/**
+ * @author Yevhenii Voevodin
+ */
+public class ProfileMapper implements Function<LdapEntry, ProfileImpl> {
+
+    /** App attribute name -> ldap attribute name . */
+    private final ImmutableMap<String, String> appToLdapAttrNames;
```
skabashnyuk: is there any reason to define this field as ```ImmutableMap``` instead of ```Map```

evoevodin: The map won't be mutable in runtime which is clear from the context, so i think you question is related rather to the declaration, my answer is that it is more obvious that map is immutable and less error prone for continuous development, while `Map` interface doesn't say anything about mutability, so as for me it is additional important information for developer

PR #754, opened 2016-09-15
CODENVY-734: Fix adding node behind the proxy for default host name

AndrienkoAleksandr:
### What does this PR do?
Fix adding node behind the proxy for default host name

### Tests written?
No

### Docs requirements?
1.  API changes  
2.  User docs changes  

Signed-off-by: Aleksandr Andrienko <aandrienko@codenvy.com>

### DIFF
--- a/cdec/installation-manager-core/src/main/java/com/codenvy/im/utils/HttpTransport.java
+++ b/cdec/installation-manager-core/src/main/java/com/codenvy/im/utils/HttpTransport.java
@@ -14,6 +14,8 @@
  */
 package com.codenvy.im.utils;
 
+import com.google.api.client.repackaged.com.google.common.annotations.VisibleForTesting;
+
 import org.eclipse.che.commons.annotation.Nullable;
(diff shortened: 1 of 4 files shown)


cdec/installation-manager-core/src/main/java/com/codenvy/im/utils/HttpTransport.java
```diff
@@ -148,15 +150,29 @@ public Path download(String path, Path destinationDir) throws IOException {
         return download(path, "GET", MediaType.APPLICATION_OCTET_STREAM, destinationDir);
     }
 
+    public HttpTransport noProxy() {
+        noProxy = true;
```
sleshchenko: Looks like there can be issues with concurent access. We can avoid this by adding parameter no proxy or by making HttpTransport as no singleton and creating httpTransportFactory

AndrienkoAleksandr: Yes, I'll fix it.

tolusha:
ok

PR #772, opened 2016-09-19
CHE-2366: Fix volumes_from usage in compose environment

garagatyi:
### What does this PR do?
Adapt Codenvy to changes eclipse/che#2487

Related to eclipse/che#2487

### Tests written?
No

### Docs requirements?
1.  ~~API changes~~
2.  ~~User docs changes~~

### DIFF
--- a/plugins/plugin-hosted/codenvy-machine-hosted/src/test/java/com/codenvy/machine/HostedMachineProviderImplTest.java
+++ b/plugins/plugin-hosted/codenvy-machine-hosted/src/test/java/com/codenvy/machine/HostedMachineProviderImplTest.java
@@ -17,7 +17,7 @@ package com.codenvy.machine;
 import com.codenvy.machine.authentication.server.MachineTokenRegistry;
 
 import org.eclipse.che.api.core.util.LineConsumer;
-import org.eclipse.che.api.environment.server.compose.model.ComposeServiceImpl;
+import org.eclipse.che.api.environment.server.model.CheServiceImpl;
 import org.eclipse.che.api.machine.server.util.RecipeRetriever;
 import org.eclipse.che.commons.env.EnvironmentContext;
 import org.eclipse.che.commons.subject.SubjectImpl;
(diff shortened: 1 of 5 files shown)


garagatyi:
@skabashnyuk @evoevodin Please review

plugins/plugin-hosted/codenvy-machine-hosted/src/test/java/com/codenvy/machine/HostedMachineProviderImplTest.java
```diff
@@ -114,27 +113,26 @@ public void setUp() throws Exception {
...
-                                                        PROJECT_FOLDER_PATH,
-                                                        false,
-                                                        false,
-                                                        emptySet(),
-                                                        emptySet(),
-                                                        SNAPSHOT_USE_REGISTRY,
-                                                        MEMORY_SWAP_MULTIPLIER,
-                                                        machineTokenRegistry,
-                                                        emptySet());
+        provider = new HostedMachineProviderImpl(dockerConnector,
```
evoevodin: wdyt about creating a builder class for `HostedComposeMachineProviderImpl` for test purpose and place it near `HostedMachineProviderImplTest` which may be helpful and more clear, as it may used default values for more than a half of parameters

wsmaster/codenvy-hosted-platform-api-impl/src/test/java/com/codenvy/api/workspace/TestObjects.java
```diff
@@ -64,7 +64,8 @@ public static WorkspaceImpl createWorkspace(String owner, String devMachineRam,
         Map<String, ExtendedMachineImpl> machines = new HashMap<>();
         machines.put("dev-machine", new ExtendedMachineImpl(singletonList("org.eclipse.che.ws-agent"),
                                                             emptyMap(),
-                                                            new HashMap<>(singletonMap("memoryLimitBytes", Long.toString(Size.parseSize(devMachineRam))))));
+                                                            new HashMap<>(singletonMap("memoryLimitBytes",
```
evoevodin: Does the map have to be mutable at this point?
If so, than `ExtendedMachineImpl` constructor should be responsible for that, if not
you can simply pass the instance of `singletonMap(...)`

plugins/plugin-hosted/codenvy-machine-hosted/src/test/java/com/codenvy/machine/HostedMachineProviderImplTest.java
garagatyi: I think that this class doesn't fit builder API

wsmaster/codenvy-hosted-platform-api-impl/src/test/java/com/codenvy/api/workspace/TestObjects.java
garagatyi: Constructor does copy map. I didn't notice. I'll fix that, thanks.

Issue #456, opened 2016-07-26
Puppet config isn't updated properly in time of changing Codenvy hostname when there are machine nodes


dmytro-ndp:
Puppet config becomes misconfigured after the changing Codenvy onprem host name:
1) _/etc/puppet/puppet.conf_ on machine node contained absolute value of **certname** property
Fixup: https://github.com/codenvy/codenvy/blob/master/cdec/installation-manager-core/src/main/java/com/codenvy/im/managers/helper/NodeManagerHelperCodenvy4Impl.java#L245
```
nodeCommands.add(createCommand(format("sudo sed -i 's/certname = %1$s/certname = %2$s/g' /etc/puppet/puppet.conf", node.getHost(), node.getHost().replace(oldHostName, newHostName)), node));
```

2) _/etc/puppet/autosign.conf_ on master wasn't updated with new machine node hostnames.

3) *$swarm_node* variable in _/etc/puppet/manifests/nodes/codenvy/codenvy.pp_ on master wasn't updated with new machine node hostnames.

4) we need to check if existed nodes with new hostname are available before changing hostname.

Codenvy on-prem version: <= 4.6.0-SNAPSHOT
Problem can be reliably reproduced: Yes

bmicklea:
@dmytro-ndp is there a workaround for this?

dmytro-ndp:
yes, I described fixup in description of the issue.

PR #805, opened 2016-09-21
(#456) fix update of hostname of Codenvy on-prem with machine nodes

dmytro-ndp:
List of points:
1) fix absolute value of **certname** property in _/etc/puppet/puppet.conf_ on machine node;
2) update _/etc/puppet/autosign.conf_ with new machine node hostnames on master;
3) update *$swarm_node* variable in _/etc/puppet/manifests/nodes/codenvy/codenvy.pp_ on master  with new machine node hostnames;
4) check if existed nodes with new hostname are available before changing hostname.

### DIFF
--- a/cdec/installation-manager-core/src/main/java/com/codenvy/im/commands/CommandLibrary.java
+++ b/cdec/installation-manager-core/src/main/java/com/codenvy/im/commands/CommandLibrary.java
@@ -533,7 +533,7 @@ public class CommandLibrary {
      * @see "http://www.unix.com/302910488-post2.html?s=c351328813d6bc7a509c0a5dee5dc9e7"
      */
     private static String getCheckRemotePortOpenedCommand(String remoteHost, int remotePort) {
-        return format("2>/dev/null >/dev/tcp/%s/%s", remoteHost, remotePort);
+        return format("sleep 3 && echo >/dev/tcp/%s/%s", remoteHost, remotePort);
     }
 
     private static class EmptyCommand implements Command {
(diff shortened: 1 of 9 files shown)


cdec/installation-manager-core/src/main/java/com/codenvy/im/managers/helper/NodeManagerHelper.java
```diff
@@ -91,6 +95,32 @@ public Command getValidatePuppetMasterAccessibilityCommand(String puppetMasterNo
         return CommandLibrary.createCheckRemotePortOpenedCommand(puppetMasterNodeDns, 8140, node);
     }
```
skabashnyuk: javadoc

cdec/installation-manager-core/src/main/java/com/codenvy/im/commands/CommandLibrary.java
```diff
@@ -533,7 +533,7 @@ public static Command createCheckRemotePortOpenedCommand(String remoteHost, int
      * @see "http://www.unix.com/302910488-post2.html?s=c351328813d6bc7a509c0a5dee5dc9e7"
      */
     private static String getCheckRemotePortOpenedCommand(String remoteHost, int remotePort) {
-        return format("2>/dev/null >/dev/tcp/%s/%s", remoteHost, remotePort);
+        return format("sleep 3 && echo >/dev/tcp/%s/%s", remoteHost, remotePort);
```
tolusha: Why do we need sleep command ?

tolusha:
+1

cdec/installation-manager-core/src/main/java/com/codenvy/im/commands/CommandLibrary.java
dmytro-ndp: because of without sleep it fails

Issue #502, opened 2016-08-02
Create 'resource reached error' table in dasboard


vinokurig:
![4a7989b0-4dc7-11e6-8604-8bf3f8f50dfb](https://cloud.githubusercontent.com/assets/7668752/17319413/1ec5a964-5896-11e6-8d89-141ce3004557.png)
We need to have a 'resource reached error' table in dasboard to notify user that resources limit reached

Before starting workspace if exception with special error code is received this table should be showed in dasboard. Also if a message is received by websocket it should be showed. When a cancel message is received by websocket this table should be removed.

**Depends on :** https://github.com/codenvy/codenvy/issues/408

PR #809, opened 2016-09-21
CODENVY-502: check system ram limit channel to handle app notifications

ashumilova:
### What does this PR do?

Displays system ram limit notifications.

https://github.com/codenvy/codenvy/issues/502

PR depends on https://github.com/eclipse/che/pull/2529

Signed-off-by: Ann Shumilova <ashumilova@codenvy.com>

### DIFF
--- a/dashboard/src/components/api/codenvy-system.factory.js
+++ b/dashboard/src/components/api/codenvy-system.factory.js
@@ -1,0 +1,78 @@
+/*
+ *  [2015] - [2016] Codenvy, S.A.
+ *  All Rights Reserved.
+ *
+ * NOTICE:  All information contained herein is, and remains
+ * the property of Codenvy S.A. and its suppliers,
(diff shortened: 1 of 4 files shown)


dashboard/src/components/api/codenvy-system.factory.js
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  /**
+   * Gets system RAM limit info.
+   */
+  getSystemRAMLimit() {
+    let promise = this.systemAPI.getLimit().$promise;
+    promise.then((info) => {
+      if (info.systemRamLimitExceeded) {
+        this.notification = this.applicationNotifications.addErrorNotification('Low RAM', 'The system is low on RAM. New workspaces will not start until more RAM is available.');
```
olexii4: We could use one string constants instead. WDYT?

PR #817, opened 2016-09-21
CHE-2369: add a mechanism which checks state of the ws agent

svor:
### What does this PR do?
Adds a service which periodically pings ws agent and notify a user when something is happened.

https://github.com/eclipse/che/issues/2369

### Previous Behavior
When ws agent stops responding a user don't get any response about this and don't know what is happened.

### Tests written?
Yes

@vparfonov please take a look


assembly/onpremises-ide-packaging-war-platform-api/src/main/java/com/codenvy/api/deploy/OnPremisesIdeApiModule.java
```diff
@@ -277,7 +283,7 @@ protected void configure() {
                         ),
                         new UriStartFromRequestFilter("/api/license/legality")
                 )
-        );
+                                            );
```
benoitf: looks strange ? formatter issue ?

svor: yes! fixed, thx!

PR #830, opened 2016-09-22
Use also TypeScript for codenvy dashboard

benoitf:
### New Behavior
Switch from ES6 to TypeScript (like in Che dashboard)

### Tests written?
Yes

### Docs requirements?
1.  API changes  
2.  User docs changes  

 - fix some .js referencing direct .js files
 - fix package.json command that was only launching 'bower install' on che-dashboard (so che module was not built if dependencies were not the same than codenvy dashboard dependencies)

Change-Id: I30df70e31f3f631ff917a5f833b7f55c09247585


riuvshin:
+1

Issue #741, opened 2016-09-14
Use correct groupId for site artifact


garagatyi:
Now site artifact uses incorrect maven group id **com.codenvy**. Replace it with **com.codenvy.onpremises**

PR #838, opened 2016-09-23
CODENVY-741 rename group id for site

vkuznyetsov:
Rename group ID for site
https://github.com/codenvy/codenvy/issues/741

### DIFF
--- a/wsmaster/codenvy-hosted-integration-tests/codenvy-integration-tck-jpa/pom.xml
+++ b/wsmaster/codenvy-hosted-integration-tests/codenvy-integration-tck-jpa/pom.xml
@@ -15,8 +15,7 @@
     from Codenvy S.A..
 
 -->
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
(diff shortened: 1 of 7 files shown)


vkuznyetsov:
ci-build

wsmaster/codenvy-hosted-integration-tests/codenvy-integration-tck-jpa/pom.xml
```diff
@@ -193,8 +192,7 @@
                                         che-core-api-user,
                                         che-core-api-machine,
                                         che-core-api-workspace,
-                                        che-core-api-ssh
-                                    </includeArtifactIds>
```
garagatyi: Please fix formatting

vkuznyetsov: I run mvn sortpom:sort command. It formatted a pom file on its own. Do you think we need to format code each time we run the sortpom?

garagatyi: Then its ok

Issue #582, opened 2016-08-15
Allow per user overrides for all resource limit configs


bmicklea:
Most enterprises set the workspace and RAM limitations for their organization. However, in all enterprises there are certain developers who need exceptional resources for their work. Codenvy should allow the admin to override the limit settings on a per-user basis.

Ultimately this should be done through the UI, but in the meantime allowing this to be executed through an enhancement to the Permissions API will allow enterprises to use the feature sooner. The API would have to support:
- Adding overrides to a user
- Listing overrides per install (list of all users with overrides and what those overrides are) - no input parameter needed
- Listing overrides per user (list of the overrides for a particular user) - would require the input of the user's email

This may be part of the "organization" work that is going on. Ideally the system would let you identify a user in the database and provide them with overrides of one or more workspace and RAM limitations.

gazarenkov:
Well, by our model:
- Organizations have hierarchy
- User which is a member of top Organization MAY have permission to create sub-organization and to set resource limits (assign licence)  - that's what may be called Admin permissions in this sense

So conceptually "Admin" may create sub-organization of one user and assign dedicated licence (resources) to it (actually to this user)

bmicklea:
That's fine - for usability we may implement an in-client default that grants admin privileges for a sub-organization to everyone who has admin privileges to the parent organization.

TylerJewell:
It is a clever way to handle it.  The next step is for sergey to propose a CLI syntax.  Let's look and agree on it.

tolusha:
@TylerJewell We don't need @sleshchenko to design CLI syntax. We just need a real user cases.

dmytro-ndp:
It seems we need have this issue be done before https://github.com/codenvy/codenvy/issues/409

tolusha:
It is not actual anymore concerning CLI part.

bmicklea:
Correct - but can we see an outline of the planned API methods to handle this?

PR #877, opened 2016-09-30
CODENVY-582 Add ability to define free resources limit

sleshchenko:
### What does this PR do?
Adds ability to define free resources limit by user that has "manageCodenvy" permission

https://github.com/codenvy/codenvy/issues/582

### Previous Behavior
Accounts were able to use resources which are defined by external configuration

### New Behavior
By default accounts are able to use resources which are defined by external configuration. System admin can override limit for each account separately.

### Tests written?
Yes


sleshchenko:
@skabashnyuk @akorneta @evoevodin @mshaposhnik Please review my changes

wsmaster/codenvy-hosted-api-resource/src/main/java/com/codenvy/resource/api/free/FreeResourcesLimitService.java
```diff
@@ -0,0 +1,142 @@
...
+    public FreeResourcesLimitService(FreeResourcesManager freeResourcesManager) {
+        this.freeResourcesManager = freeResourcesManager;
+    }
+
+    @POST
+    @Consumes(APPLICATION_JSON)
+    @Produces(APPLICATION_JSON)
+    @ApiOperation(value = "Store free resources limit",
+                  response = FreeResourcesLimitDto.class)
+    @ApiResponses({@ApiResponse(code = 201, message = "The resources limit successfully stored"),
```
mshaposhnik: Maybe 204?

wsmaster/codenvy-hosted-api-resource/src/main/java/com/codenvy/resource/api/free/FreeResourcesProvider.java
```diff
@@ -29,29 +32,37 @@
...
-    private final long                ramPerOrganization;
+    private final FreeResourcesManager freeResourcesManager;
+    private final AccountManager       accountManager;
+    private final OrganizationManager  organizationManager;
+    private final long                 ramPerUser;
+    private final long                 ramPerOrganization;
 
     @Inject
-    public FreeResourcesProvider(AccountManager accountManager,
+    public FreeResourcesProvider(FreeResourcesManager freeResourcesManager,
```
mshaposhnik: Singleton...

wsmaster/codenvy-hosted-api-resource/src/main/java/com/codenvy/resource/spi/jpa/JpaFreeResourcesLimitDao.java
```diff
@@ -0,0 +1,122 @@
...
+        EntityManager manager = managerProvider.get();
+        try {
+            final FreeResourcesLimitImpl existedLimit = get(resourcesLimit.getAccountId());
+            existedLimit.setResources(resourcesLimit.getResources());
+        } catch (NotFoundException n) {
+            manager.persist(resourcesLimit);
+        }
+    }
+
+    @Transactional
```
akorneta: It seems that @Transactional in this case useless because the method with private access modifiers it won't be executed in the transaction.

wsmaster/codenvy-hosted-api-resource/src/main/java/com/codenvy/resource/api/free/FreeResourcesProvider.java
sleshchenko: Added

wsmaster/codenvy-hosted-api-resource/src/main/java/com/codenvy/resource/spi/jpa/JpaFreeResourcesLimitDao.java
sleshchenko: Removed

wsmaster/codenvy-hosted-api-resource/src/main/java/com/codenvy/resource/api/free/FreeResourcesLimitServicePermissionsFilter.java
```diff
@@ -0,0 +1,60 @@
...
+        switch (genericMethodResource.getMethod().getName()) {
+            case STORE_FREE_RESOURCES_LIMIT_METHOD:
+            case GET_FREE_RESOURCES_LIMITS_METHOD:
+            case GET_FREE_RESOURCES_LIMIT_METHOD:
+            case REMOVE_FREE_RESOURCES_LIMIT_METHOD:
+                final Subject currentSubject = EnvironmentContext.getCurrent().getSubject();
+                if (currentSubject.hasPermission(SystemDomain.DOMAIN_ID, null, SystemDomain.MANAGE_CODENVY_ACTION)) {
+                    return;
+                }
+            default:
```
evoevodin: Document [fall through](https://github.com/codenvy/codenvy-dev-resources/wiki/Formatting#switch-statement)
Document [fall through](https://github.com/codenvy/codenvy-dev-resources/wiki/Formatting#switch-statement)

wsmaster/codenvy-hosted-api-resource/src/main/java/com/codenvy/resource/api/free/FreeResourcesManager.java
```diff
@@ -0,0 +1,128 @@
...
+     */
+    public FreeResourcesLimit get(String accountId) throws NotFoundException, ServerException {
+        requireNonNull(accountId, "Required non-null account id");
+        return freeResourcesLimitDao.get(accountId);
+    }
+
+    /**
+     * Removes free resources limit for account with specified id
+     *
+     * Note: after removing resources limit account will be able to use default resources
```
evoevodin: Consider adding `<p>` before the Note, generated javadoc will be like this `specified idNote:`

wsmaster/codenvy-hosted-api-resource/src/main/java/com/codenvy/resource/api/free/FreeResourcesManager.java
```diff
@@ -0,0 +1,128 @@
...
+     *
+     * @param accountId
+     *         account id to remove resources limit
+     * @throws NullPointerException
+     *         when {@code accountId} is null
+     * @throws ServerException
+     *         when any other error occurs
+     */
+    void remove(String accountId) throws ServerException {
```
evoevodin: The common question here is what if i pass id of non-existing account, it seems that this method does nothing in this case(which i think is really good) but the things like that should be documented by method documentation.
BTW does it have to be package-private?

wsmaster/codenvy-hosted-api-resource/src/main/java/com/codenvy/resource/api/free/FreeResourcesProvider.java
```diff
@@ -60,28 +77,41 @@ public FreeResourcesProvider(AccountManager accountManager,
...
+        try {
+            FreeResourcesLimit resourcesLimit = freeResourcesManager.get(accountId);
+            for (Resource resource : resourcesLimit.getResources()) {
+                freeResources.put(resource.getType(), resourceInstanceCreator.newInstance(resource));
+            }
+            limitId = resourcesLimit.getAccountId();
+        } catch (NotFoundException ignored) {
+        }
+
+        getDefaultResources(accountId).stream()
```
evoevodin: IMHO
```java
for (AbstractResource resource : getDefaultResources(accountId)) {
    freeResources.putIfAbsent(resource.getType(), resource);
}
```
is better

wsmaster/codenvy-hosted-api-resource/src/main/java/com/codenvy/resource/api/free/FreeResourcesManager.java
sleshchenko: :+1: Added comment to java docs about case with non-existing account and added public access modifier

wsmaster/codenvy-hosted-api-resource/src/main/java/com/codenvy/resource/api/free/FreeResourcesProvider.java
sleshchenko: The same to me. Fixed

wsmaster/codenvy-hosted-api-resource/src/main/java/com/codenvy/resource/api/free/FreeResourcesLimitServicePermissionsFilter.java
sleshchenko: Documented

wsmaster/codenvy-hosted-api-resource/src/main/java/com/codenvy/resource/spi/impl/FreeResourcesLimitImpl.java
```diff
@@ -0,0 +1,118 @@
...
+/**
+ * Data object for {@link FreeResourcesLimit}.
+ *
+ * @author Sergii Leschenko
+ */
+@Entity(name = "FreeResourcesLimit")
+@NamedQueries(
+        {
+                @NamedQuery(name = "FreeResourcesLimit.get",
+                            query = "SELECT l FROM FreeResourcesLimit l WHERE l.accountId= :accountId"),
```
evoevodin: `SELECT l` may be confused with `SELECT 1`  l, can you please rename it to `lim`, `limit` or something like that.

PR #882, opened 2016-09-30
More accurate ldap configuration properties validation

mshaposhnik:
### What does this PR do?
Prevents some NPE's or unexpected behavior when not all required LDAP properties are set by
throwing appropriate exceptions in those cases.

### DIFF
--- a/wsmaster/codenvy-ldap/src/main/java/com/codenvy/ldap/auth/AuthenticatorProvider.java
+++ b/wsmaster/codenvy-ldap/src/main/java/com/codenvy/ldap/auth/AuthenticatorProvider.java
@@ -225,6 +225,4 @@ public class AuthenticatorProvider implements Provider<Authenticator> {
             }
         }
     }
-
-
 }
--- a/wsmaster/codenvy-ldap/src/main/java/com/codenvy/ldap/auth/AuthenticatorProvider.java
+++ b/wsmaster/codenvy-ldap/src/main/java/com/codenvy/ldap/auth/AuthenticatorProvider.java
@@ -14,9 +14,11 @@
  */
 package com.codenvy.ldap.auth;
 
-import com.google.common.base.Strings;
+import com.google.common.annotations.VisibleForTesting;
 
 import org.eclipse.che.commons.annotation.Nullable;
+import org.eclipse.che.commons.lang.Pair;
+import org.eclipse.che.inject.ConfigurationException;
 import org.ldaptive.auth.Authenticator;
 import org.ldaptive.auth.EntryResolver;
 import org.ldaptive.auth.FormatDnResolver;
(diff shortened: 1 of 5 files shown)


wsmaster/codenvy-ldap/src/main/java/com/codenvy/ldap/sync/UserMapper.java
```diff
@@ -40,8 +41,19 @@ public UserMapper(String userIdAttr, String userNameAttr, String userEmailAttr)
...
     public UserImpl apply(LdapEntry entry) {
+        validateRequiredAttrsExists(entry);
         return new UserImpl(entry.getAttribute(idAttr).getStringValue(),
                             entry.getAttribute(mailAttr).getStringValue(),
                             entry.getAttribute(nameAttr).getStringValue());
     }
+
+    private void validateRequiredAttrsExists(LdapEntry entry) {
+        try {
+            requireNonNull(entry.getAttribute(idAttr), "Cannot identify required ID attribute in LDAP entry. Please, check configuration parameter correctness.");
```
skabashnyuk: add parameter name to the message

wsmaster/codenvy-ldap/src/main/java/com/codenvy/ldap/auth/AuthenticatorProvider.java
```diff
@@ -180,7 +205,24 @@ private PooledBindAuthenticationHandler getPooledBindAuthenticationHandler(Poole
...
         final PooledCompareAuthenticationHandler handler = new PooledCompareAuthenticationHandler(
                 connFactory);
+        checkRequiredPropertySet(USER_PASSWORD_ATTRIBUTE_PROPERTY_NAME, userPasswordAttribute);
         handler.setPasswordAttribute(userPasswordAttribute);
         return handler;
     }
+
+    @SafeVarargs
+    @VisibleForTesting
+    final void checkRequiredPropertySet(String name, String value, Pair<String, String>... nameValuePairs) {
```
skabashnyuk: may be two methods?

wsmaster/codenvy-ldap/src/main/java/com/codenvy/ldap/sync/ProfileMapper.java
```diff
@@ -44,6 +44,7 @@ public ProfileMapper(String idAttr, Pair<String, String>[] attributes) {
             this.appToLdapAttrNames = ImmutableMap.of();
         } else {
             this.appToLdapAttrNames = ImmutableMap.copyOf(Arrays.stream(attributes)
+                                                                .filter(p -> !p.first.isEmpty() && p.second != null)
```
evoevodin: AFAIK you can't get NULL injected where array of string pairs is injected, i think this check 
is useless and shouldn't be present here.

evoevodin:
@mshaposhnik please add yourself as author to the classes you changed

PR #894, opened 2016-10-04
Fix ldap migration

mshaposhnik:
### What does this PR do?
Fixes NPE on profile if ID attribute is wrong


wsmaster/codenvy-ldap/src/main/java/com/codenvy/ldap/sync/LdapSynchronizer.java
```diff
@@ -312,18 +314,22 @@ public void shutdownScheduler() {
         }
     }
 
-    private void validateRequiredFields(UserImpl user) {
+    private boolean validateRequiredFields(UserImpl user) {
         if (user.getId() == null) {
```
skabashnyuk: isValid()?

mshaposhnik: isValidUser()?

mshaposhnik: Ok

Issue #792, opened 2016-09-20
Implement workspace cap that prevents users from running more than X workspaces


bmicklea:
As an admin I need a way to set a limit on the number of workspaces a user can run simultaneously. This limit should be called: `limits.user.workspaces.run.count`

This would prevent situations where developers chew up too many resources on simultaneously running workspaces.

Default value: 3

vinokurig:
@bmicklea What default value should be set for this property? How about maximum 10 started workspaces per user?

bmicklea:
@vinokurig default should be 3

PR #969, opened 2016-10-17
CODENVY-792: Add workspace cap limit

vinokurig:
### What does this PR do?
Add started workspaces number per user limit

https://github.com/codenvy/codenvy/issues/792

### Tests written?
Yes

@sleshchenko @evoevodin please review


wsmaster/codenvy-hosted-platform-api-impl/src/test/java/com/codenvy/api/workspace/LimitsCheckingWorkspaceManagerTest.java
```diff
@@ -454,11 +469,88 @@ public void shouldPermitToCheckRamOnlyForFiveThreadsAtTheSameTime() throws Excep
...
+                                                                                              null,
+                                                                                              null,
+                                                                                              null,
+                                                                                              null,
+                                                                                              null,
+                                                                                              environmentParser,
+                                                                                              false,
+                                                                                              false,
+                                                                                              2000));
+        doReturn(emptyList()).when(manager).getByNamespace(anyString()); // <- currently started 0 workspaces
```
sleshchenko: It means that there are no any workspaces in given namespace (running or stopped). It would better to add test that will test counting all except stopped workspaces

vinokurig: Improved existing tests

wsmaster/codenvy-hosted-platform-api-impl/src/main/java/com/codenvy/api/workspace/LimitsCheckingWorkspaceManager.java
```diff
@@ -186,17 +190,47 @@ public WorkspaceImpl updateWorkspace(String id, Workspace update) throws Conflic
...
+    <T extends WorkspaceImpl> T checkStartedWorkspacesNumberAndPropagateStart(String nameSpace, WorkspaceCallback<T> callback)
+            throws ServerException, NotFoundException, ConflictException {
+        if (startedWorkspacesLimit < 0) {
+            return callback.call();
+        }
+
+        // It is important to lock in this place because:
+        // if started workspaces number limit is 10 and user has 9 started workspaces, then if he sends 2 separate requests
+        // to start a workspace, it may start both of them, because started workspaces number check is not atomic one.
+        final Lock lock = CREATE_LOCKS.get(nameSpace);
```
vinokurig: This lock can be used as multiple lock for each namespace and it is used in similar `checkCountAndPropagateCreation()` check.

sleshchenko: usage in `checkCountAndPropagateCreation()` is clear. But you should not use CREATE_LOCK here. You should create new one and use it. Or rename existing one to something common. I prefer creating new one lock because it won't lock workspaces creating on workspaces starting and vice versa

vinokurig: Thanks for note, fixed

wsmaster/codenvy-hosted-platform-api-impl/src/test/java/com/codenvy/api/workspace/LimitsCheckingWorkspaceManagerTest.java
```diff
@@ -454,11 +471,104 @@ public void shouldPermitToCheckRamOnlyForFiveThreadsAtTheSameTime() throws Excep
...
         //Execute paused threads to release the throughput limit for other threads.
         invokeProcessLatch.countDown();
         //Wait for throughput limit will be released and check that RAM check was performed in other threads.
-        verify(manager, timeout(300).times(7)).checkSystemRamAndPropagateStart(anyObject());
+        verify(manager, timeout(300).times(7)).checkLimitsAndPropagateStart(anyString(), anyObject());
+    }
+
+    @Test
+    public void shouldCallCreateCallBackIfStartedWorkspacesNumberLimitIsNotExceeded() throws Exception {
+        final LimitsCheckingWorkspaceManager manager = spy(new LimitsCheckingWorkspaceManager(2,
```
evoevodin: There are so much parameters, this tests class grows really fast.
What do  you think about creating a builder class for testing like `ManagerBuilder` with a method like `createDefault()` so it would be possible to change only parameters which are tested, eg:

```java
ManagerBuilder.createDefault()
              .setRunningLimit(2)
              .build();
```

This may simplify tests a bit.

wsmaster/codenvy-hosted-platform-api-impl/src/test/java/com/codenvy/api/workspace/LimitsCheckingWorkspaceManagerTest.java
```diff
@@ -454,11 +471,104 @@ public void shouldPermitToCheckRamOnlyForFiveThreadsAtTheSameTime() throws Excep
...
+                                                                                              2000));
+        final WorkspaceCallback callback = mock(WorkspaceCallback.class);
+        WorkspaceImpl startingWorkspace = mock(WorkspaceImpl.class);
+        WorkspaceImpl runningWorkspace = mock(WorkspaceImpl.class);
+        WorkspaceImpl stoppingWorkspace = mock(WorkspaceImpl.class);
+        WorkspaceImpl stoppedWorkspace = mock(WorkspaceImpl.class);
+        when(startingWorkspace.getStatus()).thenReturn(WorkspaceStatus.STARTING);
+        when(runningWorkspace.getStatus()).thenReturn(WorkspaceStatus.RUNNING);
+        when(stoppingWorkspace.getStatus()).thenReturn(WorkspaceStatus.STOPPING);
+        when(stoppedWorkspace.getStatus()).thenReturn(WorkspaceStatus.STOPPED);
```
evoevodin: There is a new status called `SNAPSHOTTING`, i think you should rewrite the test to be more generic, use `EnumSet` of `WorkspaceStatus` values without RUNNING status, in this way you won't need to change this test when workspace statuses are changed

PR #980, opened 2016-10-18
Use preRemove event instead of cascade remove for automatically removing of suborganizations

sleshchenko:
### What does this PR do?
Reworks automatically removing of suborganizations by using preRemove event instead of cascade removing. It is needed because cascade remove doesn't throw pre remove event for suborganization in this case. As result we can't remove root organization because members of suborganization will not be removed.

### Tests written?
Yes


sleshchenko:
@evoevodin @skabashnyuk @akorneta Please review my changes

Issue #1009, opened 2016-10-22
Dashboard doesn't display nag message about the illegal usage of Codenvy on-prem


dmytro-ndp:
Dashboard doesn't display nag message at the top of the page in case of **false** value of response on request to http://codenvy.onprem/api/license/legality: `{"value":"false"}`

There is no warning of illegal usage for admin:
![no_warning_of_illegal_usage_for_admin_22-10-2016 png](https://cloud.githubusercontent.com/assets/1197777/19622322/03f9f568-98b0-11e6-83fe-60df08682287.png)

There is no warning of illegal usage for user:
![no_warning_of_illegal_usage_for_user_22-10-2016](https://cloud.githubusercontent.com/assets/1197777/19622320/fa6ddbd6-98af-11e6-8f52-1498e70c3388.png)

Codenvy on-prem version: **5.0.0-M6-SNAPSHOT**

Issue #599, opened 2016-08-18
Move Codenvy License management staff to Codenvy API Server


dmytro-ndp:
## Short Description of Problem
There is a method [com.codenvy.im.managers.LdapManager#getNumberOfUsers](https://github.com/codenvy/codenvy/blob/4.6.2/cdec/installation-manager-core/src/main/java/com/codenvy/im/managers/LdapManager.java#L114-L114) in _Installation Manager_ which is used by [com.codenvy.im.service.LicenseService#isCodenvyUsageLegal](https://github.com/codenvy/codenvy/blob/4.6.2/cdec/installation-manager-server-war/src/main/java/com/codenvy/im/service/LicenseService.java#L172-L172) to verify if Codenvy is being used legally according to Codenvy license.

To avoid requesting to _Codenvy User DB_ directly from _Installation Manager_, and taking in mind that Codenvy API service to get Codenvy on-prem user number can't be public, we ought to move all _Codenvy License management_ staff from the _Installation Manager Server_ to _Codenvy API Server_.

## Implementation Steps
- [x] Move [com.codenvy.im.license](https://github.com/codenvy/codenvy/tree/4.6.2/cdec/installation-manager-core/src/main/java/com/codenvy/im/license) to Codenvy API Server (it uses third-part library [License4J](https://github.com/codenvy/codenvy/blob/4.6.2/cdec/installation-manager-core/pom.xml#L68):
```
        <dependency>
            <groupId>com.license4j</groupId>
            <artifactId>license4j-runtime-library</artifactId>
        </dependency>
```

- [x] Move license-related properties from *installation-manager.properties* file to the Codenvy API server /home/codenvy/codenvy-data/conf/license.properties:
1. [license-manager.public_key](https://github.com/codenvy/codenvy/blob/4.6.2/cdec/installation-manager-core/src/main/resources/codenvy/installation-manager.properties#L27) as it; 
1. [installation-manager.storage_dir](https://github.com/codenvy/codenvy/blob/4.6.2/cdec/installation-manager-core/src/main/resources/codenvy/installation-manager.properties#L22) - to property *license-manager.license-file=/home/codenvy/codenvy-data/license/license*
- [x] Move [com.codenvy.im.service.LicenseService](https://github.com/codenvy/codenvy/blob/4.6.2/cdec/installation-manager-server-war/src/main/java/com/codenvy/im/service/LicenseService.java) to Codenvy API Server:
[...]

PR #1011, opened 2016-10-23
CODENVY-1009: fix license legality check

ashumilova:
### What does this PR do?

Fix license legality check.

https://github.com/codenvy/codenvy/issues/1009

Signed-off-by: Ann Shumilova <ashumilova@codenvy.com>

### DIFF
--- a/dashboard/src/app/onprem/license-nag-message-management/nag-message.controller.ts
+++ b/dashboard/src/app/onprem/license-nag-message-management/nag-message.controller.ts
@@ -52,10 +52,10 @@ export class NagMessageCtrl {
    * @param licenseLegality
    */
   checkLegality(licenseLegality) {
-    if (licenseLegality && licenseLegality.value === 'false') {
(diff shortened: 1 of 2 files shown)


ashumilova:
@riuvshin License legality on nightly5 should be checked, cause now it's false.

riuvshin:
@ashumilova nightly5 is not onprem it is saas we can't check this on saas

ashumilova:
Then saas should not have license service deployed, if it's not used, or it should return correct value, like on beta now. Not sure, but the result of this issue https://github.com/codenvy/codenvy/issues/599 may cause it.

dmytro-ndp:
@riuvshin, @ashumilova: https://codenvy.io has Codenvy license and so it returns legalilty=true.
https://nightly5.codenvy-stg.com doesn't has Codenvy license and, as a result it's legalilty=false.
And it doesn't matter are they kind of _SaaS_ or _on-prem_ now.

PR #1043, opened 2016-10-27
CHE-1787: Remove workspace id from websocket connection path to master

vinokurig:
### What does this PR do?

Removes workspace id from websocket connection path to master.
https://github.com/eclipse/che/issues/1787

### Previous Behavior

Separate web-socket connection was created for each workspase.
### New Behavior

All workspaces are using one web-socket connection.

@vparfonov @akurinnoy @ashumilova Please review

### DIFF
--- a/dashboard/src/app/factories/load-factory/load-factory.controller.ts
+++ b/dashboard/src/app/factories/load-factory/load-factory.controller.ts
@@ -189,7 +189,7 @@ export class LoadFactoryCtrl {
    */
   startWorkspace(workspace) {
     this.workspace = workspace;
-    var bus = this.cheAPI.getWebsocket().getBus(workspace.id);
+    var bus = this.cheAPI.getWebsocket().getBus();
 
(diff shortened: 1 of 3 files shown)


dashboard/src/app/factories/load-factory/load-factory.controller.ts
```diff
@@ -189,7 +189,7 @@ export class LoadFactoryCtrl {
    */
   startWorkspace(workspace) {
     this.workspace = workspace;
-    var bus = this.cheAPI.getWebsocket().getBus(workspace.id);
```
benoitf: This will not work 
WebSocket class is storing connections per workspace id so WebSocket class has to be updated as well

vinokurig: This is done in https://github.com/eclipse/che/pull/2907

benoitf:
requires https://github.com/eclipse/che/pull/2907

PR #1044, opened 2016-10-27
Exdent LdapSynchronizer functionality

evoevodin:
### What does this PR do?
Adds more configuration options to the `LdapSyncrhonizer`.
The options are:
- new attribute for linking users, it can be _id_ or _email_
- new attribute for disabling users update when such users already exist
- new attribute for disabling users removal when users exist in db while missing from ldap storage

Also there are 3 new fileds in synchronization result:
-  up-to-date - users who don't need to be synchronized as they are the same in ldap and db
-  skipped - users who were fetched from ldap storage and were skipped by synchronizer due to configuration options
- processed - the count of the processed users(fetched from ldap).
Moreover now we count not valid users as failed to synchronize.

@skabashnyuk, @sleshchenko, @akorneta please review

### DIFF
--- a/wsmaster/codenvy-ldap/src/main/java/com/codenvy/ldap/sync/DBUserFinder.java
+++ b/wsmaster/codenvy-ldap/src/main/java/com/codenvy/ldap/sync/DBUserFinder.java
@@ -1,0 +1,122 @@
+/*
+ *  [2012] - [2016] Codenvy, S.A.
+ *  All Rights Reserved.
+ *
+ * NOTICE:  All information contained herein is, and remains
+ * the property of Codenvy S.A. and its suppliers,
+ * if any.  The intellectual and technical concepts contained
+ * herein are proprietary to Codenvy S.A.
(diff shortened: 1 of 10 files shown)


skabashnyuk:
@mshaposhnik  wdyt

mshaposhnik:
It would be berfect if you put a few words about new properties into readme.md

wsmaster/codenvy-ldap/src/main/java/com/codenvy/ldap/sync/DBUserFinder.java
```diff
@@ -0,0 +1,124 @@
...
+        }
+
+        @Override
+        @SuppressWarnings("unchecked") // user email is always string
+        public Set<String> findLinkingIds() {
+            return new HashSet<>(dbHelper.executeNativeQuery("SELECT email FROM Usr"));
+        }
+    }
+}
+
```
sleshchenko: Please remove extra line

PR #1130, opened 2016-11-10
Docker version check

mshaposhnik:
Add checking for docker client/server & API version

### DIFF
--- a/codenvy.sh
+++ b/codenvy.sh
@@ -255,6 +255,18 @@ check_docker() {
     error "Docker issues - 'docker ps' fails."
     return 1;
   fi
+
+  DOCKER_VERSION=($(docker version |  grep  "Version:" | sed 's/Version://'))
+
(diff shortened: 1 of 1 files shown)


TylerJewell:
@benoitf @mshaposhnik - requires matching improvement to be made in `cli-refactoring` branch of Eclipse Che.

TylerJewell:
I am not sure we should be enforcing a check on client / daemon versions - given the scenarios with che-launcher and our ability to package up different CLI clients into other docker containers which can interact with many different versions of daemons.

skabashnyuk:
@TylerJewell  it's not clear what to do with this pr. Should we abandon it?

TylerJewell:
I am -1 on requiring that the docker client & server be identical.  I am +1 on requiring the docker server version being >1.11.

mshaposhnik:
Removed version and API check

Issue #957, opened 2016-10-12
Resources redistribution between suborganizations


ashumilova:
Implement API that makes it possible to redistribute resources of organization between suborganizations.
It should be implemented in next way:
- By default all suborganizations is not able to use resources of parent organization.
- User who has manageResources permission for parent organization can define resources limit that is available for usage of a suborganization.
  When user allows suborganization to use some resources they become unavailable for his parent organization.

For example, user has 4 gb of RAM for his organization, it has 2 suborganizations (A and B).
He grants team A to use 1 gb and team B to use 2 gb. So 4-2-1= 1 gb is available for usage by parent organization's workspaces now.

PR #1141, opened 2016-11-14
CODENVY-957 Implement resources redistribution between suborganizations

sleshchenko:
### What does this PR do?
Added ability to redistribute resources between suborganizations.

Introduced new state for account's resources, it's called `reserved resources`. It means that account's license has resources but it can't use it for consumtion by workpace starting, creation etc.

Implemented new organization resources API that allows to define resources that should be provided for usage by suborganization from its parent organization. Only users who have `manageResources` permissions can grant suborganization to use parent organization's resources. Resources become unavailable for his parent organization when user allows suborganization to use them, so they become reserved resources. By default all suborganizations don't have any resources limit and it means that they are not able to use resources of their parent organizations.

https://github.com/codenvy/codenvy/issues/957

### New Behavior
Organization can provide own resources to usage by their suborganizations.

### Tests written?
Yes

### Docs requirements?
1.  API changes  
2.  User docs changes


sleshchenko:
@skabashnyuk @akorneta @evoevodin Please take a look my changes

sleshchenko:
These changes should be merged with changes for https://github.com/codenvy/codenvy/issues/1198

wsmaster/codenvy-hosted-api-organization/src/main/java/com/codenvy/organization/api/resource/OrganizationResourcesLimitManager.java
```diff
@@ -0,0 +1,286 @@
...
+     *         when organization with {@code resourcesLimit.getId()} was not found
+     * @throws ConflictException
+     *         when parent organization doesn't have enough resources to increase limit
+     * @throws ConflictException
+     *         when limit can't be changed because suborganization is using existed resources
+     *         or when they are redistributed to next organizations level
+     * @throws ServerException
+     *         when any other error occurs
+     */
+    public void setResourcesLimit(OrganizationResourcesLimit resourcesLimit) throws NotFoundException,
```
evoevodin: The name of the method looks inconsistent with get, i mean it's should be either
```java
setResourcesLimit 
getResourcesLimit
```

or(i like this one)
```
setLimit
getLimit
```

or event

```
set
get
```
but not 
```
setResourcesLimit
get
```

wsmaster/codenvy-hosted-api-organization/src/main/java/com/codenvy/organization/api/resource/OrganizationResourcesLimitManager.java
```diff
@@ -0,0 +1,286 @@
...
+     * @throws ConflictException
+     *         when parent organization doesn't have enough resources to increase limit
+     * @throws ConflictException
+     *         when limit can't be changed because suborganization is using existed resources
+     *         or when they are redistributed to next organizations level
+     * @throws ServerException
+     *         when any other error occurs
+     */
+    @VisibleForTesting
+    void checkResourcesAvailability(String suborganizationId,
```
evoevodin: What if we change this method to `isAvailable(...)` which doesn't throw any exceptions but returns boolean value instead, along with that we can provide another method similar to existing one, let's say it's `checkAvailable` which will throw the exception? BTW `ForbiddenException` may be good candidate when there is resources context.
The profit that i see, is that if method returns boolean there is a change to write some custom logic around it without 
usage of `try {} catch (ConflictException x) {}`.
Anyway `OrganizationResourcesLimitManager.checkResourcesAvailablity` looks to verbose to me.

wsmaster/codenvy-hosted-api-organization/src/main/java/com/codenvy/organization/api/resource/OrganizationResourcesLimitManager.java
```diff
@@ -0,0 +1,286 @@
...
+        final Map<String, Resource> existedResourceToType = existedResources.stream()
+                                                                            .collect(Collectors.toMap(Resource::getType,
+                                                                                                      Function.identity()));
+
+        for (Resource newResource : newResources) {
+            final Resource existed = existedResourceToType.get(newResource.getType());
+            if (existed != null) {
+                try {
+                    final Resource deducted = resourceAggregator.deduct(existed, newResource);
+                    // existed resource amount is more than new one
```
evoevodin: is greater than

wsmaster/codenvy-hosted-api-organization/src/main/java/com/codenvy/organization/api/permissions/OrganizationResourceDistributionServicePermissionsFilter.java
```diff
@@ -0,0 +1,92 @@
...
+        final String methodName = genericMethodResource.getMethod().getName();
+
+        final Subject currentSubject = EnvironmentContext.getCurrent().getSubject();
+        String organizationId;
+        switch (methodName) {
+            case DISTRIBUTE_RESOURCES_METHOD:
+            case RESET_DISTRIBUTED_RESOURCES:
+                //we should check permissions on parent organization level
+                String suborganizationId = (String)arguments[0];
+                final String parentId = organizationManager.getById(suborganizationId).getParent();
```
evoevodin: i don't see any purpose from having 2 more variables.
e.g the following code looks also ok to me:
```java
organizationId = organizationManager.getById((String)arguments[0]).getParent();
if (organizationid == null) {
    return;
}
break;
```

wsmaster/codenvy-hosted-api-organization/src/main/java/com/codenvy/organization/api/permissions/OrganizationResourceDistributionServicePermissionsFilter.java
```diff
@@ -0,0 +1,92 @@
...
+                    return;
+                }
+                break;
+
+            default:
+                throw new ForbiddenException("The user does not have permission to perform this operation");
+        }
+
+
+        if (!currentSubject.hasPermission(DOMAIN_ID, organizationId, OrganizationDomain.MANAGE_RESOURCES)) {
```
evoevodin: Also you use `MANAGE_ORGANIZATION_ACTION` as static import 
```
import static com.codenvy.organization.api.permissions.OrganizationPermissionsFilter.MANAGE_ORGANIZATIONS_ACTION;
```
but you don't do it with `OrganizationDomain.MANAGE_RESOURCES` is there any regularity i don't see?

wsmaster/codenvy-hosted-api-organization/src/main/java/com/codenvy/organization/api/resource/OrganizationResourceLockKeyProvider.java
```diff
@@ -0,0 +1,67 @@
...
+        String currentOrganizationId = accountId;
+        try {
+            Organization organization = organizationManager.getById(currentOrganizationId);
+            while (organization.getParent() != null) {
+                currentOrganizationId = organization.getParent();
+                organization = organizationManager.getById(currentOrganizationId);
+            }
+            return organization.getId();
+        } catch (NotFoundException e) {
+            throw new ServerException(format("Organization with id '%s' was not found during resources locking.",
```
evoevodin: > during resources locking

 looks like you just trying to get a lock key, looks like the message should be a bit different

wsmaster/codenvy-hosted-api-organization/src/main/java/com/codenvy/organization/api/resource/OrganizationResourcesDistributionService.java
```diff
@@ -0,0 +1,166 @@
...
+
+    @Inject
+    public OrganizationResourcesDistributionService(OrganizationResourcesDistributor resourcesDistributor,
+                                                    ResourceValidator resourceValidator) {
+        this.resourcesDistributor = resourcesDistributor;
+        this.resourceValidator = resourceValidator;
+    }
+
+    @POST
+    @Path("{suborganizationId}")
```
evoevodin: `reset` method path starts with `/` while this path doesn't, i find it a bit inconsistent don't you?

wsmaster/codenvy-hosted-api-organization/src/main/java/com/codenvy/organization/api/resource/OrganizationResourcesDistributionService.java
```diff
@@ -0,0 +1,166 @@
...
+    @Consumes(APPLICATION_JSON)
+    @ApiOperation(value = "Distribute resources for suborganization.",
+                  notes = "Distributed resources is unavailable for usage by parent organization." +
+                          "It should reset resources distribution before consuming.",
+                  response = OrganizationDistributedResourcesDto.class)
```
evoevodin: lie, this method returns nothing

wsmaster/codenvy-hosted-api-organization/src/main/java/com/codenvy/organization/api/resource/OrganizationResourcesDistributionService.java
```diff
@@ -0,0 +1,166 @@
...
+                       .entity(distributedResourcesPage.getItems(DtoConverter::asDto))
+                       .header("Link", createLinkHeader(distributedResourcesPage))
+                       .build();
+    }
+
+    @DELETE
+    @Path("/{organizationId}")
+    @ApiOperation(value = "Reset resources distribution.",
+                  notes = "After distribution resetting suborganization won't be able to use parent resources.")
+    @ApiResponses({@ApiResponse(code = 201, message = "Resources distribution successfully reset"),
```
evoevodin: why 201, it's 204 i think

wsmaster/codenvy-hosted-api-organization/src/main/java/com/codenvy/organization/api/resource/OrganizationResourcesDistributor.java
```diff
@@ -0,0 +1,319 @@
...
+                                            OrganizationManager organizationManager) {
+        this.organizationDistributedResourcesDao = organizationDistributedResourcesDao;
+        this.resourcesLocks = resourcesLocks;
+        this.usageManager = usageManager;
+        this.resourceAggregator = resourceAggregator;
+        this.organizationManager = organizationManager;
+    }
+
+    /**
+     * TODO Revise doc
```
evoevodin: do it! :rocket:

wsmaster/codenvy-sql-schema/src/main/resources/codenvy-schema/5.0.0-M8/1.2__add_org_resources_limit.sql
```diff
@@ -0,0 +1,37 @@
...
+-- if any.  The intellectual and technical concepts contained
+-- herein are proprietary to Codenvy S.A.
+-- and its suppliers and may be covered by U.S. and Foreign Patents,
+-- patents in process, and are protected by trade secret or copyright law.
+-- Dissemination of this information or reproduction of this material
+-- is strictly forbidden unless prior written permission is obtained
+-- from Codenvy S.A..
+--
+
+-- Organization distributed resources  ))---------------------------------------------------------
```
akorneta: delete the brackets please

wsmaster/codenvy-hosted-api-organization/src/main/java/com/codenvy/organization/spi/OrganizationDistributedResourcesDao.java
```diff
@@ -0,0 +1,80 @@
...
+import org.eclipse.che.api.core.ServerException;
+
+/**
+ * Defines data access object contract for {@link OrganizationDistributedResourcesImpl}.
+ *
+ * @author Sergii Leschenko
+ */
+public interface OrganizationDistributedResourcesDao {
+    /**
+     * Stores distributed resources for suborganization.
```
akorneta: I guess it will be helpful to describe here how to store resources when there are already exist resources for the organization.

musienko-maxim:
After running the selenium tests new regression has not been found

PR #1147, opened 2016-11-15
Prototype of upgrade method

### LINKED ISSUE #1033: CLI codenvy upgrade command
b. "codenvy upgrade y" will do the following:
---- Access the manifest that is in ~/.codenvy/manifests/upgrades to see if x -> y is allowed
---- If not allowed, terminate with friendly error message
---- If allowed, perform upgrade sequence of (cmd_stop(), cmd_backup()?, cmd_download() to get new images, cmd_start() (which does a config and launches Codenvy tomcat which does migration), and then if any errors along the way, some sort of rollback.

mshaposhnik:
### What does this PR do?
Adds method for upgrading codenvy

Fixes #1033

### DIFF
--- a/dockerfiles/cli/cli.sh
+++ b/dockerfiles/cli/cli.sh
@@ -1003,23 +1003,28 @@ cmd_upgrade() {
 
   # If here, this version is validly upgradeable.  You can upgrade from
   # $(get_installed_version) to $1
+  ## Download version images
+  info "upgrade" "Downloading $1 images..."
+  get_image_manifest ${1}
+  IFS=$'\n'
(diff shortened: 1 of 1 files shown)


mshaposhnik:
@TylerJewell @skabashnyuk @garagatyi pls review

TylerJewell:
1. Images should be downloaded and pulled before stopping services.  This is necessary to limit user maintenance window.

2. There are no messages to the admin on the console explaining the steps if ehat is happening.

3.  If there is an error in the upgrade process what is the rollback method?

mshaposhnik:
1-2 OK
3 for now i'ts only restoring of latest backup made during update

Issue #1156, opened 2016-11-15
all docker image names and versions must be used from images manifests for particular version


riuvshin:
Today we have hardcoded in several places docker images names and it versions. We must use docker images and versions from version manifest.

riuvshin:
done

PR #1157, opened 2016-11-15
use images names and versions from manifest file for particular version

riuvshin:
needed for: https://github.com/codenvy/codenvy/issues/1156
- compose now used images with versions from `images` manifest file.
- improved dev mode, now it prints really needed puppet output to console.
- remove unnecessary sed commands
@TylerJewell @garagatyi Please take a look.

### DIFF
--- a/dockerfiles/cli/cli.sh
+++ b/dockerfiles/cli/cli.sh
@@ -668,6 +668,7 @@ generate_configuration_with_puppet() {
   info "config" "Generating $CHE_MINI_PRODUCT_NAME configuration..."
   # Note - bug in docker requires relative path for env, not absolute
   log "docker_run -it --env-file=\"${REFERENCE_CONTAINER_ENVIRONMENT_FILE}\" \
+                  --env-file=/version/$CODENVY_VERSION/images \
                   -v \"${CODENVY_HOST_INSTANCE}\":/opt/codenvy:rw \
                   -v \"${CODENVY_HOST_CONFIG_MANIFESTS_FOLDER}\":/etc/puppet/manifests:ro \
                   -v \"${CODENVY_HOST_CONFIG_MODULES_FOLDER}\":/etc/puppet/modules:ro \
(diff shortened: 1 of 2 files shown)


dockerfiles/cli/cli.sh
```diff
@@ -832,18 +834,15 @@ cmd_config() {
   fi
 
   # Run the docker configurator
-  generate_configuration_with_puppet 
+  if [ "${CODENVY_DEVELOPMENT_MODE}" = "on" ]; then
+    # generate configs and print puppet output logs to console if dev mode is on
+    generate_configuration_with_puppet
+  else
+    generate_configuration_with_puppet >> "${LOGS}"
+  fi
```
TylerJewell: Why did you add this back in?  I removed this "if" logic earlier because it was suppressing certain output that was important.

TylerJewell:
This is a very slick way to make sure that the docker-compose images.  But we should discuss the if logic around the configurator for dev mode.

modules/compose/templates/docker-compose.yml.erb
```diff
@@ -83,9 +83,9 @@ services:
 # DOCKER REGISTRY
 ########################
   registry:
-    image: registry:2.5.0
+    image: <%= ENV["IMAGE_REGISTRY"] %>
     env_file:
-      - '<%= scope.lookupvar('compose::codenvy_folder') -%>/config/registry/registry.env'
+      - '/codenvy/instance/config/registry/registry.env'
```
TylerJewell: By hard coding this AND removing the sed commands, we lose the ability to generate two docker-compose.yml files. If we want to generate docker-compose-container.yml and docker-compose-host.yml at the same time, we need to use sed to change the env file.s

dockerfiles/cli/cli.sh
TylerJewell: This is ok to leave.

TylerJewell:
I am a +1 for merge.  We will solve issues with doing multiple compose templates in another issue.

dockerfiles/cli/cli.sh
riuvshin: this is really needed for dev mode, dev don't want to go to CLI logs to see puppet logs which is really needed to see is their changes applied properly

garagatyi: +1 it is needed

PR #1158, opened 2016-11-15
Several compose files

riuvshin:
It provides another compose file which could be launched with native docker-compose

@TylerJewell


TylerJewell:
@riuvshin - +1, tested on Windows and it worked.

However, if you merge this change we must rebuild codenvy/init, codenvy/init:nightly and any other tags. This fundamentally changes the name of certain files and my cached version of codenvy/init fails and needed to be rebuilt.

dockerfiles/cli/codenvy.sh
```diff
@@ -250,15 +236,15 @@ check_docker() {
...
-        info "Welcome to Codenvy!"
-        info ""
-        info "We did not detect a valid DOCKER_HOST." 
-        info ""
-        info "Rerun the CLI:"
-        info "  docker run -it --rm -v /var/run/docker.sock:/var/run/docker.sock "
-        info "                      -v <local-path>:/codenvy "
-        info "                         codenvy/cli $@"    
+      if ! docker ps > /dev/null 2>&1; then
+        printf "${GREEN}INFO:${NC} Welcome to Codenvy!\n"
```
garagatyi: Why not rewrite info command to add green prefix and add new line at the end of line instead of rewriting all the lines?

dockerfiles/cli/cli.sh
```diff
@@ -1015,7 +1032,6 @@ cmd_version() {
   text "$CHE_PRODUCT_NAME:\n"
   text "  Version:      %s\n" $(get_installed_version)
   text "  Installed:    %s\n" $(get_installed_installdate)
-  text "  CLI version:  $CHE_CLI_VERSION\n"
```
garagatyi: Do we log it anywhere? How could support know which version is called?

docs/README.md
```diff
@@ -359,7 +360,7 @@ User data is stored in:
 Instance configuration is generated by Codenvy and is updated by our internal configuration utilities. These 'generated' configuration files should not be modified and stored in:
 ```
 /codenvy.var                       # Version of Codenvy installed
-/docker-compose.yml                # Docker compose to launch internal services
+/docker-compose-container.yml                # Docker compose to launch internal services
```
garagatyi: Please fix indentation

modules/compose/templates/docker-compose-container.yml.erb
```diff
@@ -0,0 +1,150 @@
+# ###################################
+# This file is generated by puppet
+# PLEASE DON'T MODIFY BY HAND
+# ###################################
+version: '2'
+services:
+
+
+########################
```
garagatyi: Can you elaborate on why we need 2 compose files?

dockerfiles/cli/codenvy.sh
TylerJewell: The info() command requires use of LOGS which is not yet initialized.  I needed to move log initialization until after check mounts and I wanted to make it clear that this output was raw in the bootstrap phase.

dockerfiles/cli/cli.sh
TylerJewell: The cli and the product version are now identical because the cli is in a container.  It is logged in the line above.

modules/compose/templates/docker-compose-container.yml.erb
TylerJewell: The format of the compose file in a container is not suitable to run on the host.  We now generate a second file that can be launched directly on the host.

dockerfiles/cli/codenvy.sh
garagatyi: ok

dockerfiles/cli/cli.sh
garagatyi: ok

modules/compose/templates/docker-compose-container.yml.erb
garagatyi: Why do we need to run it directly on a host if we have dockerized version?

TylerJewell: For debugging purposes.

garagatyi: I'm not sure that it will be a real debugging if you have 2 different (even slightly) compose files and they will be executed in a different ways

docs/README.md
TylerJewell: fixed

PR #1160, opened 2016-11-16
Fix checking of existence last 'setPermissions' action before permissions changing

sleshchenko:
### What does this PR do?
Fix checking of existence last 'setPermissions' action before permissions changing

### Tests written?
Yes


sleshchenko:
@skabashnyuk @evoevodin Please take a look

PR #1161, opened 2016-11-16
Rework database schema loading

evoevodin:
Adds sql schema for codenvy entities which is basically extracted from the one generated by EclipseLink.
This issue depends on merge of https://github.com/eclipse/che/pull/3063, all the information about schema loading strategy and configuration properties is described in that issue.

@skabashnyuk, @sleshchenko, @akorneta please review


wsmaster/codenvy-hosted-api-machine/src/test/java/com/codenvy/api/machine/server/jpa/OnPremisesJpaRecipeDaoTest.java
```diff
@@ -127,6 +130,7 @@ public void cleanup() {
     @AfterClass
     public void shutdown() throws Exception {
         manager.getEntityManagerFactory().close();
+        H2TestHelper.shutdownDefault();
```
sleshchenko: Can you explain why we should not do the same for JpaRecipePermissionsDaoTest?

wsmaster/codenvy-sql-schema/src/main/resources/codenvy-schema/5.0.0-M8/1.1__init.sql
```diff
@@ -0,0 +1,362 @@
...
+-- constraints
+ALTER TABLE recipepermissions_actions ADD CONSTRAINT fk_recipepermissions_actions_recipepermissions_id FOREIGN KEY (recipepermissions_id) REFERENCES recipepermissions (id);
+--------------------------------------------------------------------------------
+
+
+-- Organization ----------------------------------------------------------------
+CREATE TABLE organization (
+    id          VARCHAR(255)         NOT NULL,
+    parent      VARCHAR(255),
+    account_id  VARCHAR(255)         NOT NULL,
```
sleshchenko: Should we not use `_` symbol for fields names? Looks like you do it in both ways: with _ for [workers](https://github.com/codenvy/codenvy/pull/1161/files#diff-6cd6eca6d13e6148e2265297d10383a3R216) and without `_` for [member](https://github.com/codenvy/codenvy/pull/1161/files#diff-6cd6eca6d13e6148e2265297d10383a3R303)

PR #1166, opened 2016-11-16
Codenvy-1127  docker instance implementation that that limits number of simultaneous container commits on the same node.

### LINKED ISSUE #1127: Avoid of concurrent snapshots on same node.
As we are going to enable auto snapshotting by default we have to implement some locking logic which will prevent concurrent snapshotting on particular physical node to minimize performance affect to the host.
We need this because docker commit and push are expensive operations in terms of hardware resources.

mshaposhnik:
### What does this PR do?
Adds docker instance implementation that limits number of simultaneous container commits on the same node at the same time.

Fixes #1127 

### Previous Behavior
Multiple container commits may be possible on same node.

### New Behavior
Commits are one-by-one

### Docs requirements?
None

### DIFF
--- a/plugins/plugin-hosted/codenvy-machine-hosted/src/main/java/com/codenvy/machine/HostedDockerInstance.java
+++ b/plugins/plugin-hosted/codenvy-machine-hosted/src/main/java/com/codenvy/machine/HostedDockerInstance.java
@@ -74,14 +74,14 @@ public class HostedDockerInstance extends DockerInstance {
 
     @Override
     protected void commitContainer(String repository, String tag) throws IOException {
-        final Semaphore nodeSemahore = getSemaphore(getNode().getHost());
+        final Semaphore nodeSemaphore = getSemaphore(getNode().getHost());
         try {
-            nodeSemahore.acquire();
+            nodeSemaphore.acquire();
             super.commitContainer(repository, tag);
         } catch (InterruptedException e) {
             Thread.currentThread().interrupt();
         } finally {
-            nodeSemahore.release();
+            nodeSemaphore.release();
         }
     }
 
@@ -89,7 +89,7 @@ public class HostedDockerInstance extends DockerInstance {
         Semaphore semaphore = SEMAPHORES.get(key);
         if (semaphore == null) {
             Semaphore newSemaphore = new Semaphore(concurrentCommits, true);
-            semaphore = SEMAPHORES.putIfAbsent(key,newSemaphore);
+            semaphore = SEMAPHORES.putIfAbsent(key, newSemaphore);
             if (semaphore == null) {
                 semaphore = newSemaphore;
             }
--- a/plugins/plugin-hosted/codenvy-machine-hosted/src/main/java/com/codenvy/machine/HostedDockerInstance.java
+++ b/plugins/plugin-hosted/codenvy-machine-hosted/src/main/java/com/codenvy/machine/HostedDockerInstance.java
@@ -33,6 +33,8 @@ import java.io.IOException;
 import java.util.concurrent.locks.Lock;
 
 /**
+ * Docker instance implementation that prevents multiple container commits on the same node at the same time.
+ *
  * @author Max Shaposhnik
  */
 public class HostedDockerInstance extends DockerInstance {
--- a/plugins/plugin-hosted/codenvy-machine-hosted/src/main/java/com/codenvy/machine/HostedDockerInstance.java
+++ b/plugins/plugin-hosted/codenvy-machine-hosted/src/main/java/com/codenvy/machine/HostedDockerInstance.java
@@ -14,7 +14,6 @@
  */
 package com.codenvy.machine;
 
-import com.google.common.util.concurrent.Striped;
 import com.google.inject.assistedinject.Assisted;
 
 import org.eclipse.che.api.core.model.machine.Machine;
@@ -30,16 +29,19 @@ import org.eclipse.che.plugin.docker.machine.node.DockerNode;
 import javax.inject.Inject;
 import javax.inject.Named;
 import java.io.IOException;
-import java.util.concurrent.locks.Lock;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.Semaphore;
 
 /**
- * Docker instance implementation that prevents multiple container commits on the same node at the same time.
+ * Docker instance implementation that limits number of simultaneous container commits on the given node.
  *
  * @author Max Shaposhnik
  */
 public class HostedDockerInstance extends DockerInstance {
 
-    private static final Striped<Lock> NODE_LOCKS = Striped.lazyWeakLock(16);
+    private static final Map<String, Semaphore> SEMAPHORES = new ConcurrentHashMap<>();
+    private static int concurrentCommits;
 
     @Inject
     public HostedDockerInstance(DockerConnector docker,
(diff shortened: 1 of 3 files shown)


plugins/plugin-hosted/codenvy-machine-hosted/src/main/java/com/codenvy/machine/HostedDockerInstance.java
```diff
@@ -0,0 +1,99 @@
...
+    private Semaphore getSemaphore(String key) {
+        Semaphore semaphore = SEMAPHORES.get(key);
+        if (semaphore == null) {
+            Semaphore newSemaphore = new Semaphore(concurrentCommits, true);
+            semaphore = SEMAPHORES.putIfAbsent(key,newSemaphore);
+            if (semaphore == null) {
+                semaphore = newSemaphore;
+            }
+        }
+        System.out.println("~~~~~~~~~~~~" + semaphore);
```
tolusha: Is it debug info ?

mshaposhnik: Yeah

plugins/plugin-hosted/codenvy-machine-hosted/src/main/java/com/codenvy/machine/HostedDockerInstance.java
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    @Override
+    protected void commitContainer(String repository, String tag) throws IOException {
+        final Semaphore nodeSemaphore = getSemaphore(getNode().getHost());
+        try {
+            nodeSemaphore.acquire();
+            super.commitContainer(repository, tag);
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
```
mshaposhnik: Need to throw smthng here, but only IO is possible. WDYT ?

tolusha: ` throw new IOException(e)` ?

garagatyi: +1

plugins/plugin-hosted/codenvy-machine-hosted/src/test/java/com/codenvy/machine/HostedDockerInstanceTest.java
```diff
@@ -0,0 +1,200 @@
...
+        executor.execute(() -> performCommit(repo1, TAG));
+        executor.execute(() -> performCommit(repo2, TAG));
+        waitingAnswer1.waitAnswerCall(1, TimeUnit.SECONDS);
+        waitingAnswer2.waitAnswerCall(1, TimeUnit.SECONDS);
+
+        // when
+        executor.execute(() -> performCommit(repo3, TAG));
+
+        // thread #3 should wait - semaphore is red
+        verify(dockerConnectorMock, never()).commit(Matchers.argThat(new CommitParamsMatcher(repo3)));
```
garagatyi: I think that timeout is needed here, otherwise there is no guarantee that 3rd thread was started. Also it would be nice to verify that Instance.commit is called

plugins/plugin-hosted/codenvy-machine-hosted/src/test/java/com/codenvy/machine/HostedDockerInstanceTest.java
```diff
@@ -0,0 +1,200 @@
...
+        WaitingAnswer<Void> waitingAnswer2 = new WaitingAnswer<>();
+
+        doAnswer(waitingAnswer1).when(dockerConnectorMock).commit(Matchers.argThat(new CommitParamsMatcher(repo1)));
+        doAnswer(waitingAnswer2).when(dockerConnectorMock).commit(Matchers.argThat(new CommitParamsMatcher(repo2)));
+
+        // Starting threads #1 & 2
+        executor.execute(() -> performCommit(repo1, TAG));
+        executor.execute(() -> performCommit(repo2, TAG));
+        waitingAnswer1.waitAnswerCall(1, TimeUnit.SECONDS);
+        waitingAnswer2.waitAnswerCall(1, TimeUnit.SECONDS);
```
garagatyi: repo2 also won't commit until repo1 commit is finished. Maybe remove it?

plugins/plugin-hosted/codenvy-machine-hosted/src/test/java/com/codenvy/machine/HostedDockerInstanceTest.java
```diff
@@ -0,0 +1,200 @@
...
+    @BeforeMethod
+    public void setUp() throws IOException {
+        when(dockerNode.getHost()).thenReturn("host1");
+        dockerInstance = getDockerInstance();
+        executor = Executors.newFixedThreadPool(3);
+    }
+
+
+    @Test
+    public void shouldBeAbleToCommitSimultaneously() throws Exception {
```
garagatyi: Please add test for different nodes

plugins/plugin-hosted/codenvy-machine-hosted/src/test/java/com/codenvy/machine/HostedDockerInstanceTest.java
mshaposhnik: It will - i set permits to 2

garagatyi: Oh, sorry, I missed this

plugins/plugin-hosted/codenvy-machine-hosted/src/main/java/com/codenvy/machine/HostedDockerInstance.java
mshaposhnik: Ok

plugins/plugin-hosted/codenvy-machine-hosted/src/main/java/com/codenvy/machine/HostedDockerInstance.java
```diff
@@ -0,0 +1,101 @@
...
+            throw new IOException(e);
```
garagatyi: new IOException(e, e.getLocalizedMessage());?

plugins/plugin-hosted/codenvy-machine-hosted/src/main/java/com/codenvy/machine/HostedDockerInstance.java
```diff
@@ -0,0 +1,101 @@
...
+import java.util.concurrent.Semaphore;
+
+/**
+ * Docker instance implementation that limits number of simultaneous container commits on the given node.
+ *
+ * @author Max Shaposhnik
+ */
+public class HostedDockerInstance extends DockerInstance {
+
+    private static final Map<String, Semaphore> SEMAPHORES = new ConcurrentHashMap<>();
```
evoevodin: I wouldn't say that `CONSTANT_CASE` is the right one for the mutable value, see [examples](https://github.com/codenvy/codenvy-dev-resources/wiki/Naming#constant-names)

plugins/plugin-hosted/codenvy-machine-hosted/src/test/java/com/codenvy/machine/HostedDockerInstanceTest.java
```diff
@@ -0,0 +1,244 @@
...
+
+        // thread #3 run on other node
+        when(dockerNode.getHost()).thenReturn("host2");
+        // when
+        executor.execute(() -> performCommit(repo3, TAG));
+
+        Thread.sleep(200); //to allow thread # 3 start
+
+        // thread #3 commit executed too
+        verify(dockerInstance).commitContainer(eq(repo3), eq(TAG));
```
garagatyi: You can use mockito times instead of Thread.sleep here, it can boost test

PR #1170, opened 2016-11-16
Add Codenvy License Acceptance to Docker Installer

TylerJewell:
### What does this PR do?
During the 'init' phase of running the Codenvy CLI, we ask the user to accept the Fair Source 3 license hosted on Codenvy's Web site.

Also, per requests, cleaned up some of the log management to remove unnecessary printf statements.

@riuvshin @benoitf


TylerJewell:
I thought about it, and for now, you must accept the license on each time that you do an init().  If the system is already initialized, then you do not need to accept the license.

I thought about adding a --accept-license flag, but I am not sure we need it just yet.

riuvshin:
+1

PR #1175, opened 2016-11-17
fix license accept, remove cli logs from instance folder

riuvshin:
@TylerJewell


riuvshin:
+1 to @TylerJewell  changes

Issue #1151, opened 2016-11-15
Allow to suspend API server start in development mode


garagatyi:
Sometimes developers need to debug API server bootstrapping. It is very hard with codenvy in container. 
To do that environment variable **JPDA_SUSPEND** consumed by tomcat should be set to **y**.

riuvshin:
Thanks @garagatyi !

PR #1180, opened 2016-11-17
add possibility to set JPDA_SUSPEND via ENV var

riuvshin:
needed for: https://github.com/codenvy/codenvy/issues/1151

added possiblity to enable jpda suspend mode and change debug port

@garagatyi @TylerJewell please review

### DIFF
--- a/dockerfiles/cli/cli.sh
+++ b/dockerfiles/cli/cli.sh
@@ -687,7 +687,7 @@ generate_configuration_with_puppet() {
                   -e "REGISTRY_ENV_FILE=${REGISTRY_ENV_FILE}" \
                   -e "POSTGRES_ENV_FILE=${POSTGRES_ENV_FILE}" \
                   -e "CODENVY_ENV_FILE=${CODENVY_ENV_FILE}" \
-                  -e "ENABLE_JPDA_SUSPEND=${ENABLE_JPDA_SUSPEND:-false}" \
+                  -e "CODENVY_DEBUG_SUSPEND=${CODENVY_DEBUG_SUSPEND:-false}" \
                       $IMAGE_PUPPET \
                           apply --modulepath \
                                 /etc/puppet/modules/ \
@@ -700,7 +700,7 @@ generate_configuration_with_puppet() {
                   -e "REGISTRY_ENV_FILE=${REGISTRY_ENV_FILE}" \
                   -e "POSTGRES_ENV_FILE=${POSTGRES_ENV_FILE}" \
                   -e "CODENVY_ENV_FILE=${CODENVY_ENV_FILE}" \
-                  -e "ENABLE_JPDA_SUSPEND=${ENABLE_JPDA_SUSPEND:-false}" \
+                  -e "CODENVY_DEBUG_SUSPEND=${CODENVY_DEBUG_SUSPEND:-false}" \
                       $IMAGE_PUPPET \
                           apply --modulepath \
                                 /etc/puppet/modules/ \
--- a/dockerfiles/cli/cli.sh
+++ b/dockerfiles/cli/cli.sh
@@ -687,7 +687,6 @@ generate_configuration_with_puppet() {
                   -e "REGISTRY_ENV_FILE=${REGISTRY_ENV_FILE}" \
                   -e "POSTGRES_ENV_FILE=${POSTGRES_ENV_FILE}" \
                   -e "CODENVY_ENV_FILE=${CODENVY_ENV_FILE}" \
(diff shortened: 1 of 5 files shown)


TylerJewell:
Add improvement to docs/README.md in sections where we discuss development mode

riuvshin:
@TylerJewell please review docs changes

dockerfiles/cli/cli.sh
```diff
@@ -687,6 +687,7 @@ generate_configuration_with_puppet() {
                   -e "REGISTRY_ENV_FILE=${REGISTRY_ENV_FILE}" \
                   -e "POSTGRES_ENV_FILE=${POSTGRES_ENV_FILE}" \
                   -e "CODENVY_ENV_FILE=${CODENVY_ENV_FILE}" \
+                  -e "ENABLE_JPDA_SUSPEND=${ENABLE_JPDA_SUSPEND:-false}" \
```
benoitf: it looks strange that we need to set extra env value there It's not in codenvy.env ?

benoitf:
it looks strange that property is not having any prefix. It's JPDA debug for which container ? 
CODENVY_DEBUG_MODE=true
CODENVY_DEBUG_SUSPEND=true

then it gives JPDA option to tomcat (but if you replace tomcat by jetty, option will stay the same for user)

benoitf:
example of properties in che launcher:

CHE_DEBUG_SERVER="false"
CHE_DEBUG_SERVER_PORT="8000"
CHE_DEBUG_SERVER_SUSPEND="false"

riuvshin:
@benoitf renamed env var to `CODENVY_DEBUG_SUSPEND`

benoitf:
I still don't get why 
` -e "CODENVY_DEBUG_SUSPEND=${CODENVY_DEBUG_SUSPEND:-false}" \`
is given to the CLI directly.
if each parameter is added like that it will unmaintainable, no ?


I only see parameter in codenvy.env or/and in docker-compose files
In dev mode, you could set ENV var locally and then call docker-compose (native, not containerized)
or you can also tweak codenvy.env before updating the conf.

riuvshin:
Good point, I thought it is very rare use case when you need that suspend so I don't wanted to add this codenvy.env. But in general I kind of agree with you having different flow for setting configurations is not good. So I will move this to codenvy.env I think.

riuvshin:
ah, I forget that I've decided to enable it this way because "dev mode" enable also via mount in CLI

benoitf:
It's not consistent at all
if you want to change debug port, you will not give another parameter to CLI directly as well

TylerJewell:
Do we really need to let people to change the debug port?  It sounds like a nice option, but it may add complexity for a feature that <.0001% of people would use.

benoitf:
it's not because I'm using a local binary that I also want that my server run in debug mode as well.
debug port was only an example .
If any development mode option is becoming a new direct parameter it's not working

riuvshin:
I reworked everything now it is configurable in `codenvy.env` wihtout CLI modifications
docs updated also.
@benoitf @TylerJewell @garagatyi please review

modules/codenvy/templates/codenvy.env.erb
```diff
@@ -10,14 +10,16 @@ PUPPET_SOURCE=<%= scope.lookupvar('codenvy::puppet_src_folder') -%>
 PUPPET_DESTINATION=<%= scope.lookupvar('codenvy::codenvy_folder') -%>
 
 PGUSER=<%= scope.lookupvar('codenvy::pgsql_username') %>
-PGPASSWORD=<%= scope.lookupvar('codenvy::pgsql_pass') %>
+PGPASSWORD=<%= scope.lookupvar('codenvy::pgsql_pass') -%>
```
benoitf: is - a typo ?

riuvshin: no it is a ruby thing it will prevent of new line generation

benoitf:
you could optionally reword description of the PR as example is not up-to-date

TylerJewell:
I updated the documentation in the che.env file:
1. The DEVELOPMENT_MODE property is no longer required as it is set in the CLI based upon :/repo volume mount.
2. Same for the TOMCAT reference.
3. Provided deeper description and formatting for JPDA mode.

riuvshin:
@TylerJewell no, we must have `CODENVY_ENVIRONMENT` ENV property because it is used in puppet CLI should set it if mount is provided. but without that property puppet will not be able to know that dev mode is on

TylerJewell:
Yes you are right.  We need one but not the other.  I will revert that one change.

TylerJewell:
Fixed

PR #1181, opened 2016-11-17
CODENVY-657: Do not log SourceNotFoundException

vinokurig:
### What does this PR do?
Do not log `SourceNotFoundException` when starting workspace.

### Previous Behavior
If user will create a worcspace from config with wrong recipe location and then will start it, ERROR will appear in logs. This situation shouldn't be logged because it is normal situation.

### Tests written?
No

Related to https://github.com/eclipse/che/pull/3084

@garagatyi please review


plugins/plugin-hosted/codenvy-machine-hosted/src/main/java/com/codenvy/machine/HostedMachineProviderImpl.java
```diff
@@ -151,6 +153,10 @@ protected void pullImage(CheServiceImpl service,
                                               // don't build an image on a node under maintenance
                                               .addBuildArg(MAINTENANCE_CONSTRAINT_KEY, MAINTENANCE_CONSTRAINT_VALUE),
                               progressMonitor);
+
+        }
```
garagatyi: Please fix formatting

PR #1184, opened 2016-11-17
CLI Improvements

TylerJewell:
### What does this PR do?
1. Adds a '--cli' parameter to `codenvy destroy` which, if `:/cli` is volume mounted, then will destroy the cli.log from the host.  This is not enabled by default so you can monitor the log over time.

2. Adds a 'accept-license' parameter to the 'codenvy init' command which will auto-accept the codenvy license during installation allowing for silent mode.

3. Adds in a structure that allows for parsing multiple parameters on a single command which is generally more flexible than what is currently provided on the command line.

### DIFF
--- a/dockerfiles/cli/cli.sh
+++ b/dockerfiles/cli/cli.sh
@@ -742,7 +742,32 @@ cmd_download() {
 }
 
 cmd_init() {
-  FORCE_UPDATE=${1:-"--no-force"}
+
+  # set an initial value for the flag
+  FORCE_UPDATE="--no-force"
+  AUTO_ACCEPT_LICENSE="false"
+
+  while [ $# -gt 0 ]; do
+    case $1 in
+      --no-force)
+        FORCE_UPDATE="--no-force"
+        shift ;;
+      --force)
+        FORCE_UPDATE="--force"
+        shift ;;
+      --pull)
+        FORCE_UPDATE="--pull"
+        shift ;;
+      --offline)
+        FORCE_UPDATE="--offline"
+        shift ;;
+      --accept-license)
+        AUTO_ACCEPT_LICENSE="true"
+        shift ;;
+      *) error "Unknown parameter: $1" ; return 2 ;;
+    esac
+  done
+
   if [ "${FORCE_UPDATE}" == "--no-force" ]; then
     # If codenvy.environment file exists, then fail
     if is_initialized; then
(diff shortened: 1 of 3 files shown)


TylerJewell:
@riuvshin @benoitf

dockerfiles/cli/cli.sh
```diff
@@ -742,7 +742,32 @@ cmd_download() {
...
+        FORCE_UPDATE="--no-force"
+        shift ;;
+      --force)
+        FORCE_UPDATE="--force"
+        shift ;;
+      --pull)
+        FORCE_UPDATE="--pull"
+        shift ;;
+      --offline)
+        FORCE_UPDATE="--offline"
```
riuvshin: It is not clear how force update related to "--offline"

dockerfiles/cli/cli.sh
```diff
@@ -998,6 +1040,17 @@ cmd_destroy() {
   if has_docker_for_windows_client; then
     docker volume rm codenvy-postgresql-volume > /dev/null 2>&1  || true
   fi
+
+  # Sometimes users wnat the CLI after they have destroyed their instance
```
benoitf: small typo on "want"

TylerJewell: Yes, fixed.

TylerJewell:
Ok, made a number of fixes:
1. Fixed grab_offline_images() method to pull from /codenvy/backup instead of offline
2. Simplified the parameter analysis to compress the code.
3. Updated the docs to explain offline mode
4. Added fix for grab_offline_images() to check any parameter for --offline
5. Updated the constants to reuse CHE_MINI_PRODUCT_NAME

benoitf:
why check_host_volume_mount was moved from one shell script to the other ?

TylerJewell:
Just because it was more logical to have it where I perform all of the other volume mount checks.

benoitf:
ok because now I'm kinda lost on functions that belong to cli.sh vs functions that belong to codenvy.sh

TylerJewell:
Yes, I understand.  

So the basic idea:
1. codenvy.sh --- the minimum set of functionality necessary to guarantee that volume mounts work.
2. cli.sh --- everything else, because in theory at this point you can mount :/repo and have a dev mode.

The only exception to the rule was moving that one function into the check_volume_mounts() method because it was just logical to keep it together conceptually.

PR #1188, opened 2016-11-18
use tags for socat and puppet images

riuvshin:
@TylerJewell @skabashnyuk @garagatyi

### DIFF
--- a/dockerfiles/cli/version/5.0.0-M1/images
+++ b/dockerfiles/cli/version/5.0.0-M1/images
@@ -1,6 +1,6 @@
-IMAGE_SOCAT=codenvy/socat
+IMAGE_SOCAT=codenvy/socat:1.7.3.1
 IMAGE_HAPROXY=codenvy/haproxy:1.5.18
-IMAGE_PUPPET=codenvy/puppet
+IMAGE_PUPPET=codenvy/puppet:4.6.2
 IMAGE_NGINX=nginx:1.10-alpine
 IMAGE_SWARM=swarm:1.2.5
 IMAGE_REGISTRY=registry:2.5.0
--- a/dockerfiles/cli/version/5.0.0-M2/images
+++ b/dockerfiles/cli/version/5.0.0-M2/images
@@ -1,6 +1,6 @@
-IMAGE_SOCAT=codenvy/socat
+IMAGE_SOCAT=codenvy/socat:1.7.3.1
 IMAGE_HAPROXY=codenvy/haproxy:1.5.18
-IMAGE_PUPPET=codenvy/puppet
+IMAGE_PUPPET=codenvy/puppet:4.6.2
 IMAGE_NGINX=nginx:1.10-alpine
 IMAGE_SWARM=swarm:1.2.5
 IMAGE_REGISTRY=registry:2.5.0
--- a/dockerfiles/cli/version/5.0.0-M3/images
+++ b/dockerfiles/cli/version/5.0.0-M3/images
@@ -1,6 +1,6 @@
-IMAGE_SOCAT=codenvy/socat
+IMAGE_SOCAT=codenvy/socat:1.7.3.1
 IMAGE_HAPROXY=codenvy/haproxy:1.5.18
-IMAGE_PUPPET=codenvy/puppet
+IMAGE_PUPPET=codenvy/puppet:4.6.2
 IMAGE_NGINX=nginx:1.10-alpine
 IMAGE_SWARM=swarm:1.2.5
 IMAGE_REGISTRY=registry:2.5.0
--- a/dockerfiles/cli/version/latest/images
+++ b/dockerfiles/cli/version/latest/images
@@ -1,6 +1,6 @@
-IMAGE_SOCAT=codenvy/socat:latest
+IMAGE_SOCAT=codenvy/socat:1.7.3.1
 IMAGE_HAPROXY=codenvy/haproxy:1.5.18
-IMAGE_PUPPET=codenvy/puppet:latest
+IMAGE_PUPPET=codenvy/puppet:4.6.2
 IMAGE_NGINX=nginx:1.10-alpine
 IMAGE_SWARM=swarm:1.2.5
 IMAGE_REGISTRY=registry:2.5.0
--- a/dockerfiles/cli/version/nightly/images
+++ b/dockerfiles/cli/version/nightly/images
@@ -1,6 +1,6 @@
-IMAGE_SOCAT=codenvy/socat:latest
+IMAGE_SOCAT=codenvy/socat:1.7.3.1
 IMAGE_HAPROXY=codenvy/haproxy:1.5.18
-IMAGE_PUPPET=codenvy/puppet:latest
+IMAGE_PUPPET=codenvy/puppet:4.6.2
 IMAGE_NGINX=nginx:1.10-alpine
 IMAGE_SWARM=swarm:1.2.5
 IMAGE_REGISTRY=registry:2.5.0
--- a/dockerfiles/socat/Dockerfile
+++ b/dockerfiles/socat/Dockerfile
@@ -1,3 +1,3 @@
-FROM alpine:3.1
-RUN apk --update add socat
+FROM alpine:3.4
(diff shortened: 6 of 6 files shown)


dockerfiles/cli/version/5.0.0-M1/images
```diff
@@ -1,6 +1,6 @@
-IMAGE_SOCAT=codenvy/socat
+IMAGE_SOCAT=codenvy/socat:1.7.3.1
 IMAGE_HAPROXY=codenvy/haproxy:1.5.18
-IMAGE_PUPPET=codenvy/puppet
+IMAGE_PUPPET=codenvy/puppet:4.6.2
```
benoitf: it's strange to see version that is not matching repository version
because codenvy/puppet is defined in this repository

benoitf:
it seems it breaks the release process which is that each file belonging to a single repository need to match the same version

riuvshin:
@benoitf no it does not break anything we tag those images with they're own tags because there is nothing related to codenvy release cycle at all. We needed to extend official images due to our needs that is why we created them under codenvy org.

benoitf:
ok let say your needs evolve. But upstream version is the same
so you can't tag anymore codenvy/puppet:4.6.2 as it's already tagged
if you override the existing tag with a new image, then the new needs can break previous releases of codenvy using the previous image

so IMHO it breaks.

riuvshin:
yes, but there is nothing to evolve, there we have only puppet basically, so the only purpose of changes could be update puppet version and update tag to that version. But even if I will need to update that image but keep same tag I will add `.1` so it will be `4.6.2.1`

benoitf:
I will vote -1 on this.
Same repository = same version
if you need custom version it needs to be out of this repository
so all images published by files on the repository have the same versions

riuvshin:
sorry I don;t understand. I can't use tags of codenvy like 5.0.0 for those images because there will not be changed at all. If you want I can use some other tags for that like 1.0.0, but in that case we will need to have doc with version mappings to know which version of puppet in that image....
I think you confused by something here because it is almost same image as official and it is not related to codenvy at all. in future we may use other image at all. So in fact it is service image which is not related to codenvy directly. it may be even latest because it will just work, but just in case I've tagged it to protect us from any changes.

TylerJewell:
Florent - yes I agree with roman.  

I see the images file as being an index of the versions of the dependency images that this particular version requires.  We definitely tag each version of cli but wouldn't need to tag dependencies.

What would be the benefit of force tagging something with a new version of the contents have not changed?

benoitf:
I'm not confused.
If it's not official there is nothing that say to reuse the same version (I'm ok if it's the only version of a repository. But here as source code is inside codenvy repository, every single of code is tagged with a version. 

Let say I'm a user and I see image named codenvy/puppet:4.6.2 I expect to find the source code of Dockerfile this version by grabbing source code for the selected tag. I may find a codenvy tag named 4.6.2 
I will go in https://github.com/codenvy/codenvy/tree/4.6.2
And I won't find an image named puppet there. So I think it may be more confusing.


>> What would be the benefit of force tagging something with a new version of the contents have not changed?
How can you be sure content has not been changed as the source code of the repository is evolving.

if there were a repository codenvy/socat then I'm sure there is no change and if a change occurs, then the new version is already set and a cron job can do it.
But here, let say I commit changes to https://github.com/codenvy/codenvy/blob/master/dockerfiles/socat/Dockerfile

what will determine the new version of the image ?
it will be manual (so it will fail)

TylerJewell:
Ok - that argument does make some sense.

So the core argument, @riuvshin, is that customers or users who get a list
of images in their local cache may require traceability of the images to
source code. And it is reasonable to think that they will want to go to the
source repository's tag, and then have total traceability of everything
related to codenvy for that particular tagged version.

Tyler Jewell | CEO | tyler@​codenvy.​com | 9​78​.8​84​.53​55

riuvshin:
That argument does not make any sense to me. If user will checkout codenvy 5.0.1
and inside of that tag he will see that we have dockerfile for puppet image which eversion is 4.6.2 
how that user may want to expect version of puppet or any othet dependency to be 5.0.1 ??
it is exactly same as here https://github.com/codenvy/codenvy/blob/master/dockerfiles/cli/version/nightly/images#L2 
i didn;t changes that line. And what that line means that in tag of codenvy 5.0.1 we will use tag of codenvy/haproxy:`1.5.18`
I don't understand how that could be wrong or anything else could be right.... this is how it should be. if that tag of puppet image will be changed due to version update for codenvy `5.0.2` it will be `4.6.3` for exapmle. we don't need tag everything for each codenvy release.

benoitf:
>. If user will checkout codenvy 5.0.1
>and inside of that tag he will see that we have dockerfile for puppet image which eversion is 4.6.2 
User currently download che or codenvy and these binaries include for example a version of tomcat which is not the version of the product. 

> it is exactly same as here https://github.com/codenvy/codenvy/blob/master/dockerfiles/cli/version/nightly/images#L2 
> i didn;t changes that line. 
for haproxy the version is also wrong it should match the codenvy version as it is being produced.

>if that tag of puppet image will be changed due to version update for codenvy 5.0.2 it will be 4.6.3 >for exapmle. we don't need tag everything for each codenvy release.
If you need to include binary (like ssh inside an image, puppet version will stay the same but new version will be tagged when codenvy is released)

> we don't need tag everything for each codenvy release.
It would the first time in the history of codenvy :-) You're tagging everything until now even if parent repositories are not changed

>OR we need to move docker files out of that repo.
as I said different same repository = same version so different repository could allow to use different lifecycle and then different versions

> Again the way how it is done now. you if user check out codenvy tag 5.0.0 it means everything inside EVEN dockerfiles are exactly for 5.0.0 version. we will ever replace overwrite image tags because this is wrong
It's not clear what is wrong ? "we" = ?

riuvshin:
Sorry I don't understand at all... it is same as maven dependency codenvy 5.0.0 have dependency of some lib with other version. What is wrong here?

benoitf:
> Sorry I don't understand at all... it is same as maven dependency codenvy 5.0.0 have dependency of some lib with other version. What is wrong here?

The issue is that you say "it's the same" while it's definitely not.

If I follow "same approach"

- **In pom.xml** : I reference lib with another version and at the end the artifact is versioned with codenvy.version and you publish artifact on maven repository with codenvy version


_so by following "maven approach":_
- **in Dockerfile** you reference a lib (puppet) with another version and at the end the artifact produced (docker image) is built with codenvy.version and we publish to dockerhub with codenvy.version (same pattern than for maven)

while what I see in the pull request is publishing with dependency version

it's like if for maven, for artifacts like https://github.com/eclipse/che-lib/tree/master/swagger you were defining che-swagger-war module with the version of swagger.

TylerJewell:
@riuvshin - I can see both perspectives.  Now, I think there are two kinds of dependencies:
1. Those which are really external dependencies, such as "puppet/puppet".  It is labeled and managed as a dependency.  We do not tag it or version it.

2. Those which are Codenvy managed dependencies, such as "codenvy/socat". Now, we should rarely have to change this.  But, it's got a Codenvy label on it, and then people would expect to trace the source file for it.  So in this case, even though this container is just a wrapper for another dependency, the argument is that we should be tagging this image each time we tag other images.

riuvshin:
what should I do ? leave it as a latest ? I don't understand what is wrong... 
all I need in that PR is use fixed tags to prevent unexpected changes with "latest" which may come when we  don't expect. Should I close PR?

TylerJewell:
I do think that a reasonable compromise is the following:
1. All images should absolutely have their full tags - so your PR is approved to get the version tag on the images.  Please go ahead and merge this PR and then lets deal with the image "policy" in another thread.

For the image policy, I believe:
1. If any Docker image's Dockerfile + entrypoint lives outside of our repository, then of course we would not retag it.

2. But... if the Dockerfile or entrypoint lives in our repository, which some do even though they are just simple wrappers, then that code is getting tagged with each release.  And if the dockerfile or entrypoint is retagged with our code, then we must retag the image we generate from it too.

riuvshin:
OK, the things you are talking about is out of scope here, I merge this as improvement of what we have today and lest's start separate discussion about "retagging" images.

PR #1191, opened 2016-11-19
WIP: CLI Refactoring

TylerJewell:
### What does this PR do?
1. Moves all cmd_() functions into separate files.

2. Dynamically loads all cmd_*.sh files to allow for dynamic loading of the CLI.  You can now add in a new file cmd_*.sh with a set of commands and the CLI will add that automatically.

3. Adds in action, ssh, and mount commands for users.

4. Adds in new "version_compatibility_check" which runs during the cli_init() process.  As part of this, moves the 'codenvy.ver' file out of config/ instead /instance. This was necessary to avoid confusion between the configuration an admin may do to CODENVY_VERSION in the .env file vs. what has actively been installed.

```
First, if_initialized() compare CLI version to CODENVY_VERSION in /config/.env file
      - If they match, good
      - If they don't match and one is nightly, fail
      - If they don't match, then if CLI is older fail with message to get proper CLI
      - If they don't match, then if CLLI is newer fail with message to run upgrade first
Second, if_configed() compare /config/codenvy.env to /instance/codenvy.ver
     - If they match, then continue
     - If they do not match, then if .env is newer, then fail with message to run upgrade first
     - If they do not match, then if .env is older, then fail with message that this is not good 
```

5. Adds in some smart "checks" if you are using the "latest" image tag. If you use this tag, then we will inspect the value of /version/latest.ver inside of the codenvy/cli container.  We then set the CODENVY_IMAGE_VERSION and CODENVY_VERSION to be explicitly this tagged version.  This is essential so that our version checks between the CLI and the environment file and the installed version can match up.

6. Adds in warning message during initialization that if you are installing a nightly version, a message indicates that you cannot upgrade from nightly to a non-nightly image.

7. Adds in upgrade compatibility check.
```
First, compare the CLI image version to what the admin has configured in /config/.env file
      - If they match, nothing to upgrade
      - If they don't match and one is nightly, fail upgrade is not supported for nightly
      - If they don't match, then if CLI is older fail with message that we do not support downgrade
      - If they don't match, then if CLI is newer then good
 Second, compare proposed .env version to already installed version
      - If they match, then ok to upgrade, we will update the ENV file 
      - If they do not match, then if .env is newer, then fail with message that ENV file & install must match before upgrade
      - If they do not match, then if .env is older, then fail with message that ENV file & install must match before upgrade 
```


riuvshin:
@TylerJewell Don't you think that having separate files for function with one line is a huuuuge overhead?
```
cmd_help() {
	usage
}
```

This also makes debugging pure nightmare, you have to know which file to open to check what needed function doing and there is no debugger for bash which can make life easier...
We probably can have possibility to include some custom libs like plugins for CLI, but I strongly believe that it is NOT ok for basic CLI functionality. 
@garagatyi in cbuild tool did like that

TylerJewell:
The CLI has been structured in a way where it doesn't need one *.sh file for each command.  It searches for all files with cmd_*.sh, so we can combine files of like minded functions together liek start / stop / restart / destroy, and backup / restore / offline.

riuvshin:
I understand how it is done and how it is works, but as I said this way it will be much harder to maintain it.

benoitf:
@riuvshin why would it be harder to maintain ?
you prefer to have a single file with thousand of lines of code ?
I don't see where it's harder to maintain small files.
Also with pluggability it's easier to bring new functions

benoitf:
also about sentence 

> "and there is no debugger for bash which can make life easier..."

FYI there is a debugger of bash which is gdb compliant.

TylerJewell:
I was worried about maintainability of thisn approach, too, roman.  However, I spent about 8 hours working on the version compatibility improvements with this structure and it was not bad.  Actually having each command in a different file made it easier to jump around and create a screen layout that was more productive than everything in a single file.

benoitf:
session-debug-bash

mshaposhnik:
How to get rif of old versions of codenvy.sh and cli.sh in my `/usr/local/bin` ?

mshaposhnik:
How to get rid of 
`
./run.sh version
INFO: (codenvy cli): Loading cli...
/repo/dockerfiles/cli/cli.sh: line 150: SCRIPTS_CONTAINER_SOURCE_DIR: unbound variable
`
error ?

riuvshin:
+1

skabashnyuk:
@benoitf can you approve this changes?

TylerJewell:
Max I don't understand the issue.  That code and line is not in this branch.

PR #1195, opened 2016-11-21
Fix dev mode and improve cli code

riuvshin:
Here is a fix for dev mode which was broken by https://github.com/codenvy/codenvy/commit/c4b41b0b83ab857c5dc3cf4ec1ad50a43347d67d
We thought that we don't need `CODENVY_DEVELOPMENT_TOMCAT` var in `codenvy.env`
but we actually need it, we have set this var in CLI "runtime" but this var should be passed to container with puppet and there is two ways how to fix it revert it back to `codenvy.env` OR pass it as env var to container with puppet. I've picked first one because I found references like https://github.com/codenvy/codenvy/blob/master/dockerfiles/cli/cmd_init.sh#L117 here we still doing `sed` to set `CODENVY_DEVELOPMENT_TOMCAT` in `codenvy.env` so that PR fixes that sed command.

Also I;ve improved `cmd_config`  code as I've promised after that https://github.com/codenvy/codenvy/pull/1157 PR

@TylerJewell @garagatyi @benoitf please review guys


TylerJewell:
@riuvshin - I added in improvements for Codenvy version.  What I did:
1. To be really explicit and to do hard checks, during cmd_init(), I write a codenvy.ver.do_not_modify file into the /config folder.  And during cmd_config(), I write a codenvy.ver.do_not_modify file into /instance.

2. During CLI boot, there is a version compatibility check. If for any reason that the /config and /instance versions do not match, then an error is presented to the user with instructions on how to fix this. We could fix this automatically, but a configuration and an installation being different versions should be very rare and a human should check what is going on first.

I would like to recommend that you make two more changes before merging this:
1. I believe that 'codenvy.env' file should be moved into /dockerfiles/init folder, or as part of the manifests folder (which themselves should be moved into '/dockerfiles/init'), so that everything that is delivered onto a host is a sub-directory of the Dockerfile directory for which it is built.

2.  You have two sed commands in the CLI still - 'CODENVY_HOST' and 'CODENVY_SWARM_NODES'.  Can we modify the codenvy/init docker run command to pass these values in?  So the syntax would be:
```
docker_run -v "${CODENVY_HOST_CONFIG}":/copy 
-e "CODENVY_HOST=${CODENVY_HOST}"
-e "CODENVY_SWARM_NODES=${CODENVY_HOST}:23750 
$IMAGE_INIT
```
And then advance the entrypoint of the init() method to handle these parameters properly.  If you think this is a bad idea (because we would just be using sed in the entrypoint), then perhaps we should leave it within the CLI.

dockerfiles/cli/cli.sh
```diff
@@ -441,11 +437,11 @@ get_installed_version() {
   fi
 }
 
-get_envfile_version() {
+get_configed_version() {
```
garagatyi: Maybe we can use configured instead of configed?

dockerfiles/cli/cli.sh
```diff
@@ -464,116 +460,116 @@ less_than() {
...
 
-compare_cli_version_to_envfile_version() {
+compare_cli_version_to_configed_version() {
   IMAGE_VERSION=$(get_image_version)
-  ENV_FILE_VERSION=$(get_envfile_version)
+  CONFIGED_VERSION=$(get_configed_version)
  
 
-  ## First, compare the CLI image version to what the admin has configured in /config/.env file
+  ## First, compare the CLI image version to what version was initialized in /config/*.ver.donotmodify
```
garagatyi: do not modify is used in different formats - with and without underscores

mshaposhnik:
Go fix it and merge it - seems it will reduce versions headache during upgrade :)

riuvshin:
@TylerJewell 
1) done, but due to this we now have 2 codenvy.env files in the end 1 is still in manifests folder and second one as it was in CODENVY_CONFIG/codenvy.env which probably OK because user should touch only second one and first one is more like source it always will be in original state.

2) I would like to keep those sed commands in CLI because it is easier to find them in if needed and it is more clear as for me. Unfortunately we can't get rid of those two sed commands because if we will do same what I've did for others like CODENVY_CONFIG they will be always overwrite what user defined in `codenvy.env`.

dockerfiles/cli/cli.sh
TylerJewell: done - applied it throughout

dockerfiles/cli/cli.sh
TylerJewell: fixed

TylerJewell:
@riuvshin - made my changes and pushed them in.  I think this is ready for merge.

PR #1196, opened 2016-11-21
Cleanup dockerfiles

benoitf:
### What does this PR do?
- cleanup copyright headers
- fix version of alpine (replace latest by 3.4)
- use of name "entrypoint.sh" for entrypoint scripts

PR #1208, opened 2016-11-22
fix haproxy issues, add rsyslog container to collect logs properly

riuvshin:
We've faced with really strange issues when testing codenvy in docker on AWS, haproxy does not work properly and  lose logs, probably we had same on @TylerJewell's windows and added workaround to reload haproxy. This is was due to 2 services in same container haproxy and rsyslog.
So I've used official haproxy image and added codenvy/rsyslog container which will collect logs for us, we will also need this to collect logs from postgres (we don't have any logs from postgres yet)

This need ASAP to finish nightly setup on AWS. We want to start finally test codenvy in docker.

@TylerJewell @benoitf @garagatyi 

it seems @benoitf that after this you will need to update your PR https://github.com/codenvy/codenvy/pull/1196


TylerJewell:
Have you tested this on Windows yet or do you need me to do that test?  Finally, should I make an improvements to the CLI to remove the restart of haproxy to test that behavior on windows?

Also, we will need a docs/ update as the architecture graphic is now missing a container.

riuvshin:
@TylerJewell I would like you to test this branch without workaround in CLI (I don't remember where is it, so feel free to update my PR with removing that workaround  if it works for you.

I will update docs with Eugene.

TylerJewell:
@riuvshin - This change did not fix the Windows issue with haproxy.  I ran two tests with this configuration:
1. Start services with the existing workaround.  Started successfully.
2. Removed the  HAPROXY restart workaround from wait_until_server_is_booted method, and created and started a new system.   The system fully booted, but HAProxy does not exit maintenance mode.

I verified that it was using the right version by building codenvy/cli:nightly locally and then volume mounting the source.

modules/compose/templates/docker-compose-container.yml.erb
```diff
@@ -5,6 +5,19 @@
 version: '2'
 services:
 
+########################
+# RSYSLOG
+########################
+  rsyslog:
+    image: <%= ENV["IMAGE_RSYSLOG"] %>
+    volumes:
+      - '<%= scope.lookupvar('compose::codenvy_folder') -%>/logs/haproxy/:/var/log/haproxy'
```
garagatyi: Why docker file includes 2 volumes and compose file includes only 1?

dockerfiles/rsyslog/Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM alpine:3.4
+
+RUN  apk add --update "rsyslog=8.18.0-r0" \
+  && rm -rf /var/cache/apk/*
+
+EXPOSE 514 514/udp
+
+VOLUME [ "/var/log", "/etc/rsyslog.d" ]
```
garagatyi: Why we need volumes and port exposure in a dockerfile if this will be overridden in compose file?

modules/compose/templates/docker-compose-container.yml.erb
```diff
@@ -17,12 +30,12 @@ services:
 <% end -%>
       - '<%= scope.lookupvar('compose::codenvy_folder') -%>/config/haproxy/haproxy.cfg:/usr/local/etc/haproxy/haproxy.cfg'
       - '<%= scope.lookupvar('compose::codenvy_folder') -%>/config/haproxy/maintenance.html:/etc/haproxy/maintenance.html'
-      - '<%= scope.lookupvar('compose::codenvy_folder') -%>/logs/haproxy/:/var/log/haproxy/'
```
garagatyi: If this container doesn't have volume into rsyslog container how last collects the logs?

dockerfiles/rsyslog/Dockerfile
riuvshin: it may be needed for mounting extra rsyslog configs, today we don't have any but we may have some in future.

riuvshin: will remove it

modules/compose/templates/docker-compose-container.yml.erb
riuvshin: it may be needed for mounting extra rsyslog configs, today we don't have any but we may have some in future

modules/compose/templates/docker-compose-container.yml.erb
riuvshin: haproxy send logs to rsyslog server `log         codenvy-rsyslog:514 local2` we don't need volume for that

benoitf:
> it seems @benoitf that after this you will need to update your PR #1196 
@riuvshin could you explain me why my PR should be blocked by this PR ? all comments are fixed so you should be able to approve it and I could merge my 2 days old PR

riuvshin:
@benoitf the problem is that we need that PR asap, but https://github.com/codenvy/codenvy/pull/1196 I can't approve because it heavily affects on release process. We need to define new rules for release before we merge it.

benoitf:
Ok I've created a dedicated PR for codenvy image versioning
https://github.com/codenvy/codenvy/pull/1212

dockerfiles/rsyslog/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM alpine:3.4
```
benoitf: missing copyright header

riuvshin: done

benoitf: + build and run instructions in comments

dockerfiles/rsyslog/rsyslog.conf
```diff
@@ -9,9 +9,10 @@ module(load="imuxsock")    # local system logging support (e.g. via logger comma
 #module(load="imklog")     # kernel logging support (previously done by rklogd)
 module(load="immark")      # --MARK-- message support
 module(load="imudp")       # UDP listener support
-
+module(load="imtcp")       # TCP listener support
```
benoitf: do we still need udp listener ?

dockerfiles/rsyslog/Dockerfile
riuvshin: done

dockerfiles/rsyslog/rsyslog.conf
riuvshin: yes we need udp, but in some cases we will need tcp also

riuvshin:
@TylerJewell that is super weird that you still have the issue, because now we are using official image without any modifications, maybe it is some docker for windows specific issue. I will try to prepare some test case for you later where we will try to run only that image with compose or even docker run to see how it works. So let's keep workaround for now.

PR #1212, opened 2016-11-23
Define all codenvy/* images to have codenvy versioning

benoitf:
As per https://github.com/codenvy/codenvy/pull/1188#issuecomment-261662827
all codenvy/* images defined in codenvy/codenvy repository should have version of codenvy/codenvy tag.

codenvy/codenvy:nightly inherit from codenvy/jdk:nightly
codenvy/haproxy, codenvy/puppet, codenvy/socat are need to match latest, 5.0.0-M7, etc

Change-Id: I3b5922ed22ac8db5bf041c3b23ce50c3306776de

### DIFF
--- a/dockerfiles/cli/build.sh
+++ b/dockerfiles/cli/build.sh
@@ -11,5 +11,5 @@
 IMAGE_NAME="codenvy/cli"
 . $(cd "$(dirname "$0")"; pwd)/../build.include
 
-init
-build
+init "$@"
+build "$@"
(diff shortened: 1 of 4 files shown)


benoitf:
updated

benoitf:
@TylerJewell @riuvshin

TylerJewell:
Yes, I believe this is importnat, @riuvshin .  If an image is a codenvy/* image with a codenvy entrypoint then we should tag whenever we tag a release.

riuvshin:
@benoitf I've force pushed here, for now we have only 2 images under codenvy with dockerfiles only. Once we will merge this I will need to set up CI jobs for those images.

PR #1217, opened 2016-11-23
Start to make some common files for codenvy/cli and eclipse/che-cli 

benoitf:
- allow to use ansi text with printf
- add ansi text formatting constants
- replace "/codenvy" by $CHE_CONTAINER_ROOT
- replace "codenvy" by $CHE_MINI_PRODUCT_NAME
- replace "Codenvy" by $CHE_PRODUCT_NAME
- initialize usage in its own function (for allowing to override constant)
- add sh script to build the cli image 

- use of eclipse/che-base as base image for common scripts for both eclipse/che-cli and codenvy/cli

### DIFF
--- a/dockerfiles/build.include
+++ b/dockerfiles/build.include
@@ -1,0 +1,34 @@
+#!/bin/sh
+# Copyright (c) 2016 Codenvy, S.A.
+# All rights reserved. This program and the accompanying materials
+# are made available under the terms of the Eclipse Public License v1.0
+# which accompanies this distribution, and is available at
+# http://www.eclipse.org/legal/epl-v10.html
+#
+# Contributors:
+#   Florent Benoit - Initial Implementation
+
+init() {
+  BLUE='\033[1;34m'
+  GREEN='\033[0;32m'
+  RED='\033[0;31m'
+  NC='\033[0m'
+  if [[ $# -eq 0 ]] ; then
+    TAG="nightly"
+    echo "No tag provided, using nightly as default"
+  else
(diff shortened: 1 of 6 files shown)


benoitf:
pr updated

TylerJewell:
These are all approved changes

benoitf:
@TylerJewell

riuvshin:
why this merged with only one approve? 
why  CLI using `FROM eclipse/che-base:nightly` ? how you guys going to manage it with other versions?

benoitf:
When che is tagged, che images are tagged
Codenvy is using che so codenvy cli image will use che-base tagged image as eclipse/che-cli

riuvshin:
I mean how you are going to keep nightly CLI always using actual che-base ? 
those changes affect processes like release and CI 
CI: to merge this PR we have to set up trigger on CI to rebuild CLI if che-base was changed ( I think due to this CLI does not work on my side at all.)
Release: that affected release process because we need to set those tags here, and there is no any info about that, this is super important because If I will do tag and will not set tags before we won't be able to build images from that tag. So I need strategy here how I should release this what are the steps before I start and etc.

riuvshin:
I've added dependency on CI, now each build of `che-base:nightly` will trigger `cli:nightly`
I also added docker pull `eclipse/che-base:nightly` in CLI job to make sure it is actual if they will run on separate CI slaves.

But for releases I don't see whole picture I think we have to write somewhere docker images dependencies and keep it always up to date

PR #1218, opened 2016-11-23
print compose logs to console in dev mode, add license mount, sync conf

riuvshin:
1) in dev mode we have to print compose output to console to make devs life easier
2) added missing mount for license file to not loss it after restart
3) sync script with deployment

@garagatyi @TylerJewell @benoitf

### DIFF
--- a/dockerfiles/cli/cmd_start.sh
+++ b/dockerfiles/cli/cmd_start.sh
@@ -53,8 +53,13 @@ cmd_start() {
   # Start Codenvy
   # Note bug in docker requires relative path, not absolute path to compose file
   info "start" "Starting containers..."
-  log "docker_compose --file=\"${REFERENCE_CONTAINER_COMPOSE_FILE}\" -p=$CHE_MINI_PRODUCT_NAME up -d >> \"${LOGS}\" 2>&1"
-  docker_compose --file="${REFERENCE_CONTAINER_COMPOSE_FILE}" \
-                 -p=$CHE_MINI_PRODUCT_NAME up -d >> "${LOGS}" 2>&1
+  COMPOSE_UP_COMMAND="docker_compose --file=\"${REFERENCE_CONTAINER_COMPOSE_FILE}\" -p=\"${CHE_MINI_PRODUCT_NAME}\" up -d"
+
+  if [ "${CODENVY_DEVELOPMENT_MODE}" != "on" ]; then
+    COMPOSE_UP_COMMAND+=" >> \"${LOGS}\" 2>&1"
+  fi
+
+  log ${COMPOSE_UP_COMMAND}
+  eval ${COMPOSE_UP_COMMAND}
   check_if_booted
 }
--- a/modules/codenvy/templates/cleanUpWorkspaceStorage.sh.erb
+++ b/modules/codenvy/templates/cleanUpWorkspaceStorage.sh.erb
@@ -1,6 +1,16 @@
 #!/bin/sh
 
 DST_FOLDER=${1}
-CLEANUP_REGEX='^(/opt/codenvy-data/fs/([0-9a-f]{2}/){3}workspace[0-9a-z]+)$'
+CLEANUP_REGEX='^(.*/codenvy-data/fs/([0-9a-f]{2}/){3}workspace[0-9a-z]+)$'
 
-[[ ${DST_FOLDER} =~ ${CLEANUP_REGEX} ]] && sudo rm -rf "${DST_FOLDER}"
+hasSudoCommand() {
+  hash sudo 2>/dev/null && return 0 || return 1
+}
+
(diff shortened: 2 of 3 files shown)


TylerJewell:
+1 - this is a great idea.

dockerfiles/cli/cmd_start.sh
```diff
@@ -53,8 +53,13 @@ cmd_start() {
...
-  docker_compose --file="${REFERENCE_CONTAINER_COMPOSE_FILE}" \
-                 -p=$CHE_MINI_PRODUCT_NAME up -d >> "${LOGS}" 2>&1
+  COMPOSE_UP_COMMAND="docker_compose --file=\"${REFERENCE_CONTAINER_COMPOSE_FILE}\" -p=\"${CHE_MINI_PRODUCT_NAME}\" up -d"
+
+  if [ "${CODENVY_DEVELOPMENT_MODE}" != "on" ]; then
+    COMPOSE_UP_COMMAND+=" >> \"${LOGS}\" 2>&1"
+  fi
+
+  log ${COMPOSE_UP_COMMAND}
+  eval ${COMPOSE_UP_COMMAND}
```
benoitf: should we use at some point cli_eval/silent_eval functions instead of eval with some flags

I'm not sure in dev mode we need to see this log everytime on the console. Looks like each developer has its view on this, so I would prefer a toggle flag

modules/codenvy/templates/cleanUpWorkspaceStorage.sh.erb
```diff
@@ -1,6 +1,16 @@
 #!/bin/sh
 
 DST_FOLDER=${1}
-CLEANUP_REGEX='^(/opt/codenvy-data/fs/([0-9a-f]{2}/){3}workspace[0-9a-z]+)$'
```
garagatyi: Why did you change this file?

modules/compose/templates/docker-compose-container.yml.erb
```diff
@@ -146,6 +146,7 @@ services:
        - '<%= scope.lookupvar('compose::codenvy_folder') -%>/logs/codenvy/:/opt/codenvy-data/logs/'
        - '<%= scope.lookupvar('compose::codenvy_folder') -%>/logs/codenvy/che-machines-logs/:/opt/codenvy-data/che-machines-logs/'
        - '<%= scope.lookupvar('compose::codenvy_folder') -%>/data/codenvy/fs/:/opt/codenvy-data/fs/'
+       - '<%= scope.lookupvar('compose::codenvy_folder') -%>/data/codenvy/license/:/opt/codenvy-data/license/'
```
garagatyi: Why did you change this file?

modules/codenvy/templates/cleanUpWorkspaceStorage.sh.erb
riuvshin: bo bikoz

modules/compose/templates/docker-compose-container.yml.erb
riuvshin: without mount /license if license is generated it will be gone after codenvy restart because it is not mounter

dockerfiles/cli/cmd_start.sh
riuvshin: maybe we can improve it like this, my only motivation here is that due to frequent changes in CLI guys fail to start codenvy and they come to me and saying it didn't work. if they will see error message on console it will be much easier for them to figure out where is a problem.

modules/codenvy/templates/cleanUpWorkspaceStorage.sh.erb
```diff
@@ -1,6 +1,16 @@
...
+CLEANUP_REGEX='^(.*/codenvy-data/fs/([0-9a-f]{2}/){3}workspace[0-9a-z]+)$'
 
-[[ ${DST_FOLDER} =~ ${CLEANUP_REGEX} ]] && sudo rm -rf "${DST_FOLDER}"
+hasSudoCommand() {
+  hash sudo 2>/dev/null && return 0 || return 1
+}
+
+if echo "${DST_FOLDER}" | grep -q -E "${CLEANUP_REGEX}"; then
```
riuvshin: that script written in a way to be compatible with old codenvy where we have sudo. So for now we want to keep it like that and once codenvy will be only in docker we will clean this up.

dockerfiles/cli/cmd_start.sh
benoitf: I think it would great to have it displayed on console if there are errors

riuvshin: +1 but I don't see now how to do that, I think it is more about refactoring in a general way for managing that

PR #1221, opened 2016-11-24
WIP: CLI Refactoring - Consolidate codenvy/init + puppet containers

TylerJewell:
### What does this PR do?
As a general rule, we were finding that it was possible for a user to have configuration in /config that would wander from the version that was installed. This is really bad.  To solve this, we looked at how to let the user only have a single configuration parameter - codenvy.env, and then all other files go into a single instance folder.

The new folder structure is:
```
/codenvy.env
/docs
/instance
/instance/codenvy.ver.do_not_modify
/instance/docker-compose-*
/instance/config
/instance/data
/instance/dev
/instance/logs
```

A number of changes have been made:
1. The codenvy/init container now inherits from puppet/agent 
2. This consolidates two containers down to one.
3. /manifests and /modules were moved into /dockerfiles/init folder
4. Before, we either required users to mount a single :/codenvy OR all three of :/codenvy/instance, :/codenvy/config, and :/codenvy/backup. Now, we can simplify this, and the user must ALWAYS mount :/codenvy, and can OPTIONALLY also mount :/codenvy/instance and :/codenvy/backup only if they want to override the default location of these items.
5. The startup sequence was simplfied to provide much smarter help to the end user.
6. Introduced CHE_FORMAL_PRODUCT_NAME variable to print out "Codenvy" in certain locations instead of "codenvy" or "CODENVY".
7. Eliminates unnecessary checks in version_compatibility function where we had to check the configured version to the installed version. Now the only check is CLI to installed version.
8. Consolidated cmd_start(), cmd_restart(), and cmd_stop() into a single *.sh file. Also updated the CLI_parse to have an explicit set of available commands.
9. Fixes bug with cmd_destroy(), where if you were in development mode, then an extra /dev folder is copied into /instance, and it needs to be indivdiually removed.
10. Adds in check that if you are using the nightly image, if your local cached image is >1 day older than the one on DockerHub, will print out a warning message.


benoitf:
- about item 6
> Introduced CHE_FORMAL_PRODUCT_NAME variable to print out "Codenvy" in certain locations instead of "codenvy" or "CODENVY".

I don't think we need another variable here. If uppercase is needed then CHE_PRODUCT_NAME can be converted to uppercase. (--> this avoid the work in all CLIs to define several constants only for uppercase)

- on the move of functions to cmd_start file :
I don't see why docker helper commands have been moved to cmd_start.sh
functions like container_exist_by_name are not only used inside cmd_start  but by cmd_action, cmd_list-nodes

dockerfiles/cli/cli.sh
```diff
@@ -35,16 +35,16 @@ cli_init() {
   CODENVY_HOST=${CODENVY_HOST:-${DEFAULT_CODENVY_HOST}}
 
   if [[ "${CODENVY_HOST}" = "" ]]; then
-    info "Welcome to Codenvy!"
+    info "Welcome to $CHE_FORMAL_PRODUCT_NAME!"
     info ""
     info "We did not auto-detect a valid HOST or IP address."
-    info "Pass CODENVY_HOST with your hostname or IP address."
+    info "Pass ${CHE_PRODUCT}_HOST with your hostname or IP address."
```
benoitf: I don't see where CHE_PRODUCT is defined

TylerJewell: good catch.  Fixed.

On the other points:
1. Yes, I agree we should only define a single name "Codenvy" and then convert that to ALL CAPS, all lower.  I haven't done that in this update, but I agree this is probably a good improvement.

2. I can move the functions out of cmd_start().  I was thinking that if there were functions for reuse that were only used in a single command, that they would be better suited in the file that used them.  But I get your point and placed them back into the base location.

benoitf: About #1 I have not seen usage of CHE_PRODUCT_NAME in the code

TylerJewell: It is defined in the codenvy.sh file

benoitf: Yes I means until now, we were not using it so maybe we can merge variables. Also in case of che  CHE_PRODUCT_NAME is "ECLIPSE CHE" so it will not work for property name ( uppercase of mini product name will)

TylerJewell:
@benoitf - I have added in all of the items that you have requested, other than the smart logic around the three forms of product names.  I think we can defer that to the refactoring you are doing.

riuvshin:
@TylerJewell help command does not work :)

riuvshin:
it works only if you didn't pass any commands to the cli

riuvshin:
@TylerJewell I've added missing help command and tested both dev and prod modes everything works! 😎  so we can merge this

Issue #1226, opened 2016-11-25
Codenvy in Docker: Workspace agents are not stopped during codenvy stop or restart


skabashnyuk:
Workspace agents are not stopped during codenvy stop or restart. I assume we didn't call
```catalina.sh stop``` . I think we may consider to reuse code from che https://github.com/eclipse/che/blob/master/assembly/assembly-main/src/assembly/bin/docker.sh#L246-L252

**Reproduction Steps:**

1. start codenvy in docker
2. start workspace
3. stop codenvy.

**Expected behavior:**
no docker containers are running

**Observed behavior:**

workspace agent container still running

PR #1228, opened 2016-11-25
gracefully stop ws master with SIGTERM, fix typos, add stop timeout 3min

riuvshin:
needed for https://github.com/codenvy/codenvy/issues/1226
1) gracefully stop ws master (which also invoke stopping running wokrspaces)
2) add timeout 3min for compose stop action (we have same for puppet based codenvy this needed to make sure that all workspaces will be stopped if we have many of them running)
3) fix typos

@skabashnyuk @garagatyi @TylerJewell


dockerfiles/alpine_jdk/Dockerfile
```diff
@@ -82,6 +82,5 @@ RUN apk upgrade --update && \
             /opt/codenvy-data/che-machines /opt/codenvy-data/che-machines-logs \
             /opt/codenvy-data/conf /opt/codenvy-data/conf/logback \
             /opt/codenvy-data/license
-ADD entrypoint.sh /opt/entrypoint.sh
-RUN chmod a+x /opt/entrypoint.sh
-ENTRYPOINT ["/opt/entrypoint.sh"]
+
+RUN postfix -c /etc/postfix start
```
benoitf: it looks strange to have this as a run command
it looks like a CMD or entrypoint

dockerfiles/cli/cmd_start.sh
```diff
@@ -74,9 +74,9 @@ cmd_stop() {
 
   info "stop" "Stopping containers..."
   if is_initialized; then
-    log "docker_compose --file=\"${REFERENCE_CONTAINER_COMPOSE_FILE}\" -p=$CHE_MINI_PRODUCT_NAME stop >> \"${LOGS}\" 2>&1 || true"
+    log "docker_compose --file=\"${REFERENCE_CONTAINER_COMPOSE_FILE}\" -p=$CHE_MINI_PRODUCT_NAME stop -t 180 >> \"${LOGS}\" 2>&1 || true"
     docker_compose --file="${REFERENCE_CONTAINER_COMPOSE_FILE}" \
```
benoitf: 180 should be a variable

dockerfiles/codenvy/Dockerfile
```diff
@@ -22,4 +22,6 @@ RUN set -x \
     > /usr/bin/docker; chmod +x /usr/bin/docker
 ADD assembly/onpremises-ide-packaging-tomcat-codenvy-allinone/target/onpremises-ide-packaging-tomcat-codenvy-allinone-*.zip /
 RUN unzip -q onpremises-ide-packaging-tomcat-codenvy-allinone-*.zip -d /opt/codenvy-tomcat && rm onpremises-ide-packaging-tomcat-codenvy-allinone-*.zip
-CMD /opt/codenvy-tomcat/bin/catalina.sh ${JPDA} run
+COPY dockerfiles/codenvy/entrypoint.sh /
+RUN chmod +x entrypoint.sh
```
benoitf: why do we need to perform a chmod and file permission 
31  dockerfiles/alpine_jdk/entrypoint.sh → dockerfiles/codenvy/entrypoint.sh 100755 → 100644
is being changed

dockerfiles/alpine_jdk/Dockerfile
riuvshin: I agree, It was in entrypoint  here but I've moved it to codenvy. 
I don't this we should have this at all here... because if it will be in entrypoint it will be overridded by codenvy entrypoint because codenvy doing FROM this one.
and even if it will work it does not match single process pattern.

So I will remove it from jdk image. We agreed with @TylerJewell that if user wants emails he should configure smtp.

dockerfiles/codenvy/Dockerfile
riuvshin: agree, I've just used how it was in jdk image, will update

dockerfiles/alpine_jdk/Dockerfile
riuvshin: Yeah I get this. But im thinking about removing this at all.

dockerfiles/codenvy/Dockerfile
riuvshin: I just did similar to init or cli

dockerfiles/alpine_jdk/Dockerfile
riuvshin: @benoitf for now I will keep it, fixed to CMD

dockerfiles/cli/cmd_start.sh
riuvshin: Done

PR #1236, opened 2016-11-28
ENTERPRISE-25: improve license message service

olexii4:
### What does this PR do?
Improve license message service

selection_038


dashboard/src/app/onprem/license-messages/license-limit-dialog/license-limit-dialog.controller.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  }
+
+  /**
+   * It will hide the dialog box.
+   */
+  hide(): void {
+    let now: Date = new Date();
+    this.$cookies.put(this.key, 'true', {
+      expires: new Date(now.getFullYear() + 1, now.getMonth(), now.getDate())// set the expiration to 12 months
```
dmytro-ndp: Please, set period more or equals to 10 years.

dmytro-ndp:
@olexii4: please, add displaying error message if admin riches user license limit in time of adding new user at the _Users_ page.
Also remove flag of displaying error message after removing user at the _Users_ page if **legality**    service doesn't return issue of succeeding user limit.

dmytro-ndp:
@olexii4: it seems _Dashboard_ doesn't remove flag **USER_LICENSE_HAS_REACHED_ITS_LIMIT** after adding valid _license_ which set user limit to value which > actual user number.

olexii4:
done

PR #1242, opened 2016-11-28
Auto snapshots configurability

riuvshin:
* allow to enable / disable auto snapshots and restore
* allow to configure custom registry for workspace snapshots
@benoitf @TylerJewell @garagatyi 

@TylerJewell this may require docs updates, can you please help me with this ?


riuvshin:
thanks @TylerJewell

PR #1243, opened 2016-11-28
use codenvy.env from dev repo in dev mode

riuvshin:
in dev mode we should take codenvy.env from dev repo not from the image
@benoitf @TylerJewell @garagatyi


TylerJewell:
If this is copying the repo .env file during init then this is ok.   But if this sourcing the env file for every config from the repo then that is not good.

Also, florent I think has refactored the Che base now.  Should this pr be part of that at this point?

riuvshin:
here I just volume mount codenvy.env file from repo if DEV MODE is ON because if you have changed it in dev mode you want that it  will be applied. 
here is how we generate this file onto host https://github.com/codenvy/codenvy/blob/master/dockerfiles/init/entrypoint.sh#L13
it just copied during `cmd_init` so my changes just put codenvy.env from repo to `/etc/puppet/manifests/codenvy.env` inside of an `init` image.

Issue #1250, opened 2016-11-30
Codenvy in docker: Enable postgresql logs


skabashnyuk:
I would like to see postgresql logs when running codenvy in docker. 
Proposed changes from current config
            log_min_duration_statement = 0 <- possible need to be configurable 
            log_line_prefix = 'user=%u,db=%d,app=%aclient=%h '
            log_checkpoints = on
            log_connections = on
            log_disconnections = on
            log_lock_waits = on
            log_temp_files = 0
            log_autovacuum_min_duration = 0
            log_error_verbosity = default

riuvshin:
Done

PR #1261, opened 2016-12-01
enable postgre logs

riuvshin:
needed for https://github.com/codenvy/codenvy/issues/1250

@skabashnyuk @garagatyi


dockerfiles/init/modules/rsyslog/templates/postgre.conf.erb
```diff
@@ -0,0 +1,3 @@
+$FileCreateMode 0644
```
benoitf: name of the file should be postgresql.conf.erb maybe ?

and /var/logs/postgresql/postgresql.log ?

PR #1265, opened 2016-12-01
fix names, sync compose files

riuvshin:
@skabashnyuk


TylerJewell:
We need to possibly hold off on any more changes until florent has finished the cli refactoring as it gets harder to synchronize now.

riuvshin:
it was not related to CLI at all, it was puppet configs

PR #1285, opened 2016-12-05
CHE-3199: add cpu limits configuration

garagatyi:
### What does this PR do?
Change Codenvy in accordance with changes in Che.


garagatyi:
@skabashnyuk @riuvshin Please review

dockerfiles/init/manifests/codenvy.env
```diff
@@ -147,6 +147,28 @@
...
+#     Value can start from slash which means it will be absolute or without slash for relative.
+#     Docker will create cgroup if it is missing.
+#CODENVY_DOCKER_PARENT_CGROUP=NULL
+
+# CPU cores limit
+#     Configuration of CPU cores used for running docker containers in workspaces.
+#CODENVY_DOCKER_CPUSET_CPUS=NULL
+
+# CPU consumption limit
+# Next 2 values set limits on CPU consumption by containers of started workspaces.
```
TylerJewell: Not formatted like others.

TylerJewell:
From a documentation + configuration point of view, I provided these Che comments, and then we should have something similar added for Codenvy, and then we would be ok.

https://github.com/eclipse/che/pull/3278#issuecomment-264949375

dockerfiles/init/manifests/codenvy.env
garagatyi: ok

garagatyi:
@TylerJewell I've finished with documentation

PR #1286, opened 2016-12-05
add depends_on for rsyslog, sync compose files

riuvshin:
@skabashnyuk @garagatyi


riuvshin:
@benoitf puppet `file` func does not have such option... but maybe I can do some expression inside of a template. I agree  with you that we have to create those files from single source OR maybe remove it. I will investigate this.

PR #1289, opened 2016-12-06
single template for compose files

riuvshin:
@benoitf @garagatyi @skabashnyuk


benoitf:
this change should be applied on
https://github.com/eclipse/che/blob/master/dockerfiles/init/modules/compose/manifests/init.pp

as well for https://github.com/eclipse/che/issues/2977

riuvshin:
@benoitf yes I will introduce same PR for CHE

Issue #1282, opened 2016-12-05
Codenvy in container doesn't check availability of port 8000 in development mode


garagatyi:
Codenvy in container doesn't check availability of port 8000 in development mode. 
So it fails to start instead of checking port and not trying to start.

**Reproduction Steps:**

Start Che in development mode. 
Try to start Codenvy in development mode.

**Expected behavior:**

Codenvy says that port is occupied and start can't be finished.

**Observed behavior:**

Codenvy tries to start and fails with error: 
```
INFO: (codenvy start): Starting containers...
Creating codenvy_nginx_1
Creating codenvy_registry_1
Creating codenvy_rsyslog_1
Creating codenvy_socat_1
Creating codenvy_postgres_1
Creating codenvy_swarm_1
Creating codenvy_haproxy_1
Creating codenvy_codenvy_1

ERROR: for codenvy  Cannot start service codenvy: driver failed programming external connectivity on endpoint codenvy_codenvy_1 (040755d14d7c82d9e527117575c6c717033f1e5bc884491a9dd58cc62ee77416): Error starting u
serland proxy: listen tcp 0.0.0.0:8000: bind: address already in use
Encountered errors while bringing up the project.
```

**Codenvy version:** 5.0.0-M8-Snapshot
**OS and version:** Ubuntu 16.04

**Additional information:**

Problem can be reliably reproduced: [Yes]

TylerJewell:
I'll take this after everything is merged.  Will combine this with improvement checks for -it on the command line.

TylerJewell:
@garagatyi - I have a pull request pending.  You will need to rebuild the /init docker image and the /cli docker image to test it.

PR #1294, opened 2016-12-06
CLI Improvements

TylerJewell:
### What does this PR do?
1. Replaces "/codenvy" volume mount with "/data" to be consistent with what Che uses. This way different tools can make assumptions about the basic volume mount.

2. Adds in a check that if you are in dev mode, then will pull the value of CODENVY_DEBUG_PORT from the codenvy.env file and checks to see if that port is available.

3. Updates the usage() method to be friendlier.

4. Adds in the ability to detect proxy configuration from Docker daemon and then automatically set that in the codenvy.env file during the init() phase.

https://github.com/codenvy/codenvy/issues/1282


dockerfiles/cli/scripts/cli.sh
```diff
@@ -91,14 +91,39 @@ cmd_init_reinit_pre_action() {
 }
 
 cmd_start_check_ports() {
+
+  # If dev mode is on, then we also need to check the debug port set by the user for availability
+  if [ "${CHE_DEVELOPMENT_MODE}" = "on" ]; then
+    USER_DEBUG_PORT=$(docker_run --env-file="${REFERENCE_CONTAINER_ENVIRONMENT_FILE}" alpine sh -c 'echo $CODENVY_DEBUG_PORT')
```
garagatyi: Just a hint on bash syntax:
If you use `local ` term before declaring variable it will prevent exposing that variable to other functions. It can be much safer way to use local variables that are not intended to be exposed.
Example:
local USER_DEBUG_PORT=....

garagatyi: You can use that syntax in your further scripts to make them more fail-safe

dockerfiles/cli/scripts/entrypoint.sh
```diff
@@ -22,54 +22,45 @@ CHE_SERVER_CONTAINER_NAME="${CHE_MINI_PRODUCT_NAME}_${CHE_MINI_PRODUCT_NAME}_1"
...
-Variables:
-    CODENVY_HOST                         IP address or hostname where ${CHE_FORMAL_PRODUCT_NAME} will serve its users
-    CLI_DEBUG                            Default=false. Prints stack trace during execution
-    CLI_INFO                             Default=true. Prints out INFO messages to standard out
-    CLI_WARN                             Default=true. Prints WARN messages to standard out
-    CLI_LOG                              Default=true. Prints messages to cli.log file
+OPTIONAL DOCKER PARAMETERS:
+  -e CODENVY_HOST=<YOUR_HOST>          IP address or hostname where ${CHE_FORMAL_PRODUCT_NAME} will serve its users
+  -v <LOCAL_PATH>:/data/instance       Where instance, user, log data will be saved
+  -v <LOCAL_PATH>:/data/backup         Where backup files will be saved
```
benoitf: /data should be replaced by ${CHE_CONTAINER_ROOT}  if the two previous lines

dockerfiles/cli/scripts/entrypoint.sh
```diff
@@ -12,7 +12,7 @@
 CHE_PRODUCT_NAME="CODENVY"
 CHE_MINI_PRODUCT_NAME="codenvy"
 CHE_FORMAL_PRODUCT_NAME="Codenvy"
-CHE_CONTAINER_ROOT="/codenvy"
+CHE_CONTAINER_ROOT="/data"
```
TylerJewell: Fixed

TylerJewell:
Do not merge this issue until the matching Che issue is merged:
https://github.com/eclipse/che/pull/3301

garagatyi:
I tried CLI from this branch and it fails correctly with good error message. Good job!

garagatyi:
But I want to mention that for some reason I could not build images using existing build scripts so had to do that manually.

benoitf:
@garagatyi yes @tolusha told me as well there is an issue with "sh/dash" I will fix scripts

dockerfiles/cli/scripts/entrypoint.sh
mshaposhnik: This change broke down the start.
```
INFO: (codenvy start): Starting containers...
Couldn't find env file: /codenvy/instance/config/postgres/postgres.env
```

dockerfiles/cli/scripts/cli.sh
```diff
@@ -86,19 +86,65 @@ server_is_booted_extra_check() {
...
+  fi
+  if [[ ! ${HTTPS_PROXY} = "" ]]; then
+    sed -i'.bak' "s|#${CHE_PRODUCT_NAME}_HTTPS_PROXY_FOR_${CHE_PRODUCT_NAME}=.*|${CHE_PRODUCT_NAME}_HTTPS_PROXY_FOR_${CHE_PRODUCT_NAME}=${HTTPS_PROXY}|" "${REFERENCE_CONTAINER_ENVIRONMENT_FILE}"
+    sed -i'.bak' "s|#${CHE_PRODUCT_NAME}_HTTPS_PROXY_FOR_${CHE_PRODUCT_NAME}_WORKSPACES=.*|${CHE_PRODUCT_NAME}_HTTPS_PROXY_FOR_${CHE_PRODUCT_NAME}_WORKSPACES=${HTTPS_PROXY}|" "${REFERENCE_CONTAINER_ENVIRONMENT_FILE}"
+  fi
+  if [[ ! ${NO_PROXY} = "" ]]; then
+    #
+    # NOTE --- Notice that if no proxy is set, we must append 'codenvy-swarm' to this for docker networking
+    #
+    sed -i'.bak' "s|#${CHE_PRODUCT_NAME}_NO_PROXY_FOR_${CHE_PRODUCT_NAME}=.*|${CHE_PRODUCT_NAME}_NO_PROXY_FOR_${CHE_PRODUCT_NAME}=${NO_PROXY},codenvy-swarm|" "${REFERENCE_CONTAINER_ENVIRONMENT_FILE}"
```
mshaposhnik: If we add the `codenvy-swarm` host at the end of the `NO_PROXY` string, it must be followed by comma (because of non-stansdart hostname), or insert it into beginning of the string or between IP's. Examples:
Not working:
```
CODENVY_NO_PROXY_FOR_CODENVY=localhost,127.0.0.1,172.17.0.1,codenvy-swarm
```
Working:
```
CODENVY_NO_PROXY_FOR_CODENVY=localhost,127.0.0.1,172.17.0.1,codenvy-swarm,
```
or
```
CODENVY_NO_PROXY_FOR_CODENVY=localhost,127.0.0.1,codenvy-swarm,172.17.0.1
```
or
```
CODENVY_NO_PROXY_FOR_CODENVY=codenvy-swarm,localhost,127.0.0.1,172.17.0.1
```

benoitf: @mshaposhnik do you have rebuild the base image first ?

dockerfiles/cli/scripts/entrypoint.sh
TylerJewell: Found the issue and fixed it in the /init templates. The /init image has to be rebuilt with these changes.

dockerfiles/cli/scripts/cli.sh
TylerJewell: It was a hard coded value in the templates that are now fixed.  I have updated the sed commands to always add a comma after the codenvy-swarm entry.

PR #1298, opened 2016-12-07
RELEASE: set tag versions

riuvshin:
@benoitf for now I keep versions of images we own like rsyslog and socat as is. after release I will start work on that, this is related to issue where we should use tagged with current version all owned by us images ( I can't find that issue can you please help me find this.) 
@skabashnyuk


benoitf:
@riuvshin codenvy/codenvy#1212 it's a pull request

PR #1309, opened 2016-12-08
fix no proxy issue

riuvshin:
always add pipe in the end of no proxy for JVM


TylerJewell:
Please do matching for Che too.

riuvshin:
sure https://github.com/eclipse/che/pull/3325/files

PR #1318, opened 2016-12-12
Fix start of terminal agent in Codenvy in Docker

garagatyi:
### What does this PR do?
Fixes addition of terminal agent into assembly which fixes start of terminal agent


garagatyi:
@skabashnyuk @evoevodin Please review

Issue #1153, opened 2016-11-15
Distribute agents from separate container, not from API


garagatyi:
Now we host agent binaries in tomcat in API container. It is not responsibility of API and should be separated.
We can use any server that hosts files.

PR #1346, opened 2016-12-14
Distribute agents from separate container, not from API

tolusha:
### What does this PR do?
Distribute agents from separate container, not from API

https://github.com/codenvy/codenvy/issues/1153

### Previous Behavior
Agents were being downloaded from master.

### New Behavior
Agents will be downloaded from dedicated docker container.

### Tests written?
Manual testing


benoitf:
Is that 9000 port is a port that clients need to connect ? (if yes we have to check the port locally)
https://github.com/codenvy/codenvy/blob/master/dockerfiles/cli/scripts/cli.sh#L130-L135
it seems we should check the port in dev mode

To provide the agents, in the PR, you're using alpine + lighthttpd
as we're already requiring nginx image, shouldn't it be better to use nginx image ? so we don't require/need a new image

dockerfiles/agents/Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+#    docker run -p 9000:9000 codenvy/agents
+
+FROM alpine:3.4
+
+RUN apk update \
+    && apk add lighttpd \
+    && echo "server.port = 9000" >> /etc/lighttpd/lighttpd.conf \
+    && ln -s /home/user/che /var/www/localhost/htdocs/agent-binaries
+
+RUN addgroup -S user && adduser -S -g user user
```
riuvshin: do we really need user here?

dockerfiles/agents/Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+
+COPY dependency /home/user/che/
+
+EXPOSE 9000
+
+CMD ["lighttpd", "-f", "/etc/lighttpd/lighttpd.conf", "-D"]
```
riuvshin: where is lighttpd.conf ? maybe we need generate that config with puppet?

riuvshin:
@benoitf no 9000 port is internal all connections will be to haproxy like <CHE_API>/agent-binaries

benoitf:
@riuvshin 
```
+<% if scope.lookupvar('compose::env') != 'production' -%>
 +    ports:
 +      - '9000:9000'
 +<% end -%>
```

riuvshin:
yes this is for dev mode, I've asked to add this because I thought it will be tomcat and we will need be able to debug it in dev mode. so we have to remove that part.
that container with agents was added as haproxy backend so all clients will connect to haproxy

dockerfiles/agents/Dockerfile
tolusha: Agree. should be managed by puppet

tolusha:
@benoitf 
The task is about to add a dedicated container not to reuse existed one.

benoitf:
@tolusha yes I was talking about image, not the container

dockerfiles/init/modules/compose/templates/docker-compose.yml.erb
```diff
@@ -158,13 +159,32 @@ services:
...
 ########################
+# AGENTS
+########################
+  agents:
+    image: <%= ENV["IMAGE_AGENTS"] %>
+    volumes:
+      - '<%= scope.lookupvar('compose::codenvy_folder') -%>/config/lighttpd/lighttpd.conf:/etc/lighttpd/lighttpd.conf'
+<% if scope.lookupvar('compose::env') != 'production' -%>
+    ports:
+      - '9000:9000'
```
benoitf: if this code is still there
then
https://github.com/codenvy/codenvy/blob/master/dockerfiles/cli/scripts/cli.sh#L130-L135 need to check 900 port

if the local listener is dropped, then no need to check this port

dockerfiles/agents/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+#   docker build -t codenvy/agents .
+#
+# use:
+#    docker run -p 9000:9000 codenvy/agents
+
+FROM alpine:3.4
+
+RUN apk update \
+    && apk add lighttpd \
+    && ln -s /data /var/www/localhost/htdocs/agent-binaries
```
benoitf: to reduce size of the image, cache of apk is usually dropped as /tmp folder : 
`rm -rf /var/cache/apk/*` and sometimes `/tmp` folder
`

dockerfiles/agents/build.sh
```diff
@@ -0,0 +1,39 @@
...
+TERMINAL_AGENTS=("org.eclipse.che:exec-agent:${POM_VERSION}:tar.gz:linux_amd64"
+                 "org.eclipse.che:exec-agent:${POM_VERSION}:tar.gz:linux_arm7");
+
+for artifact in ${TERMINAL_AGENTS[@]}; do
+    prefix=$(echo ${artifact} | cut -d ':' -f 5)
+    mvn dependency:copy -Dartifact=${artifact} -DoutputDirectory=${DEPENDENCY_DIR}/${prefix}/terminal -Dmdep.stripVersion=true
+    mv ${DEPENDENCY_DIR}/${prefix}/terminal/exec-agent-${prefix}.tar.gz ${DEPENDENCY_DIR}/${prefix}/terminal/websocket-terminal-${prefix}.tar.gz
+done
+
+mvn dependency:copy -Dartifact=com.codenvy.onpremises:onpremises-ide-packaging-tomcat-ext-server:${POM_VERSION}:tar.gz \
```
benoitf: could we use a maven pom + assembly.xml file to copy all the required dependencies ?
then we could only call a single maven command

dockerfiles/init/modules/lighttpd/templates/lighttpd.conf.erb
```diff
@@ -0,0 +1,323 @@
+###############################################################################
+# Default lighttpd.conf for Gentoo.
```
benoitf: I'm not sure comment is accurate as we're based on alpine

dockerfiles/agents/build.sh
```diff
@@ -0,0 +1,39 @@
...
+
+# prepare
+DIR=$(cd "$(dirname "$0")"; pwd)
+DEPENDENCY_DIR="${DIR}/dependency"
+POM_VERSION=$(cat ${DIR}/../../pom.xml | grep "^        <version>.*</version>$" | awk -F'[><]' '{print $3}')
+
+rm -rf ${DEPENDENCY_DIR}
+
+TERMINAL_AGENTS=("org.eclipse.che:exec-agent:${POM_VERSION}:tar.gz:linux_amd64"
+                 "org.eclipse.che:exec-agent:${POM_VERSION}:tar.gz:linux_arm7");
```
benoitf: in which case the linux_arm7 will work ?
as all docker containers being based on linux/amd64 I don't see a usecase where arm arch will work ?

dockerfiles/init/modules/lighttpd/templates/lighttpd.conf.erb
```diff
@@ -0,0 +1,323 @@
...
+# }}}
+
+# {{{ debug
+# debug.log-request-header   = "enable"
+# debug.log-response-header  = "enable"
+# debug.log-request-handling = "enable"
+# debug.log-file-not-found   = "enable"
+# }}}
+
+# vim: set ft=conf foldmethod=marker et :
```
benoitf: maybe we could drop tons of the commented lines ?

dockerfiles/agents/build.sh
tolusha: The only case I can see is in creating ssh machine (artic device)

dockerfiles/init/modules/lighttpd/templates/lighttpd.conf.erb
```diff
@@ -0,0 +1,323 @@
...
+
+# event handler (defaults to "poll")
+# see performance.txt
+#
+# for >= linux-2.4
+#   server.event-handler = "linux-rtsig"
+# for >= linux-2.6
+#   server.event-handler = "linux-sysepoll"
+# for FreeBSD
+#   server.event-handler = "freebsd-kqueue"
```
riuvshin: this is default content of lighttpd.conf

dockerfiles/init/modules/lighttpd/templates/lighttpd.conf.erb
riuvshin: this is default content of lighttpd.conf it have info for all operations systems I think

dockerfiles/init/modules/lighttpd/manifests/init.pp
```diff
@@ -0,0 +1,12 @@
+class lighttpd {
+  file { "/opt/codenvy/config/lighttpd":
+    ensure  => "directory",
+    mode    => "755",
```
riuvshin: this is common folder permission we always use that mode for folders we manage

dockerfiles/init/modules/compose/templates/docker-compose.yml.erb
riuvshin: Anatoly wanted to keep open that port in dev mode to test server without codenvy.
We don;t need check this port I think as it is only in dev mode
(but I still want OK to remove that.)

benoitf: either code is removed or either the port is checked in dev mode (as we do for debug port)

dockerfiles/init/modules/lighttpd/templates/lighttpd.conf.erb
riuvshin: this is default content of that config, Im OK to drop comments.

tolusha:
@benoitf 
Updated

riuvshin:
@tolusha Please prvide  PR for customer-saas same way, you may need @benoitf  help

tolusha:
`lighttpd` is the most light weight web serve I've found. That's was the purpose of using it.

tolusha:
@skabashnyuk 
Whose answer you are waiting for?

skabashnyuk:
ok up to you

PR #1366, opened 2016-12-16
Adding runbook to admin docs

bmicklea:
Adds the runbook content from readme.io to these docs. This content still needs to be updated for dockeried codenvy.


bmicklea:
@JamesDrummond - can you check the page meta info like the permalink, etc... I copied it from one of the other admin pages.

PR #1367, opened 2016-12-16
Allow to use templating for product in docs

benoitf:
Introduce a plugin for "product"
allow to use `product.name`, `product.mini_name`, `product.formal_name`

example : `{% product.mini_name %}`


benoitf:
@JamesDrummond ?

TylerJewell:
Ok - for those reviewing this, in the plugin, this is where the various constants would be added. So we would have similar constants that we add to the CLI:
1. CLI image name
2. CLI init image name
3. Various product names

There may be other things in the future.

benoitf:
rebased pull request

benoitf:
@JamesDrummond well it's true that we can use jekyll as well with site. prefix
But I was thinking it was more natural to use `{% product.mini_name %}`instead of {% site .product.mini_name %} as it's shorter.

we may probably as well use yaml to populate the plugin

BTW I've not written many docs so I let you and @TylerJewell pickup what is the best for you

bmicklea:
@JamesDrummond I think we're going with the in-Jekyll approach for this, right?  If so this PR should be closed without merging, right?

JamesDrummond:
@bmicklea Close without merging. Yes.

PR #1379, opened 2016-12-19
Add Simulated Scaling Docs

TylerJewell:
### What does this PR do?
1. Adds a section to the managing doc of the admin-guide to discuss how to use Docker Machine to simulate running a 3 node Codenvy cluster where 2 of the nodes are workspace node and the third is a Codenvy node.

2. Need to get more feedback from @garagatyi on my last open comments, which is why we are having admins modify codenvy.env to setup the cluster instead of using the `codenvy add-node` command which is the prescribed way.


TylerJewell:
Closing to recreate given the git history rewrite.

PR #1381, opened 2016-12-20
Fixed pom for docs

JamesDrummond:
This PR is dependent on https://github.com/eclipse/che/pull/3443 before merge. Line 52 will change to master. This maven build is dependent on host having docker installed.

### DIFF
--- a/docs/pom.xml
+++ b/docs/pom.xml
@@ -40,11 +40,96 @@
                         </goals>
                         <configuration>
                             <target>
-                                <exec dir="${project.basedir}" executable="jekyll" failonerror="true">
-                                    <arg value="build" />
-                                    <arg value="--config" />
-                                    <arg value="assembly/build_war.yml" />
+                                <available file="${project.build.directory}/che-repo" property="cheRepoExists" type="dir"/>
+                                <mkdir dir="${project.build.directory}/che-repo"/>
+                                <exec dir="${project.basedir}" executable="docker" failonerror="true">
+                                    <arg value="run" />
+                                    <arg value="--name" />
+                                    <arg value="jekyll_git_che" />
+                                    <arg value="jekyll/jekyll" />
+                                    <arg value="sh" />
+                                    <arg value="-c" />
+                                    <arg value="cd /srv/jekyll/ ; git init ; git remote add -t add-docs-images -f origin https://github.com/eclipse/che.git ; git checkout add-docs-images" />
+                                </exec>                                
+                                <exec dir="${project.basedir}" executable="docker" failonerror="true">
+                                    <arg value="cp" />
+                                    <arg value="jekyll_git_che:/srv/jekyll/" />
+                                    <arg value="${project.build.directory}/che-repo" />
                                 </exec>
+                                <exec dir="${project.basedir}" executable="docker" failonerror="true">
+                                    <arg value="rm" />
+                                    <arg value="-f" />
+                                    <arg value="jekyll_git_che" />
+                                </exec>
+                                <exec dir="${project.build.directory}" executable="docker" failonerror="true">
+                                    <arg value="run" />
+                                    <arg value="-d" />
+                                    <arg value="--name" />
+                                    <arg value="jekyll-war-build" />
+                                    <arg value="jekyll/jekyll" />
+                                    <arg value="tail" />
+                                    <arg value="-f" />
+                                    <arg value="/dev/null" />
+                                </exec>
+                                <exec dir="${project.build.directory}" executable="docker" failonerror="true">
+                                    <arg value="cp" />
+                                    <arg value="${project.basedir}" />
+                                    <arg value="jekyll-war-build:/tmp" />
+                                </exec>
+                                <exec dir="${project.build.directory}" executable="docker" failonerror="true">
+                                    <arg value="exec" />
+                                    <arg value="jekyll-war-build" />
+                                    <arg value="sh" />
+                                    <arg value="-c" />
+                                    <arg value="mv /tmp/docs/* /srv/jekyll/" />
+                                </exec>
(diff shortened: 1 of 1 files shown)


docs/pom.xml
```diff
@@ -40,11 +40,96 @@
                         </goals>
                         <configuration>
                             <target>
-                                <exec dir="${project.basedir}" executable="jekyll" failonerror="true">
+                                <available file="${project.build.directory}/che-repo" property="cheRepoExists" type="dir"/>
```
benoitf: is that requirement of docker is ok ? how do you manage codenvy in codenvy ?

docs/pom.xml
```diff
@@ -40,11 +40,96 @@
...
+                                <mkdir dir="${project.build.directory}/che-repo"/>
+                                <exec dir="${project.basedir}" executable="docker" failonerror="true">
+                                    <arg value="run" />
+                                    <arg value="--name" />
+                                    <arg value="jekyll_git_che" />
+                                    <arg value="jekyll/jekyll" />
+                                    <arg value="sh" />
+                                    <arg value="-c" />
+                                    <arg value="cd /srv/jekyll/ ; git init ; git remote add -t add-docs-images -f origin https://github.com/eclipse/che.git ; git checkout add-docs-images" />
```
benoitf: version of che should be specified.
It's useful when we have to perform bugfixes release because in that case we don't use master branch of che inside codenvy

docs/pom.xml
```diff
@@ -40,11 +40,96 @@
...
+                                    <arg value="${project.build.directory}" />
+                                </exec>
+                                <exec dir="${project.basedir}" executable="docker" failonerror="true">
+                                    <arg value="rm" />
+                                    <arg value="-f" />
+                                    <arg value="jekyll-war-build" />
+                                </exec>
+                                <delete file="${project.build.directory}/_site/feed.xml" />
+                                <delete file="${project.build.directory}/_site/feed.xslt.xml" />
+                                <delete file="${project.build.directory}/_site/sitemap.xml" />
```
benoitf: Instead of launching 15 times the docker container, I think we should give a shell script to docker container and only run once
Also docker run command should be using --rm flag no ? as it will end with several non-cleaned containers

docs/pom.xml
```diff
@@ -57,7 +142,7 @@
                     <failOnMissingWebXml>false</failOnMissingWebXml>
                     <webResources>
                         <resource>
-                            <directory>target/_site/docs</directory>
+                            <directory>${project.build.directory}/_site</directory>
```
benoitf: generation of the site should happen in the target folder.
So all generated stuff is removed when we call `mvn clean` else, it will mix source code and generated code in the current folder

docs/pom.xml
benoitf: also sparse checkout could be done to only keep docs folder

JamesDrummond: So I would use project.version of the pom.xml file or nightly?

docs/pom.xml
benoitf: Ok but codenvy.io ?

docs/pom.xml
JamesDrummond: Yes that is a great idea to write an external script. Originally I was worried this wouldn't be acceptable.

The --rm I don't think can be used as I need to run `cp` and `exec` on the running command from "host". We could use `tail -f /dev/null` but won't that block `--rm` from being used?

docs/pom.xml
benoitf: yes but as said it would be a lot more compliant to use maven artifacts instead of doing some clone operations.  
It simplify dependencies as it's a feature natively provided by maven

docs/pom.xml
JamesDrummond: @benoitf I think this is what I have done. I will change any `exec dir="${project.basedir}"` to `exec dir="${project.build.directory}"` to ensure this. Docker commands do no use `volumes` but uses instead `cp` so there is no chance for it to write to ${project.basedir}.

benoitf: Ok

docs/pom.xml
benoitf: I just think that 10 commands are overkill no ? And at the end all docker containers should be removed

docs/pom.xml
JamesDrummond: @benoitf How would I know to clone a branch of che versus using master which is needed for nightly builds. If I find "-SNAPSHOT" in the project.version I use nightly and if it is not there then use project.version to checkout that tagged release of che? How do I find sha1 for a certain tagged release? `git checkout <sha1 rev>`

docs/pom.xml
JamesDrummond: You cannot build in codenvy.io unless you port bind an external docker daemon to the workspace machine.

docs/pom.xml
benoitf: Because it seems it's reinventing the wheel that something that is provided by maven
With maven, che repository could produce maven artifacts that could be used in codenvy ( or any other repository based on maven ( including artik, etc.) 
So, no need to parse sha-1, no need to clone repositories, no need to know the branch name, etc it's done by maven

docs/pom.xml
JamesDrummond: Binding external docker daemons is something I would like to look into next quarter. There are many users that want to use docker in the workspace machine. It would be part of the CLI like che-mount utility. Most docker daemons use docker.sock though which I cannot mount inside a workspace machine. This requires some investigation. 

I would like to actually end up with the host providing their own external docker daemon automatically. Either secure a dind container or provide a VM with docker daemon. This requires additional investigation though.

JamesDrummond:
Will not be merged.

PR #1406, opened 2016-12-22
Provide capitalized product_mini_name

slemeur:
add:
* product_mini_name:  "codenvy"
* product_mini_name_capitalized:  "Codenvy"


bmicklea:
People will most often use the capitalized version so that name is really long what if we switched them to:

product_mini_name: "Codenvy"
product_mini_name_cmd: "codenvy"

or 

product_mini_name_cli: "codenvy"

slemeur:
close without merging.
Docs moved to another repo + changes in the product_mini_names format

PR #1427, opened 2016-12-26
CODENVY-1321 Add checking instance id to avoid NPE

sleshchenko:
### What does this PR do?
Add checking instance id to avoid NPE.

https://github.com/codenvy/codenvy/issues/1321

### Tests written?
Yes

---
This change is [Reviewable](https://reviewable.io/reviews/codenvy/codenvy/1427)

### DIFF
--- a/wsmaster/codenvy-hosted-api-permission/src/main/java/com/codenvy/api/permission/server/InstanceParameterValidator.java
+++ b/wsmaster/codenvy-hosted-api-permission/src/main/java/com/codenvy/api/permission/server/InstanceParameterValidator.java
@@ -1,0 +1,64 @@
+/*
+ *  [2012] - [2017] Codenvy, S.A.
+ *  All Rights Reserved.
+ *
+ * NOTICE:  All information contained herein is, and remains
+ * the property of Codenvy S.A. and its suppliers,
+ * if any.  The intellectual and technical concepts contained
+ * herein are proprietary to Codenvy S.A.
+ * and its suppliers and may be covered by U.S. and Foreign Patents,
+ * patents in process, and are protected by trade secret or copyright law.
+ * Dissemination of this information or reproduction of this material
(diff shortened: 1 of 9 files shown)


wsmaster/codenvy-hosted-api-permission/src/main/java/com/codenvy/api/permission/server/InstanceParameterValidator.java
```diff
@@ -0,0 +1,52 @@
...
+import org.eclipse.che.api.core.NotFoundException;
+
+import javax.inject.Inject;
+
+/**
+ * Validates that provided instance parameter is valid
+ *
+ * @author Sergii Leschenko
+ */
+public class InstanceParameterValidator {
```
mshaposhnik: S/ton?

wsmaster/codenvy-hosted-api-permission/src/main/java/com/codenvy/api/permission/server/InstanceParameterValidator.java
```diff
@@ -0,0 +1,54 @@
...
+     * @param domain
+     *         the domain of specified {@code instance}
+     * @param instance
+     *         the instance to check
+     * @throws BadRequestException
+     *         if specified {@code instance} is not valid
+     * @throws NotFoundException
+     *         if specified {@code domain} is unsupported
+     */
+    public void validate(String domain, String instance) throws BadRequestException, NotFoundException {
```
evoevodin: Maybe +
- check domain for null, throw 400
- put `@Nullable` on instance field

sleshchenko: Thanks, good suggestions

PR #1434, opened 2016-12-27
CODENVY-1392: fix sync agent start check

### LINKED ISSUE #1392: Workspace start occasionally fails if docker image doesn't contain ssh and rsync
Workspace start occasionally fails if docker image doesn't contain ssh and rsync. 

Related to:
#1271 
#1371 

**Reproduction Steps:**

Use bitnami stacks to create and start WS several times.

**Expected behavior:**

These WSs start successfully. 

**Observed behavior:**

Sometimes WS start fails and sometimes succeeds. 

**Codenvy version:** m9-snapshot
**OS and version:** Ubuntu 16.04

**Additional information:**

Problem only started happening recently, didn't happen in an older version of Codenvy: [Yes] (because of other reasons such WSs always failed start before merging solution for #1271)
Problem can be reliably reproduced: [No - it's random]

garagatyi:
### What does this PR do?
Change way of verification that sync agent is launched in machine.

Fixes #1392 

### Previous Behavior
Start of WS sometimes failed when ssh or/and rsync are not installed in container.

### New Behavior
Start of WS finishes successfully when ssh or/and rsync are not installed in container.

### Tests written?
No

### DIFF
--- a/plugins/plugin-hosted/codenvy-machine-hosted/src/main/java/com/codenvy/machine/agent/launcher/ExternalRsyncAgentLauncherImpl.java
+++ b/plugins/plugin-hosted/codenvy-machine-hosted/src/main/java/com/codenvy/machine/agent/launcher/ExternalRsyncAgentLauncherImpl.java
@@ -16,6 +16,8 @@ package com.codenvy.machine.agent.launcher;
 
 import org.eclipse.che.api.agent.server.launcher.AbstractAgentLauncher;
 import org.eclipse.che.api.agent.server.launcher.AgentLaunchingChecker;
+import org.eclipse.che.api.agent.server.launcher.CommandExistsAgentChecker;
+import org.eclipse.che.api.agent.server.launcher.CompositeAgentLaunchingChecker;
 import org.eclipse.che.api.agent.shared.model.Agent;
 import org.eclipse.che.api.core.ServerException;
 import org.eclipse.che.api.machine.server.spi.Instance;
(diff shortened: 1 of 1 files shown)


plugins/plugin-hosted/codenvy-machine-hosted/src/main/java/com/codenvy/machine/agent/launcher/ExternalRsyncAgentLauncherImpl.java
```diff
@@ -41,7 +43,10 @@
     @Inject
     public ExternalRsyncAgentLauncherImpl(@Named("che.agent.dev.max_start_time_ms") long agentMaxStartTimeMs,
                                           @Named("che.agent.dev.ping_delay_ms") long agentPingDelayMs) {
-        super(agentMaxStartTimeMs, agentPingDelayMs, AgentLaunchingChecker.DEFAULT);
+        super(agentMaxStartTimeMs,
+              agentPingDelayMs,
+              new CompositeAgentLaunchingChecker(new CommandExistsAgentChecker("rsync"),
+                                                 AgentLaunchingChecker.DEFAULT));
```
skabashnyuk: are you sure it's good idea to create objects here, may be we should delegate it to container?

garagatyi: Do you suggest to create ExternalRsyncAgentLauncherImpl in module instead or smth else?

Issue #1431, opened 2016-12-26
dev mode not consistent


riuvshin:
after introducing agents as separate container by @tolusha dev mode become not consistent because we mount only codenvy API server for dev mode but not agents. 
so we have to mount agents binaries to agents container same way as we do this with api server.
NOTE that we don;t need this for che as in che we don't have agents container so this is only needed for codenvy.

I've tried to do this but after consolidation of che and codenvy CLI I don't understand how to make something specific for codenvy only.

TylerJewell:
Methods can be overridden so in the codenvy cli we can duplicate and override any method from the base cli. We also have pre and post logic methods for key commands like init.  I will show you on Wednesday.

TylerJewell:
@riuvshin - I will be back in the office on Thursday. 

Prior to that, where I would start with modifications are files in this folder:
https://github.com/codenvy/codenvy/tree/master/dockerfiles/cli/scripts

These methods override the same methods called in the base CLI. if the method does not existing in this repository, then it uses the same method in the base.

I would do a search for all of the locations that have DEVELOPMENT_MODE statements, and then add the appropriate logic there.

riuvshin:
@TylerJewell thx for explanations, i have already some progress on this.

riuvshin:
PRs proposed.

PR #1436, opened 2016-12-27
in dev mode use agents from the repo same way as api server

riuvshin:
fixes needed for https://github.com/codenvy/codenvy/issues/1431

Those changes affects ONLY dev mode.
to make dev mode consistent we have to pick agents (ws agent and terminal) same way as api server to containers from `:/repo`

@TylerJewell @benoitf please carefully review this I want to be sure I didn't missed anything


dockerfiles/init/manifests/codenvy.pp
```diff
@@ -288,6 +288,10 @@
   $puppet_src_folder = getValue("CHE_CONFIG","/path/to/codenvy/codenvy/puppet/sources")
 # path to codenvy tomcat for development mode
   $codenvy_development_tomcat = getValue("CHE_ASSEMBLY","/path/to/codenvy_tomcat")
+# path to codenvy ws agent for development mode
+  $codenvy_development_ws_agent = getValue("WS_AGENT_ASSEMBLY","/path/to/codenvy_tomcat")
+# path to codenvy terminal agent for development mode
+  $codenvy_development_terminal_agent = getValue("TERMINAL_AGENT_ASSEMBLY","/path/to/codenvy_tomcat")
```
garagatyi: Terminal is not related to tomcat server. Why it is on such path?

riuvshin: this is "default" path which never be used as we always set this in CLI but I agree i will update  this to be more clear, thanks

dockerfiles/cli/scripts/cli.sh
```diff
@@ -174,6 +174,26 @@ cmd_config_post_action() {
     log "docker volume create --name=codenvy-postgresql-volume >> \"${LOGS}\""
     docker volume create --name=codenvy-postgresql-volume >> "${LOGS}"
   fi
+
+  if [ "${CHE_DEVELOPMENT_MODE}" = "on" ]; then
```
garagatyi: Maybe we should have single folder with agents instead of adding small peace of code in this script on addition of each agent?

riuvshin: in fact all agents will appear in single dir

riuvshin:
@TylerJewell I've updated PR, can you please review changes, I changed some ENV vars names to be more clear.

PR #1470, opened 2017-01-04
expose swarm port to host

riuvshin:
As per @skryzhny request, we should have exposed swarm port to the host.

---
This change is [Reviewable](https://reviewable.io/reviews/codenvy/codenvy/1470)


TylerJewell:
What about cli preflight checks?

riuvshin:
sorry for that here is PR  https://github.com/codenvy/codenvy/pull/1471

Issue #1506, opened 2017-01-11
NPE when creating workspace when there is already running ones


mkuznyetsov:
After https://github.com/eclipse/che/issues/3082 has been merged, RamChecker can produce NullPointerException, because it uses WorkspaceService#getByNamespace to get runtimes, which are no longer returned there. 

Needs to rewrite RamChecker accordingly to this

PR #1507, opened 2017-01-11
Fix NPE when using getByNamespace to fetch runtimes

mkuznyetsov:
### What does this PR do?
Fix NPE where runtimes were fetched by WorkspaceService#getByNamespace method.

https://github.com/codenvy/codenvy/issues/1506

Then needs to revert https://github.com/eclipse/che/pull/3661 after merge


wsmaster/codenvy-hosted-api-resource/src/main/java/com/codenvy/resource/api/usage/tracker/RamResourceUsageTracker.java
```diff
@@ -51,16 +54,24 @@ public RamResourceUsageTracker(Provider<WorkspaceManager> workspaceManagerProvid
...
-                                                                .map(ws -> ws.getRuntime().getMachines())
-                                                                .flatMap(List::stream)
-                                                                .mapToInt(machine -> machine.getConfig()
-                                                                                            .getLimits()
-                                                                                            .getRam())
-                                                                .sum();
+        WorkspaceManager manager = workspaceManagerProvider.get();
+
+        final List<String> runningWorkspaces = manager.getByNamespace(account.getName())
+                                                       .stream()
```
sleshchenko: Fix formatting here

mkuznyetsov:
@garagatyi @sleshchenko it has been reworked, now it uses new method from workspaceManager with boolean parameter

PR #1533, opened 2017-01-16
Make single docker image layer when adding tomcat

skabashnyuk:
Reduce image size from 552Mb to 355Mb.
Instead of copy zip file with one command and extract archive in next I've used tar.gz archive in combination with docker ADD command.


TylerJewell:
Wow!!  I didn't know that would add such an improvement!  Can the same work for Che?

skabashnyuk:
Che is using the same technique already.

benoitf:
yes this approach was used for che when I used symlink as well.
Also, maybe I could add build.sh as well to codenvy/codenvy

Issue #1498, opened 2017-01-10
Snapshot isn't deleted from the local docker repository when workspace is stopped and snapshot had been used to start workspace


skryzhny:
After QA tests session I'm observing lot of snapshots in local docker registry
```
# docker images | less
REPOSITORY                                                                            TAG                 IMAGE ID            CREATED             SIZE
codenvy-stg.com:5000/machine_snapshot_s5rjptdlv923ypob                                latest              3413155ff49d        9 minutes ago       820.2 MB
eclipse-che/workspacewx52ce577dk67jha_machine0cjsrw3qi33t3br4_iedexmain_dev-machine   latest              3413155ff49d        9 minutes ago       820.2 MB
codenvy-stg.com:5000/machine_snapshot_opp4whc645jjoc2t                                latest              f9eb4f33562a        2 days ago          816.8 MB
codenvy-stg.com:5000/machine_snapshot_99qwc00a9ew10xce                                latest              97adb5a59d82        2 days ago          833.7 MB
codenvy-stg.com:5000/machine_snapshot_azyjhiflprm9b3x9                                latest              3ca42687545e        2 days ago          816.8 MB
codenvy-stg.com:5000/machine_snapshot_4nlbdzrhqnf12ise                                latest              6230d79f4372        2 days ago          796 MB
codenvy-stg.com:5000/machine_snapshot_n1i4d386x0aiofqr                                latest              82cffd33acdd        2 days ago          819.3 MB
codenvy-stg.com:5000/machine_snapshot_054ecba419hrrjwa                                latest              c02c90a85406        2 days ago          830.6 MB
codenvy-stg.com:5000/machine_snapshot_5uiy8w27p6bf9ab0                                latest              175dfe1be40b        2 days ago          827.7 MB
codenvy-stg.com:5000/machine_snapshot_6y1v17p55iucnw80                                latest              c8e9008a3a5f        3 days ago          833.7 MB
codenvy-stg.com:5000/machine_snapshot_4gtcde2igplnriww                                latest              e76eada4645f        3 days ago          816.8 MB
codenvy-stg.com:5000/machine_snapshot_3zdwc350zl1iek3g                                latest              5a43a6611070        3 days ago          816.8 MB
codenvy-stg.com:5000/machine_snapshot_kwrzk8jtbfnpg7bw                                latest              35b8dcc09e91        3 days ago          820.2 MB
[...]

PR #1566, opened 2017-01-20
CODENVY-1498: Cleanup snapshots images after workspace start

mmorhun:
### What does this PR do?
Cleans up images of snapshots after workspace start.
This images appear after build of workspace image from snapshot (which image is pulled by swarm during build).
Deletion of images is implemented through `ExecutorService` because of bug in pure docker swarm. The bug is that swarm cannot delete just pulled image, because swarm hasn't seen it yet. To avoid deletion fail it is needed to wait a few seconds. But we cannot wait in the workspace starting thread, so this job is placed into separate thread.

#1498 

### Previous Behavior
After restoring of workspace from snapshot, snapshot image wasn't deleted from node.

### New Behavior
After workspace starting snapshot image is deleted.

### Tests written?
No


plugins/plugin-hosted/codenvy-machine-hosted/src/main/java/com/codenvy/machine/HostedMachineProviderImpl.java
```diff
@@ -116,6 +130,17 @@ public HostedMachineProviderImpl(DockerConnector docker,
         this.docker = docker;
         this.dockerCredentials = dockerCredentials;
         this.tokenRegistry = tokenRegistry;
+
+        this.snapshotImagesCleanerService = createSnapshotImagesCleanerExecutor();
+
+    }
+
+    private ScheduledExecutorService createSnapshotImagesCleanerExecutor() {
```
garagatyi: It would look simpler without separate method

plugins/plugin-hosted/codenvy-machine-hosted/src/main/java/com/codenvy/machine/HostedMachineProviderImpl.java
```diff
@@ -167,10 +192,37 @@ protected void pullImage(CheServiceImpl service,
             if (workDir != null) {
                 FileCleaner.addFile(workDir);
             }
+            // We use build instead of pull because of better stability.
```
garagatyi: First line in this comment duplicates javadocs of the whole method

plugins/plugin-hosted/codenvy-machine-hosted/src/main/java/com/codenvy/machine/HostedMachineProviderImpl.java
```diff
@@ -215,4 +267,22 @@ protected void buildImage(CheServiceImpl service,
...
+    @PreDestroy
+    private void preDestroy() {
+        finalizeSnapshotImagesCleaner();
+    }
+
+    /**
+     * Tries to clean up snapshots images which is in the cleaner queue, but were canceled due to server stopping.
+     * To avoid freezing of server on stop performs post clean in separate thread.
+     */
+    private void finalizeSnapshotImagesCleaner() {
```
garagatyi: It would look simpler without separate method

plugins/plugin-hosted/codenvy-machine-hosted/src/main/java/com/codenvy/machine/HostedMachineProviderImpl.java
```diff
@@ -215,4 +267,22 @@ protected void buildImage(CheServiceImpl service,
...
+        List<Runnable> queue = snapshotImagesCleanerService.shutdownNow();
+        Thread snapshotsCleaner = new Thread(() -> queue.forEach(Runnable::run));
+        snapshotsCleaner.setDaemon(true);
+        snapshotsCleaner.setName("PostSnapshotImagesCleaner");
```
garagatyi: Consider another name, for example PreDestroySnapshotImagesCleaner, FinalizingSnapshotImagesCleaner, SnapshotImagesCleaningFinalizer

mmorhun: Thank you, current name isn't good.

plugins/plugin-hosted/codenvy-machine-hosted/src/main/java/com/codenvy/machine/HostedMachineProviderImpl.java
mmorhun: I do not think so. Separate method group logic, provides method docs, and from its name someone can quickly figure out what that code does.

skabashnyuk: inline this method please. You can add this comment to code.

plugins/plugin-hosted/codenvy-machine-hosted/src/main/java/com/codenvy/machine/HostedMachineProviderImpl.java
```diff
@@ -167,10 +187,36 @@ protected void pullImage(CheServiceImpl service,
...
 
     /**
+     * Sometimes swarm cannot delete image after its pull during a few seconds.
+     * To workaround that problem and avoid redundant delay on workspace start
+     * we must clean up snapshot image in separate thread after some delay.
+     *
+     * @param image
+     *         image to clean, e.g. 172.11.12.13:5000/machine_snapshot_abcdef1234567890:latest
+     */
+    private void submitCleanSnapshotImageTask(String image) {
```
skabashnyuk: why don't you use schedule annotation? It will looks much more simpler?

mmorhun: because our schedule annotations designed for periodic tasks, unlike this which require to run cleanup task only once after some delay.

Issue #1550, opened 2017-01-18
Optimization of cli execution time for port open check


benoitf:
I've noticed using traces of all commands executed by the CLI that the check of opening port is taking a very long time (30s on my computer)

I've added
```
PS4='+$(date "+%T %N")\011 (${BASH_SOURCE}:${LINENO}): ${FUNCNAME[0]:+${FUNCNAME[0]}(): } '
set -x
```
at the top of CLI entrypoint.sh to get traces

Code is at
https://github.com/codenvy/codenvy/blob/5.1.0/dockerfiles/cli/scripts/cli.sh#L131-L163

Improvements could include :
- use of a batch command (so we can check in one docker container run all the ports)
- cache the result of an open port (see [1])
- allow to skip port open

[1] cache:
for example for port 80
we do
```
$(port_open 80 && echo "${GREEN}[AVAILABLE]${NC}" || echo "${RED}[ALREADY IN USE]${NC}") \n"
```

and few lines after
```if ! $(port_open 80)```which call again a new docker container

**Reproduction Steps:**

**Codenvy version:**    5.1.0

**OS and version:**    macOS sierra

**Docker version:** 1.12.6

TylerJewell:
One issue seems to be that we haven't found a clean way to check a port once and then set a flag.  If you notice the port_open command, we run it twice and so it's a double task.  I thought we had found a solution, but the variable flag for the test and printout was not working in a combined macro command.

TylerJewell:
BTW - in the branch for Che where we are adding in checks during preflight for open network connections, we have also added in the option of `--fast` to skip all preflight checks.  So this would be a way to skip port checking.

Previously I tried to author a single line which would do a `port_open && || ` as a batch set of commands.  But Bash does not seem to accept this.  So we should probably create a single method that performs a port_check, output, and interrupt all within a single method if possible.  The problem with interrupting in a single port check is that I wanted to run all of the preflight checks so that the end user can see all ports that are failing at one go instead of interrupting at the first port.

benoitf:
thinking overnight I was thinking about the following pattern. (to keep the fact that end user should know each failing port and not the first one)

: we run a single docker run command with all port mapping like `-p 8080:8080 -p 9000:9000`

--> if it's a success (probably most of the time) we have a single execution (so we dramatically reduce the boot time) and then we can display OK for each ports.
[...]

PR #1575, opened 2017-01-20
CLI - Adapt to Che CLI changes for custom assembly, faster port check

TylerJewell:
### What does this PR do?
1. Adds in ability to mount `:/assembly` so that it mounts a Tomcat binary directly without mounting the entire repository.
2. Adds in a refactoring of the preflight checks to test all ports at the same time.  This makes overall performance faster.  If any of the ports fail, then we iterate through each port check one at a time.

https://github.com/codenvy/codenvy/issues/1550

### Changelog
Adapt to Eclipse Che #1550 - adds `:/assembly` volume mount

### Release Notes
N/A

### Docs PR
N/A

### DIFF
--- a/dockerfiles/cli/scripts/cli.sh
+++ b/dockerfiles/cli/scripts/cli.sh
@@ -7,7 +7,7 @@
 #
 
 cli_pre_init() {
-  GLOBAL_HOST_IP=${GLOBAL_HOST_IP:=$(docker_run --net host eclipse/che-ip:nightly)}
+  GLOBAL_HOST_IP=${GLOBAL_HOST_IP:=$(docker_run --net host ${UTILITY_IMAGE_CHEIP})}
   DEFAULT_CODENVY_HOST=$GLOBAL_HOST_IP
   CODENVY_HOST=${CODENVY_HOST:-${DEFAULT_CODENVY_HOST}}
   CODENVY_PORT=80
--- a/dockerfiles/cli/scripts/cli.sh
+++ b/dockerfiles/cli/scripts/cli.sh
@@ -111,11 +111,52 @@ cmd_init_reinit_pre_action() {
   fi
 }
 
+check_all_ports(){
+  docker run -d -p 80:80 \
+                -p 443:443 \
+                -p 2181:2181 \
+                -p 5000:5000 \
+                -p 23750:23750 \
+                -p 23751:23751 \
+                -p 32001:32001 \
+                -p 32101:32101 \
+                --name fake ${UTILITY_IMAGE_ALPINE} \
+                                 httpd -f -p 80 \
+                                          -p 443 \
+                                          -p 2181 \
+                                          -p 5000 \
+                                          -p 23750 \
+                                          -p 23751 \
+                                          -p 32001 \
+                                          -p 32101 -h /etc/ > /dev/null 2>&1
+
+  NETSTAT_EXIT=$?
+  docker rm -f fake > /dev/null 2>&1
+
+  NETSTAT_EXIT_DEBUG=""
+  if debug_server; then
+    docker run -d -p $CODENVY_DEBUG_PORT:$CODENVY_DEBUG_PORT \
+                  -p 9000:9000 \
+                  --name fake ${UTILITY_IMAGE_ALPINE} \
+                                   httpd -f -p $CODENVY_DEBUG_PORT \
+                                            -p 9000 -h /etc/ > /dev/null 2>&1
+    NETSTAT_EXIT_DEBUG=$?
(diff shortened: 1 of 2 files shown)


dockerfiles/cli/scripts/cli.sh
```diff
@@ -111,11 +111,52 @@ cmd_init_reinit_pre_action() {
...
+                -p 443:443 \
+                -p 2181:2181 \
+                -p 5000:5000 \
+                -p 23750:23750 \
+                -p 23751:23751 \
+                -p 32001:32001 \
+                -p 32101:32101 \
+                --name fake ${UTILITY_IMAGE_ALPINE} \
+                                 httpd -f -p 80 \
+                                          -p 443 \
```
benoitf: do we need to listen with httpd ? Is that -p on docker is not enough ?

TylerJewell: I did a small test with it and without it because I thought the same thing.  However, without the httpd, the test would pass even though the port was bound.

dockerfiles/cli/scripts/cli.sh
```diff
@@ -128,39 +169,36 @@ cmd_start_check_ports() {
...
+    text   "         port 2181 (zookeeper): $(port_open 2181 && echo "${GREEN}[AVAILABLE]${NC}" || echo "${RED}[ALREADY IN USE]${NC}") \n"
+    text   "         port 5000 (registry):  $(port_open 5000 && echo "${GREEN}[AVAILABLE]${NC}" || echo "${RED}[ALREADY IN USE]${NC}") \n"
+    text   "         port 23750 (socat):    $(port_open 23750 && echo "${GREEN}[AVAILABLE]${NC}" || echo "${RED}[ALREADY IN USE]${NC}") \n"
+    text   "         port 23751 (swarm):    $(port_open 23751 && echo "${GREEN}[AVAILABLE]${NC}" || echo "${RED}[ALREADY IN USE]${NC}") \n"
+    text   "         port 32001 (jmx):      $(port_open 32001 && echo "${GREEN}[AVAILABLE]${NC}" || echo "${RED}[ALREADY IN USE]${NC}") \n"
+    text   "         port 32101 (jmx):      $(port_open 32101 && echo "${GREEN}[AVAILABLE]${NC}" || echo "${RED}[ALREADY IN USE]${NC}") \n"
+    if debug_server; then
+      text   "         port ${CODENVY_DEBUG_PORT} (debug):     $(port_open ${CODENVY_DEBUG_PORT} && echo "${GREEN}[AVAILABLE]${NC}" || echo "${RED}[ALREADY IN USE]${NC}") \n"
+      text   "         port 9000 (lighttpd):  $(port_open 9000 && echo "${GREEN}[AVAILABLE]${NC}" || echo "${RED}[ALREADY IN USE]${NC}") \n"
```
benoitf: It looks like we duplicate a lot of code
For example value 9000.  Lighthttpd. 
It would be great to give an array with all ports/ description 9000/ lighthttpd and the function uses that array

TylerJewell: Yes I agree it would be a better implementation and easier to maintain.

TylerJewell: I added a first pass of the improvement.

I am not quite sure how to handle it.  But I think we can probably put this array into `cmd_pre_init` for a custom CLI assembly and then this would probably remove the need for the custom port check methods in all of the various CLIs.

dockerfiles/cli/scripts/cli.sh
benoitf: ok I think this is different on docker for mac
```
$ netstat -an | grep LISTEN
tcp4       0      0  *.8080                 *.*                    LISTEN     
```

```
docker run -p 8080:8080 -p 9000:9000  --name fb alpine:3.4 echo "hello"
c75c8218bbcd0467f9901619d7f91766e0fa82f9344de150847b2343a2538732
docker: Error response from daemon: driver failed programming external connectivity on endpoint fb (9bc1f0ac88b59c3627355031934034d78f5412bd3db231e09951f6fd61382236): Error starting userland proxy: Bind for 0.0.0.0:8080 failed: port is already allocated.
```
(no need to run as a daemon and also httpd

TylerJewell: That is what you need to run if you are native on the host.  However, the command that is in this PR is designed for running inside of a container and then testing the host's bound ports.  Just did a test on Mac and it worked as expected.

TylerJewell:
Che has been merged - awaiting romans approval.

benoitf:
@riuvshin ?

Issue #1592, opened 2017-01-24
Change workspace idle timeout for on-premises installs


bmicklea:
The workspace timeout is meant to be set in the `codenvy.env` file in this section:

```
# Idle Timeout
#     The length of time that a user is idel with their workspace when the system will suspend the
#     workspace by snapshotting it and then stopping it. Idleness is determined by the length of
#     time that the user has not interacted with the workspace, meaning that one of our agents
#     has not received instructions. Leaving a browser window open counts as idleness time.
#CODENVY_MACHINE_WS_AGENT_INACTIVE_STOP_TIMEOUT_MS=600000
```

Requirement:
1) The default value should be 14,400,000 MS (4 hours)

bmicklea:
@skabashnyuk please consider for your next sprint.

TylerJewell:
No - when a default codenvy.env, *nothing* should be uncommented by default unless it's one of the important system config values like CODENVY_HOST.  So, please rethink the requirement description.

riuvshin:
default value should be changed here https://github.com/codenvy/codenvy/blob/master/dockerfiles/init/manifests/codenvy.pp#L216

bmicklea:
description updated

riuvshin:
PR proposed

riuvshin:
I think I can incude this in 5.1.2 bug fix release

JamesDrummond:
Should this be done for che as well?

riuvshin:
in che we don't have this ENV var afaik.

JamesDrummond:
#CHE_MACHINE_WS_AGENT_INACTIVE_STOP_TIMEOUT_MS=600000

bmicklea:
@riuvshin - where are the text comments that appear in the `codenvy.env` file stored?  I'd like to add a "Default = 14,400,000 MS (4 hours)" comment there too.

riuvshin:
hmm in che we have `CHE_MACHINE_WS_AGENT_INACTIVE_STOP_TIMEOUT_MS` but in che default value hardcoded in code we can't set default value with puppet.

riuvshin:
@bmicklea now better https://github.com/codenvy/codenvy/pull/1593/files#diff-4e1211b38190902dee42a5833ea2efb5 ?

PR #1593, opened 2017-01-24
increase default ws agent inactive timeout

riuvshin:
fixes https://github.com/codenvy/codenvy/issues/1592


riuvshin:
@james not sure I get what to change im not able find `idel` in text

PR #1594, opened 2017-01-24
updated to remove mention of beta

bmicklea:
### What does this PR do?
Removes mention of beta from notification on codenvy.io


bmicklea:
Spoke to @vkuznyetsov and this file will be removed altogether (not needed anymore). Closing the PR without merging.

PR #1596, opened 2017-01-24
CODENVY-1443: do not include project files into snapshot

### LINKED ISSUE #1443: Make Codenvy not to include projects of workspace into snapshot
When Codenvy run in containers it includes projects of workspace into snapshot of workspace. But they are also stored in projects backup storage.
This means that we can speed up creation of snapshots by excluding projects from snapshot somehow.
Since snapshotting doesn't respect .gitignore or any other rules to reduce snapshot size in can take a lot of time to snapshot workspace with big amount of binaries in target folders for example.

**Reproduction Steps:**

1. Create and start WS.
2. Create project.
3. Snapshot WS.
4. Take snapshot image id (for example by catching it from WS output on WS restoring)
5. Execute in system terminal `docker run <caught snapshot image> ls /projects/` 

**Expected behavior:**

Error that path doesn't exist.

**Observed behavior:**

Project folder is shown.

**Codenvy version:** 5.0.0-Snapshot
**OS and version:** Ubuntu 16.04

**Additional information:**

Problem only started happening recently, didn't happen in an older version of Codenvy: [No]
Problem can be reliably reproduced: [Yes]

garagatyi:
### What does this PR do?
Add non bind-mount volume to /projects folder to prevent commiting
it into snapshot.
Cleanup code and conf after removing native codenvy impl.
Add tests.

Fixes #1443 

### Previous Behavior
Projects files were included into snapshot

### New Behavior
Projects files are not included into snapshot

### Tests written?
Yes


plugins/plugin-hosted/codenvy-machine-hosted/src/main/java/com/codenvy/machine/agent/CodenvyInContainerInfrastructureProvisioner.java
```diff
@@ -36,38 +38,50 @@
...
                                         DockerExtConfBindingProvider.EXT_CHE_LOCAL_CONF_DIR);
+        devMachine.setEnvironment(environmentVars);
 
-        List<String> agents = envConfig.getMachines().get(devMachineName).getAgents();
+        ExtendedMachineImpl extendedMachine = envConfig.getMachines().get(devMachineName);
+        List<String> agents = new ArrayList<>(extendedMachine.getAgents());
         agents.add(agents.indexOf("org.eclipse.che.ws-agent"), "com.codenvy.rsync_in_machine");
+        extendedMachine.setAgents(agents);
+
+        ArrayList<String> volumes = new ArrayList<>(devMachine.getVolumes());
```
skabashnyuk:
```
List<String> volumes = new ArrayList<>(devMachine.getVolumes());
```

garagatyi: ok

Issue #1602, opened 2017-01-25
Restore script fails if ws-machine container without installed sudo.


AndrienkoAleksandr:
Restore script fails on images without sudo. It was reproduced on images library/fedora:24 and library/ubuntu:16.04. We need fix restore and backup scripts. There are a lot images without installed sudo.

Steps to reproduce: 
1. Create workspace from wsConfig: https://github.com/AndrienkoAleksandr/WSConfigRecipes/blob/master/Ubuntu16_04 or https://github.com/AndrienkoAleksandr/WSConfigRecipes/blob/master/Fedora24 .
2. Run workspace.
Expected result: worspace should be successfully launched.
Actual result: workspace start failed with error:
"Could not start workspace ubuntu_16_04. Reason: Start of environment 'default' failed. Error: Restoring of workspace workspaceveezyksxe000thv7 filesystem terminated on node2.nightly.codenvy-stg.com node. Process failed. Exit code 127"

Output Logs:
2017-01-25 08:54:19,039[ceSharedPool-12]  [INFO ] [o.e.c.a.w.s.WorkspaceRuntimes 664]   - Launching 'org.eclipse.che.terminal' agent at workspace workspace4h1218r2trt7e3du
2017-01-25 08:55:08,579[ceSharedPool-12]  [INFO ] [o.e.c.a.w.s.WorkspaceRuntimes 664]   - Launching 'org.eclipse.che.ssh' agent at workspace workspace4h1218r2trt7e3du
2017-01-25 08:55:14,855[ceSharedPool-12]  [INFO ] [o.e.c.a.w.s.WorkspaceRuntimes 664]   - Launching 'com.codenvy.rsync_in_machine' agent at workspace workspace4h1218r2trt7e3du
2017-01-25 08:55:19,402[ceSharedPool-12]  [ERROR] [DockerEnvironmentBackupManager 540]  - Error occurred during backup/restore on 'node1.nightly.codenvy-stg.com' : Warning: Permanently
bash: sudo: command not found
rsync: connection unexpectedly closed (0 bytes received so far) [sender]
rsync error: remote command not found (code 127) at io.c(226) [sender=3.1.2]
2017-01-25 08:55:19,403[ceSharedPool-12]  [ERROR] [DockerEnvironmentBackupManager 340]  - Process failed. Exit code 127
java.io.IOException: Process failed. Exit code 127
[...]

PR #1619, opened 2017-01-26
CODENVY-1602: Fix restore and backup scripts to add ability start workspaces from docker images without sudo.

AndrienkoAleksandr:
### What does this PR do?
Fix restore and backup scripts to add ability start workspaces from docker images without sudo.

https://github.com/codenvy/codenvy/issues/1602

### Docs requirements?
1.  API changes (No)
2.  User docs changes(No)  

@garagatyi @riuvshin @tolusha  review please.

Signed-off-by: Aleksandr Andrienko <aandrienko@codenvy.com>

### DIFF
--- a/dockerfiles/init/modules/codenvy/templates/rsyncbackup.sh.erb
+++ b/dockerfiles/init/modules/codenvy/templates/rsyncbackup.sh.erb
@@ -25,9 +25,12 @@ if [[ $(stat -c %a /opt/rsync_key) != 600 ]]; then
     chmod 0600 /opt/rsync_key
 fi
 
+RSYNC_COMMAND=$(ssh -i /opt/rsync_key -l ${SRC_USER_NAME} -o StrictHostKeyChecking=no -p ${SRC_PORT} ${SRC_HOST} \
+            "if hash sudo 2>/dev/null;then echo 'sudo rsync';else echo 'rsync';fi")
+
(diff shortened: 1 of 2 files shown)


dockerfiles/init/modules/codenvy/templates/rsyncbackup.sh.erb
```diff
@@ -25,9 +25,12 @@ if [[ $(stat -c %a /opt/rsync_key) != 600 ]]; then
     chmod 0600 /opt/rsync_key
 fi
 
+RSYNC_PATH=$(ssh -i /opt/rsync_key -l ${SRC_USER_NAME} -o StrictHostKeyChecking=no -p ${SRC_PORT} ${SRC_HOST} \
```
riuvshin: maybe we have to rename $RSYNC_PATH to something else to make it more obvious to read, because RSYNC_PATH is not actually path to rsync you need this only to detect run with sudo or not. but in general looks good

AndrienkoAleksandr: Maybe RSYNC_COMMAND would be better?

riuvshin: RSYNC_COMMAND looks better for me @garagatyi wdyt?

garagatyi: +1

Issue #1572, opened 2017-01-20
Codenvy can't access registry if deployed on AWS and DNS uses external IP


skryzhny:
When deployed on AWS and DNS pointing to external IP codenvy tomcat can't access registry used for storing snapshots.
So it can't delete unused old snapshots after new ones pushed to registry.
Need to introduce new property used for communication between codenvy and registry containers, by default it should be name of registry container.

**Reproduction Steps:**
Lanch dockerized codenvy 5.1.0 on AWS and use external IP of instance in DNS
See errors in codenvy logs about inability connect to registry.
```
2017-01-18 20:18:41,638[ceSharedPool-14]  [ERROR] [c.p.d.m.DockerInstanceProvider 177]  - connect timed out
java.net.SocketTimeoutException: connect timed out
        at java.net.PlainSocketImpl.socketConnect(Native Method) ~[na:1.8.0_102]
        at java.net.AbstractPlainSocketImpl.doConnect(AbstractPlainSocketImpl.java:350) ~[na:1.8.0_102]
        at java.net.AbstractPlainSocketImpl.connectToAddress(AbstractPlainSocketImpl.java:206) ~[na:1.8.0_102]
        at java.net.AbstractPlainSocketImpl.connect(AbstractPlainSocketImpl.java:188) ~[na:1.8.0_102]
        at java.net.SocksSocketImpl.connect(SocksSocketImpl.java:392) ~[na:1.8.0_102]
        at java.net.Socket.connect(Socket.java:589) ~[na:1.8.0_102]
        at sun.net.NetworkClient.doConnect(NetworkClient.java:175) ~[na:1.8.0_102]
        at sun.net.www.http.HttpClient.openServer(HttpClient.java:432) ~[na:1.8.0_102]
        at sun.net.www.http.HttpClient.openServer(HttpClient.java:527) ~[na:1.8.0_102]
        at sun.net.www.http.HttpClient.<init>(HttpClient.java:211) ~[na:1.8.0_102]
        at sun.net.www.http.HttpClient.New(HttpClient.java:308) ~[na:1.8.0_102]
        at sun.net.www.http.HttpClient.New(HttpClient.java:326) ~[na:1.8.0_102]
        at sun.net.www.protocol.http.HttpURLConnection.getNewHttpClient(HttpURLConnection.java:1169) ~[na:1.8.0_102]
[...]

PR #1625, opened 2017-01-26
add possibility to set extra hosts for codenvy containers

riuvshin:
### What does this PR do?
this PR add possibility to set extra hosts rules to codenvy containers

https://github.com/codenvy/codenvy/issues/1572

@TylerJewell please check language in codenvy.env

@bmicklea @JamesDrummond I need your help guys if we need add something about this in docs.


dockerfiles/init/manifests/codenvy.env
```diff
@@ -473,3 +473,14 @@
...
 #     manage it with the CLI.
 #CODENVY_SWARM_NODES=172.17.0.1:2375
+
+
+############################################################
+#####                                                  #####
+#####                CODENVY EXTRA HOSTS               #####
+#####                                                  #####
+# Extra hosts
+#     A comma separated list of <DNS>:<IP> combinations which will be added to /etc/hosts file of
```
benoitf: i would replace dns by host or hostname

riuvshin: I was also thinking about that

riuvshin: done

dockerfiles/init/manifests/codenvy.env
```diff
@@ -473,3 +473,14 @@
...
 #CODENVY_SWARM_NODES=172.17.0.1:2375
+
+
+############################################################
+#####                                                  #####
+#####                CODENVY EXTRA HOSTS               #####
+#####                                                  #####
+# Extra hosts
+#     A comma separated list of <HOST>:<IP> combinations which will be added to /etc/hosts file of
+#     each codenvy container. This could be useful for same cases where codenvy system should be aware
```
garagatyi: each codenvy container here can lead to user's believing that it will be used in workspaces too.

dockerfiles/init/manifests/codenvy.pp
```diff
@@ -305,6 +305,10 @@
   $codenvy_repo = getValue("CHE_REPO","off")
 
 ###############################
+# Codenvy developmet mode
```
garagatyi: Can you elaborate on why it is related to development mode? I don't understand.

dockerfiles/init/modules/compose/templates/docker-compose.yml.erb
```diff
@@ -112,6 +136,12 @@ services:
       options:
         syslog-address: "tcp://localhost:5140"
         syslog-facility: "local3"
+<% if scope.lookupvar('compose::codenvy_extra_hosts') != 'NULL' -%>
+    extra_hosts:
+<% scope.lookupvar('compose::codenvy_extra_hosts').split(",").each do |val| -%>
+<%='      - "' + val + '"' %>
+<% end -%>
+<% end -%>
```
benoitf: if it's possible also to use some sort of macros to avoid copy/paste of the same code on each container

dockerfiles/init/manifests/codenvy.env
skryzhny: I prefer use HOSTNAME instead of HOST

dockerfiles/init/manifests/codenvy.pp
benoitf: BTW there is a typo on "developmet"

TylerJewell:
I went through the entire codenvy.env template and cleaned it up.  We had some duplicates with the new EXTRA HOSTS

benoitf:
@TylerJewell AFAIK these extra hosts are not part of the "Workspace Config"

TylerJewell:
The description from roman is that the workspace hosts are added to every container of each workspace. So from an end user point of view, the perception is that they are configuring their workspaces.  I understand that at a technical level it is something else, but I am thinking of it at a perception level.

benoitf:
Even at perception level, it will configure hosts in codenvy "system" containers (the one defined in docker-compose file), not the codenvy "user" containers (like workspaces)

TylerJewell:
Then let me rewrite that and move it to the system level.

benoitf:
ok I think content has been moved but not updated
+ # each machine and container. This could be useful for 
 +#     cases where workspaces should be aware of DNS names that are not resolvable
 +#     through traditional or common way.

because it doesn't target workspaces and it matches only codenvy system containers.

riuvshin:
can we have have othet changes in codenvy.env as a separate PR? Im not able to figure out now what is what.  those extra hosts is ONLY for codenvy containers NOT workspaces.

TylerJewell:
I don't want to try and pull this apart now.  It took a half hour and had a lot of changes.

@benoitf - just a little confused from the last statement.  Are extra hosts applied to all containers, or just system containers?  I think the language that what was written assumes we apply it to every container that we create, both system and workspace.

benoitf:
@TylerJewell yes it was confusing as reported by @garagatyi but AFAIK it will only be set in the docker-compose containers.

TylerJewell:
Hmmm that seems weird, let's get on phone with Alex and Roman to discuss.  As I think we have wires crossed.

benoitf:
it would be nice to split PRs in two parts (or two commits)
- one that is introducing new property
- one that is updating properties

riuvshin:
@benoitf I've updated template to avoid of copy paste. I will split that PR in two because it also hard for me to deal with different changes

riuvshin:
this will be another PR, I don;t like mess in pull requests same as @benoitf and for me it is important to separate changes. to properly review them.

@garagatyi @benoitf  can you please review changes again

bmicklea:
@JamesDrummond - did this get added to Codenvy docs?

JamesDrummond:
Not yet. Added it to my agenda.

PR #1632, opened 2017-01-27
improve addon to have own modules

riuvshin:
needed for https://github.com/codenvy/customer-saas/issues/126
those changes allows addon.pp have module which allow not only set variables but create new files if needed


riuvshin:
@benoitf brain tree was here only because we didn't have a way to create it only for saas.

PR #1637, opened 2017-01-29
Add postflight check of swarm nodes

TylerJewell:
Signed-off-by: Tyler Jewell <tjewell@codenvy.com>

### What does this PR do?
This PR adapts changes made to Eclipse Che to allow the inherited CLI to inherit its `usage()` statement from the core CLI so that as the base CLI makes improvements, the new CLI will have it inherited automatically.

Also provides a first implementation of a postflight check, which pulls the list of Swarm nodes from `codenvy.env` and then uses `curl` to ping each one's `/info` URL to see if it is reachable.

Must merge this at the same time as we merge: https://github.com/eclipse/che/pull/3940

### Changelog
Adapt CLI to changes in Eclipse Che #3940
Add postflight check of Swarm nodes as a test

### Release Notes
We are working to make sure Codenvy starts in any environment perfectly every time. We've added some additional preflight checks related to mandatory minimums for disk and memory available from the host. If the minimums are not enough to start at least one workspace, then Codenvy bootup will be interrupted. Preflight checks can now be skipped with either `--fast` or `--skip:preflight` on the command line.

We have also introduced postflight checks. We parse `codenvy.env` to get a list of Swarm nodes for your cluster. We then check that each is reachable using curl and that their `/info` URL is returning valid data. A failure of this command is an early indicator that firewalls were not configured correctly.  You can skip this check with `--fast` or `--skip:postflight`.

### Docs PR
https://github.com/codenvy/codenvy-docs/pull/59

### DIFF
--- a/dockerfiles/cli/scripts/cli.sh
+++ b/dockerfiles/cli/scripts/cli.sh
@@ -6,7 +6,7 @@
 # http://www.eclipse.org/legal/epl-v10.html
 #
 
-cli_post_init() {
+post_init() {
   GLOBAL_HOST_IP=${GLOBAL_HOST_IP:=$(docker_run --net host ${UTILITY_IMAGE_CHEIP})}
   DEFAULT_CODENVY_HOST=$GLOBAL_HOST_IP
   CODENVY_HOST=${CODENVY_HOST:-${DEFAULT_CODENVY_HOST}}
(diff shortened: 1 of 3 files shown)


TylerJewell:
@benoitf - as part of the testing, I would ask for you to adapt ARTIK and Codenvy SAAS to do the same.  Note that there was a small refactoring of the method names and how I make use of pre_init and post_init in Che.

benoitf:
I've updated methods name

PR #1660, opened 2017-01-31
Allow to call LDAP synchronization from REST API only for users with MANAGE_CODENVY permission

evoevodin:
Added binding for `LdapSynchronizerPermissionsFilter` it seems that it's missing.
Moreover LDAP authentication tests are reworked, authentication is applied on the same directory structure which allows to clearly see the difference between different authentication types.

#### Changelog
Only users with _MANAGE_CODENVY_ permissions can call LDAP synchronization from REST API.

#### Release Notes
N/A for bugs

#### Docs PR
https://github.com/codenvy/codenvy-docs/pull/65


TylerJewell:
Missing changelog, release notes and docs reference.

evoevodin:
Changelog added

evoevodin:
@bmicklea can you verify that changelog is ok?

bmicklea:
I'm not really clear on this one. The title and issue description seem very different than the changelog. If this is an enhancement (I think it is) please describe how the functionality has changed.

evoevodin:
I restructured title and description a bit, so it's more like security improvement right now

bmicklea:
Tweaked the changelog entry - won't this require changes to the ldap admin docs page?

evoevodin:
Up to you, the section https://codenvy.com/docs/admin-guide/ldap/index.html#synchronization-strategy
describes API method without any restrictions, i can add a _NOTE_ that only the user with _MANAGE_CODENVY_ permission can call the method, if it makes sense to you.

PR #1661, opened 2017-01-31
pick prop value depending on scope

riuvshin:
to be able to override default value of `db_schema_flyway_scripts_locations` for saas we need pick var value depending on scope where it is declared.
this approach needed for some rare cases where we want override default value, it was possible to change it by modifying `codenvy.env` but for saas that prop must be always different.

### Changelog and Release Note Information
**Changelog**: set `db_schema_flyway_scripts_locations` property value depending on scope.


TylerJewell:
Add Changelog entry, please.

riuvshin:
@TylerJewell those changes are hidden from a user, it is internal stuff, does it still require changelog?

TylerJewell:
Yes, every PR always requires a 1-line changelog, even for minor internal improvements.  It's a changelog for maintainers as much as end users.

riuvshin:
@TylerJewell done

PR #1680, opened 2017-02-02
Don't backup workspace if restore is failed

### LINKED ISSUE #1663: Launch backup of workspaces should be disabled if restoration of workspace was failed
Analysing logs from https://github.com/codenvy/codenvy/issues/1662 I found that:
After we failed to restore projects from backup we still trying to synchronize projects back to backup storage.
This may lead to situation when we effectively erasing backup.
Possible steps:
- Start workspace
- Restoration job failed  - /projects in BAD, inconsistent state.
- Stop of workspace
- Backup procedure starts to synchronise /projects. So GOOD backup from backup storage is replaced with BAD one from workspace

**We should not launch backup procedure if previous restoration procedure was failed.**
Maybe we need to have separate API method that removes workspace without attempting to backup projects.

**Reproduction Steps:**
Don't know

**Codenvy version:**   5.1.2

evoevodin:
### What does this PR do?

Fixes #1663.
If _unbindWorkspace_ is called before _bindWorkspace_ completed or called,
then warning is logged and unbind(backup) is not performed. So basically
if workspace start is failed or interrupted no one overrides old binaries.

#### Changelog
Ensures workspaces are properly backed up even if a subsequent workspace start fails.

#### Release Notes
N/A for bugs

#### Docs PR
N/A for bugs

### DIFF
--- a/plugins/plugin-hosted/codenvy-machine-hosted/src/main/java/com/codenvy/machine/RemoteDockerNode.java
+++ b/plugins/plugin-hosted/codenvy-machine-hosted/src/main/java/com/codenvy/machine/RemoteDockerNode.java
@@ -36,10 +36,11 @@ import static org.slf4j.LoggerFactory.getLogger;
  * REST client for remote machine node
  *
  * @author Alexander Garagatyi
+ * @author Yevhenii Voevodin
  */
 public class RemoteDockerNode implements DockerNode {
-    private static final Logger  LOG                  = getLogger(RemoteDockerNode.class);
-    private static final Pattern NODE_ADDRESS         = Pattern.compile(
+    private static final Logger  LOG          = getLogger(RemoteDockerNode.class);
+    private static final Pattern NODE_ADDRESS = Pattern.compile(
             "((?<protocol>[a-zA-Z])://)?" +
             // http://stackoverflow.com/questions/106179/regular-expression-to-match-dns-hostname-or-ip-address
             "(?<host>(([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\\-]*[a-zA-Z0-9])\\.)*([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\\-]*[A-Za-z0-9]))" +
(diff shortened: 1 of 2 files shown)


plugins/plugin-hosted/codenvy-machine-hosted/src/main/java/com/codenvy/machine/RemoteDockerNode.java
```diff
@@ -51,6 +52,10 @@
     private final String                         nodeHost;
     private final String                         nodeIp;
 
+    private volatile boolean isBound;
```
vzhukovskii: What about using an enum here instead of operating with boolean flags?

garagatyi: This is definitely flag. Can you elaborate on what are the pros in usage of enum here?

evoevodin: This is not something exposed outside and needed for a single method, so i would stick to current approach

vzhukovskii: It won't be exposed as API. Operating with one enumerated status variable will be better for code readability and understanding. And in the same time it will reduce mistakes made in future. And what about cohesion, those two methods will be connected using one variable instead of three.

evoevodin: i've just changed the logic to use a single flag, so keep calm, nothing to argue about

vzhukovskii: Okay, I see, question has been resolved with it, thanks.

evoevodin: But yes, enum based implementation would be better i think

evoevodin:
@bmicklea can you please verify that changelog is okay

evoevodin:
@JamesDrummond can you please verify that the changelog is okay

evoevodin:
Thank you

PR #1684, opened 2017-02-03
CHE-3811: che-lib:use che-orion-editor

vparfonov:
### What does this PR do?

Move all JS files of Orion from plugin-orion to the che-lib repository. This allow as avoid problem with compilation plugin-orion on Windows OS because during compilation we modify  some JS files with using patch utils which not available on Windows OS machine. 
So for now we move this step to the che-lib, by default our contributors don't need modify this code. 

Related to: https://github.com/eclipse/che-lib/pull/18

#3811 

### Changelog:

Move all JS files of Orion from plugin-orion to the che-lib repository


assembly/onpremises-ide-compiling-war-ide/pom.xml
```diff
@@ -152,6 +152,12 @@
             <groupId>org.eclipse.che.plugin</groupId>
             <artifactId>che-java-testing-testng-ide</artifactId>
         </dependency>
+
```
skabashnyuk: sort pom

PR #1692, opened 2017-02-03
OpenJDK instead of Oracle JDK

eivantsov:
@TylerJewell @riuvshin @skabashnyuk  I have not tested it with an actual installation, just verified that Java is installed and added to path:

```
/ # java -version
openjdk version "1.8.0_111"
OpenJDK Runtime Environment (IcedTea 3.2.0) (Alpine 8.111.14-r1)
OpenJDK 64-Bit Server VM (build 25.111-b14, mixed mode)

```


eivantsov:
I will test and then merge. Please hold on with merging.

eivantsov:
Tested. Codenvy started. No issues found.

dockerfiles/init/modules/codenvy/templates/codenvy.env.erb
```diff
@@ -22,7 +22,7 @@ JPDA_ADDRESS=<%= scope.lookupvar('codenvy::codenvy_debug_port') %>
 <% if scope.lookupvar('codenvy::codenvy_debug_suspend') == "true" %>JPDA_SUSPEND=y<% end -%>
 <% end %>
 
-JAVA_HOME=/opt/jdk
+JAVA_HOME=/usr/lib/jvm/default-jvm/jre
```
riuvshin: @eivantsov @skabashnyuk maybe we should change JAVA_HOME to JDK not JRE?
like `JAVA_HOME=/usr/lib/jvm/java-1.8-openjdk/bin` I know that it starts with JRE but im not sure if we will have all options with JRE with debugging and some JVM OPTS that we set-up. WDYT guys?

dockerfiles/codenvy/Dockerfile
```diff
@@ -15,84 +15,25 @@
...
 
-ENV JAVA_VERSION_MAJOR=8 \
-    JAVA_VERSION_MINOR=102 \
-    JAVA_VERSION_BUILD=14 \
-    JAVA_PACKAGE=server-jre \
-    JAVA_JCE=standard \
-    JAVA_HOME=/opt/jdk \
-    PATH=${PATH}:/opt/jdk/bin \
-    GLIBC_VERSION=2.23-r3 \
+ENV JAVA_HOME=/usr/lib/jvm/default-jvm/jre \
```
riuvshin: same question as here https://github.com/codenvy/codenvy/pull/1692/files#r99466225

riuvshin:
Btw @eivantsov we need exactly same changes for saas here https://github.com/codenvy/customer-saas/blob/master/dockerfiles/saas/Dockerfile

dockerfiles/init/modules/codenvy/templates/codenvy.env.erb
mshaposhnik: +1 maven server and dugging may fail. Let's configure JDK for now.

riuvshin:
we need to revert those changes, we have issues with openJDK

PR #1701, opened 2017-02-04
Adapt CLI for refactoring of CHE-4006

TylerJewell:
Signed-off-by: Tyler Jewell <tjewell@codenvy.com>

### What does this PR do?
Applies refactoring changes to support Che CLI refactoring including updated image names, CLI API check, and different hook method names.

https://github.com/eclipse/che/pull/4006

#### Changelog
Adapt CLI for CHE-4006 refactoring
Change `stop --force` to be `stop --skip:graceful`

#### Release Notes
N/A

#### Docs PR
N/A


dockerfiles/cli/scripts/cmd_init.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
+# Copyright (c) 2016 Codenvy, S.A.
+# All rights reserved. This program and the accompanying materials
+# are made available under the terms of the Eclipse Public License v1.0
+# which accompanies this distribution, and is available at
+# http://www.eclipse.org/legal/epl-v10.html
+#
+
+
+cmd_init_reinit_pre_action() {
```
benoitf: I think we had a typo there (reinit vs preinit) and it wasn't an issue as typo was in the hook name as well but now that the pre/post hooks are automated, is it an issue ?

benoitf:
I think that we could probably get rid of the cli entrypoint and use the entrypoint defined in the base image
This will allow to be sure that it's the "base" stuff that loads the "cli" stuff and not cli stuff that call "base" stuff to then init the cli stuff. (in short : we can then be sure that base script can evolve and check what is defined in the cli image without any breakage)

TylerJewell:
You are correct.  I moved it into the base and it works now.  I tested Che and Codenvy.  Good idea.

dockerfiles/cli/scripts/cmd_init.sh
TylerJewell: It's not a typo - I checked it a couple times.  It's a special hook that is only called if you are *not* reiniting the system.

PR #1702, opened 2017-02-06
Update CHANGELOG.md

JamesDrummond:
Updated per new changelog standard. Only PR's have changelogs.


JamesDrummond:
@bmicklea Please review changes.

bmicklea:
@JamesDrummond I think it's better as just the last bulleted item in the list rather than a separate bold line.

PR #1745, opened 2017-02-09
PG configuration according http://dalibo.github.io/pgbadger/ 

skabashnyuk:
### What does this PR do?

I would like to change postgresql configuration to be able to use pgbadger to analyse postgresql state.
Example of report
https://transfer.sh/itzox/out.html
http://dalibo.github.io/pgbadger/samplev7.html

Also we might think about set ``` log_min_duration_statement = 0```  I think now we already doing that by ```log_statement = 'all'```

#### Changelog
Change postgresql configuration to be able to use pgbadger to generate usage reports

### DIFF
--- a/dockerfiles/init/modules/postgres/templates/postgresql.conf.erb
+++ b/dockerfiles/init/modules/postgres/templates/postgresql.conf.erb
@@ -415,9 +415,10 @@ log_min_duration_statement = 0
 log_checkpoints = on
 log_connections = on
 log_disconnections = on
-#log_duration = off
+log_duration = off
 log_error_verbosity = default		# terse, default, or verbose messages
 #log_hostname = off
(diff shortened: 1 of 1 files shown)


skabashnyuk:
@codenvy/pm  please review

JamesDrummond:
Will this be used only by us for saas or do we think customers/users will use this too? Just checking to see if documentation would be helpful to customer/users to know how to produce this usage report.

@bmicklea Would the usage report data charts be useful to marketing?

bmicklea:
This is very low-level information. I don't think exposing it to users is a
good idea as it will only tempt them to try and tune things without a good
understanding of the impacts.

No on docs or release notes.  Changelog only.
-- 

Brad Micklea | Codenvy | bmicklea@​codenvy.​com | 4​16​.7​07​.07​92

PR #1764, opened 2017-02-11
WIP - [cli] Clearer  networking configuration & CLI override for CODENVY_HOST

TylerJewell:
DO NOT MERGE WITHOUT REVIEW

### What does this PR do?
Adds in additional content to the `codenvy.env` file that outlines clear instructions on our network protocol discovery and progression for various connections that must be established.  This copies the content that we have within the Che configuration. 

We need this to be carefully reviewed as I am sure that some of the properties in this file are relevant to Che, but absolutely not relevant for Codenvy.  We need to make sure that the Codenvy version is accurate.  Once this is reviewed, I will open up a PR for codenvy documentation to make this clear as well.

Also adds in changes needed to adapt to https://github.com/eclipse/che/pull/4114, so that CODENVY_HOST provided on command line will override value in `codenvy.env`.

#### Changelog
Add networking progression and configuration to `codenvy.env`
CLI override for `CODENVY_HOST` variable.

#### Release Notes
N/A

#### Docs PR
TBD


dockerfiles/init/modules/codenvy/templates/codenvy.env.erb
```diff
@@ -1,3 +1,11 @@
+# Include all custom CODENVY_* properties that user may has defined in its codenvy.env file
+<% ENV.each do |k,v| -%>
+<% if k.include? "CODENVY_" then  -%>
```
garagatyi: I'm curious whether it work with Saas packaging also?

dockerfiles/init/manifests/codenvy.env
```diff
@@ -86,17 +86,121 @@
...
+#
+# Affects how the Codenvy server connects to the Docker Daemon, how the Che server and
+# workspaces establish connections to one another, and also how remote browser clients
+# should discover and connect to workspaces.
+#
+# Codenvy goes through a progression algorithm to establish the protocol, IP address and
+# port to establish communications with it is booting or starting a workspace.
+#
+# Browser --> Codenvy Server
+#    1. Default is 'http://localhost:${SERVER_PORT}/wsmaster/api'.
```
garagatyi: I think default is taken from CODENVY_HOST

dockerfiles/init/manifests/codenvy.env
```diff
@@ -86,17 +86,121 @@
...
+#    2. Else use the value of CODENVY_API
```
garagatyi: I'm not aware of CODENVY_API var. Did you mean CODENVY_HOST?

garagatyi: I would prefer to put rule that some value is taken from **var**, and default value in case that var is not set next **val**.

dockerfiles/init/manifests/codenvy.env
```diff
@@ -86,17 +86,121 @@
...
+#
+# Codenvy Server --> Docker Daemon Progression:
```
garagatyi: Codenvy server connects to swarm, not docker daemon. We always know where swarm is because we run container with it.
Should we allow user to rewrite Swarm API location?
On the other hand Codenvy connects to several docker daemons via Swarm. Probably we should describe here how it works, which property contains list of nodes.

dockerfiles/init/manifests/codenvy.env
```diff
@@ -86,17 +86,121 @@
...
+#    1. Use the value of CODENVY_DOCKER_DAEMON__URL
+#    2. Else, use the value of DOCKER_HOST system variable
+#    3. Else, use Unix socket over unix:///var/run/docker.sock
+#
+# Codenvy Server --> Workspace Connection:
+#    - If CODENVY_DOCKER_SERVER__EVALUATION__STRATEGY is 'default':
```
garagatyi: We use server evaluation strategy "codenvy". I'm not sure whether it makes sense to allow user to change implementation unless he knows internals of the product. Also there are other places that influence server -> ws communication. For example our url transformer that converts URLs using patterns provided in property. Also there is behavior influenced by host OS, and whether some property set.
@benoitf probably knows better since he added that code into codebase.

dockerfiles/init/manifests/codenvy.env
```diff
@@ -86,17 +86,121 @@
...
+#        2. Else, if server connects over Unix socket, then use localhost
+#        3. Else, use DOCKER_HOST
+#    - If CODENVY_DOCKER_SERVER__EVALUATION__STRATEGY is 'docker-local':
+#        1. Use the address of the workspace container within the docker network 
+#           and exposed ports
+#        2. If address is missing, if server connects over Unix socket, then use 
+#           localhost and exposed ports
+#        3. Else, use DOCKER_HOST and published ports
+#
+# Browser --> Workspace Connection:
```
garagatyi: The [same](https://github.com/codenvy/codenvy/pull/1764/files#r100690553)

dockerfiles/init/manifests/codenvy.env
```diff
@@ -86,17 +86,121 @@
...
+#    - If CODENVY_DOCKER_SERVER__EVALUATION__STRATEGY is 'docker-local':
+#        1. If set use the value of CODENVY_DOCKER_IP_EXTERNAL
+#        2. Else use the address of the workspace container within the docker network, 
+#           if it is set
+#        3. If address is missing, if server connects over Unix socket, then use 
+#           localhost
+#        4. Else, use DOCKER_HOST
+#
+# Workspace Agent --> Codenvy Server
+#    1. Default is 'http://codenvy-host:${SERVER_PORT}/wsmaster/api', where
```
garagatyi: I think we use value of CODENVY_HOST, not codenvy-host.
Also protocol depends on installation settings.

dockerfiles/init/manifests/codenvy.env
```diff
@@ -86,17 +86,121 @@
...
+#        1. If set use the value of CODENVY_DOCKER_IP_EXTERNAL
+#        2. Else use the address of the workspace container within the docker network, 
+#           if it is set
+#        3. If address is missing, if server connects over Unix socket, then use 
+#           localhost
+#        4. Else, use DOCKER_HOST
+#
+# Workspace Agent --> Codenvy Server
+#    1. Default is 'http://codenvy-host:${SERVER_PORT}/wsmaster/api', where 
+#       'codenvy-host' is IP of server.
```
garagatyi: codenvy-host (CODENVY_HOST) can be hostname, not only IP

dockerfiles/init/manifests/codenvy.env
```diff
@@ -86,17 +86,121 @@
...
+#    2. Else, use value of CODENVY_WORKSPACE_CHE__SERVER__ENDPOINT
```
garagatyi: Please consider changing language of sequence of options. It is not clear,  in my opinion,  when option works and when not.

dockerfiles/init/manifests/codenvy.env
```diff
@@ -86,17 +86,121 @@
...
+#    3. Else, if 'docker0' interface is unreachable, then 'codenvy-host' replaced 
+#       with 172.17.42.1 or 192.168.99.1
+#    4. Else, print connection exception
+
+# Codenvy Server API Endpoint
+#     The location of the API end point where dashboard and IDE clients will look for
+#     interacting with the Codenvy server, which we also call the workspace master.
+#CODENVY_API=http://localhost:${SERVER_PORT}/wsmaster/api
```
garagatyi: Are you sure that we have this property? Can you elaborate on when you found it.

dockerfiles/init/manifests/codenvy.env
```diff
@@ -86,17 +86,121 @@
...
+
+# Docker Daemon URL
+#     How the Codenvy server will discover the location of the Docker daemon. If this is
```
garagatyi: See my comment about swarm above.

dockerfiles/init/manifests/codenvy.env
```diff
@@ -86,17 +86,121 @@
...
+# Workspace Address Resolution Strategy
+#     The strategy to determine how workspace servers users launch will be exposed. 
+#     If the server and your workspaces are on different networks that flow through a
+#     firewall, those firewalls can prevent communications between Codenvy and workspaces.
+#     The default strategy exposes workspaces with the IP of Docker and ephemeral ports.
+#     A Docker local strategy uses internal Docker network IP addresses and exposed ports.
+#
+#     Options:
+#       - 'default':      internal address is DOCKER_HOST with ephemeral ports
+#       - 'docker-local': internal address is Docker network container address with
```
garagatyi: Codenvy uses another approach. It can depends on url transformer property but by default we proxy all request using our HAProxy container.

dockerfiles/cli/scripts/cmd_config.sh
```diff
@@ -95,6 +94,17 @@ generate_configuration_with_puppet() {
     fi
   fi
 
+  for element in "${CLI_ENV_ARRAY[@]}" 
+  do
+    var1=$(echo $element | cut -f1 -d=)
+    var2=$(echo $element | cut -f2 -d=)
+
+    if [[ $var1 == CHE_* ]] ||
+       [[ $var1 == ${CHE_PRODUCT_NAME}_* ]]; then
```
TylerJewell: It does not.  In the case of ARTIK it would be ARTIK_*.  And in the case of Codenvy and Codenvy SAAS the would both be CODENVY_*.

dockerfiles/init/modules/codenvy/templates/codenvy.env.erb
TylerJewell: Yes, it does as well.  The SAAS CLI packaging inherits directly from the Codenvy CLI packaging.  In fact, the `/init` portion of SAAS is a (mostly) perfect duplicate of the same in Codenvy.

dockerfiles/init/manifests/codenvy.env
TylerJewell: Can you verify this?  If so, will remove.  I suspect that the networking evaluation cycle for Codenvy is quite different than che, and this is just che's cycle repeated for codenvy.

dockerfiles/init/manifests/codenvy.env
TylerJewell: No, I meant CODENVY_API.  In Che, we actually have a CHE_API= variable that is the URL of the entrypoint.  I suspect that with Codenvy no such variable exists, as it's internally coded and not something we should expose.

in the case of CODENVY, we have the reverse proxy, so it probably is CODENVY_HOST, but I need someone to tell me what actually happens and then can update content.

dockerfiles/init/manifests/codenvy.env
TylerJewell: Thanks - updated to make the SWARM reference and use SWARM variables.

dockerfiles/init/manifests/codenvy.env
TylerJewell: In the case of Che, users / admins can configure certain variables because it's not always the case that they can get Che -> Workspace to connect.  So they have to configure variables to make this connection work.

If in Codenvy there is *nothing* that an end user admin will need to configure and that we guarantee that Codenvy -> Worskpace can always make a connection, then we do not have to disclose the strategy and can just call it internal.  But if there are any variables that affect how Codenvy discovers a ws, then I need to document it.

dockerfiles/init/manifests/codenvy.env
TylerJewell: Are you sure?  There is nothing an admin ever needs to configure that allows a browser to reach a workspace other than CODENVY_HOST?  In Che, we have CHE_DOCKER_EXTERNAL_IP as an extra parameter that tells browser how to reach Codenvy.  It doesn't have to be set all the time, but sometimes.

dockerfiles/init/manifests/codenvy.env
TylerJewell: ok

dockerfiles/init/manifests/codenvy.env
TylerJewell: ok

dockerfiles/init/manifests/codenvy.env
TylerJewell: You tell me.  These are just copies over from Che.  I am hoping you tell me which properties are used by Codenvy and which ones are not.  There is a CHE_API for Che.  It's very possible it does not exist, but need engineers to tell me.

dockerfiles/init/manifests/codenvy.env
TylerJewell: Not sure I understand the confusion.

TylerJewell:
At this point, there are open questions as to:
a: Do these variables exist for codenvy?
b: Are these variables ones that we want admins to configure (yes / no)?

I need an engineer to help a bit. They all exist within Che, but it's not clear if they also exist for Codenvy.

#CODENVY_API=http://localhost:${SERVER_PORT}/wsmaster/api
#CODENVY_DOCKER_DAEMON__URL=NULL.
#DOCKER_HOST=tcp://localhost:2375 
#CODENVY_DOCKER_IP=172.17.0.1
#CODENVY_DOCKER_IP_EXTERNAL=NULL

dockerfiles/cli/scripts/cmd_config.sh
garagatyi: ok, thx

dockerfiles/init/manifests/codenvy.env
garagatyi: Well in this case che doesn't differ from codenvy much. CLI prints che/codenvy url for access from browser. So I think that it is valid to say that browser connects to master by this url.

dockerfiles/init/manifests/codenvy.env
garagatyi: I believe that for codenvy it should be CODENVY_HOST.

dockerfiles/init/manifests/codenvy.env
garagatyi: For terminal and ws-agent next properties influence connection machine.proxy_wsagent_server_url_template
machine.proxy_terminal_server_url_template

Strategy should not be changed since we have only 1 strategy designed to work in codenvy.
Also there is some host dependent logic influenced by some variables, but it is better to ask Florent about that.

dockerfiles/init/manifests/codenvy.env
garagatyi: Answer it depends. Florent knows better since he is the author of these changes.

dockerfiles/init/manifests/codenvy.env
garagatyi: Starting from default is confusing to me. I would prefer to read docs that says:
1. If <> is set then
1.1. If <> is ....
1.2 Otherwise ...
2. Otherwise default value <> is used.

garagatyi: But I suppose it is just a matter of taste.

dockerfiles/init/manifests/codenvy.env
garagatyi: I don't remember seeing it before.

TylerJewell: ok - removed.

TylerJewell:
I believe I have updated everything according to alex's guidelines.  So I believe this is ready to merge.

dockerfiles/init/manifests/codenvy.env
```diff
@@ -106,6 +127,21 @@
 #     configure this as we manage it with the CLI.
 #CODENVY_SWARM_NODES=172.17.0.1:2375
 
+# Docker Host
+#     How Codenvy will connect to the Docker host if CODENVY_DOCKER_DAEMON__URL is not 
+#     set. This value can be provided on the command line as well. If this value is
+#     not set, then Codenvy will use the volume mount of unix:///var/run/docker.sock.
+#DOCKER_HOST=tcp://localhost:2375
```
garagatyi: I'm not sure how this works in CLI, so let @TylerJewell  and @benoitf check it

dockerfiles/init/manifests/codenvy.env
```diff
@@ -106,6 +127,21 @@
...
+
+# External Docker IP Address
+#     The external IP address used by browsers to connect to workspace containers that
+#     are running on a Docker network. After Codenvy has created a workspace runtime,
+#     those Docker containers must be reachable by remote browsers. B 
+#     This will generally work if the browser and the workspace are on the same network.
+#     However, if the browser and the workspaces are on different networks, then you
+#     may need to set this value to the external IP address of your host or Docker to
+#     make workspaces reachable.
+#CHE_DOCKER_IP_EXTERNAL=NULL
```
garagatyi: Maybe it is needed to mention that it works with single node only? Also As far as I can see it only works if codenvy is running on the same host as WSs

dockerfiles/init/modules/codenvy/templates/codenvy.env.erb
riuvshin: In codenvy we configure everything with properties, I think this is bad idea to use props and ENV vars in the same time.

TylerJewell: Yes, I agree we can remove this logic from this PR.

riuvshin:
I've updated PR

PR #1766, opened 2017-02-13
INFRA-68; remove compiling-ide module

mshaposhnik:
### What does this PR do?
Simplifies packaging by removing separate ide-compiling module and move everything into assembly-ide-war.

https://github.com/codenvy/infrastructure/issues/68

#### Changelog
Simplified packaging by removing compiling ide assembly module.

#### Release Notes
We have simplified the packaging for custom assemblies by removing the `compiling ide` assembly module. This does impact any existing Codenvy custom assemblies as they will need changes to the 
`pom.xml` to build on 5.4.

#### Docs PR
TBD


benoitf:
@mshaposhnik I don't know policies for codenvy but IMHO Release Notes should be added as well as it's a task and not a bugfix

mshaposhnik:
@codenvy/pm pls approve

JamesDrummond:
Not all tasks need release notes. I generally feel if it effects users **directly** than release notes should be provided. I am not sure that this would need release notes unless it would break a users extension/plugin.

skabashnyuk:
Oh. yes it will brake it. instead of compiling-ide-war users should use assembly-ide-war.
FYI @eivantsov @bmicklea it have to be done in all  existed onprem custom assembly <5.4

bmicklea:
This needs release notes and docs

TylerJewell:
I think this is just an internal change and doesn't impact how end users work with building the system.

eivantsov:
For example, assembly that Software AG or EclipseSOurce has will be broken.

skabashnyuk:
About docs topic. Right now we are on the early stage of explaining how to build custom codenvy assembly. AFAIK only mentioned @eivantsov  companies have it. 
What to do.
in assembly-ide-war/pom.xml replace
```
        <dependency>
            <groupId>com.codenvy.onpremises</groupId>
            <artifactId>compiling-ide-war</artifactId>
            <classifier>classes</classifier>
            <scope>provided</scope>
        </dependency>
```
to with
```
        <dependency>
            <groupId>com.codenvy.onpremises</groupId>
            <artifactId>assembly-ide-war</artifactId>
            <classifier>classes</classifier>
            <scope>provided</scope>
        </dependency>
```
@codenvy/pm  please approve

JamesDrummond:
@bmicklea Should we create a extension migration page in documentation? Or should we provide a migration file in `eclipse/che` and link to it from docs? It could list 

5.3.x => 5.4.x :
1. Replace in assembly-ide-war/pom.xml
```
        <dependency>
            <groupId>com.codenvy.onpremises</groupId>
            <artifactId>compiling-ide-war</artifactId>
            <classifier>classes</classifier>
            <scope>provided</scope>
        </dependency>
```
with
```
        <dependency>
            <groupId>com.codenvy.onpremises</groupId>
            <artifactId>assembly-ide-war</artifactId>
            <classifier>classes</classifier>
            <scope>provided</scope>
        </dependency>
```

bmicklea:
@JamesDrummond I think adding that to the custom assemblies documentation is the right thing to do. We can be specific that it's 5.3 > 5.4. We'll leave it there for a few releases then remove it once we're far enough away from 5.3.

bmicklea:
@JamesDrummond can you create a docs issue and label it 5.3 and link it to here, then we can let the devs merge this PR.

mshaposhnik:
@bmicklea can i merge now ?

bmicklea:
@mshaposhnik yes

Issue #1749, opened 2017-02-09
Respect dns options of a Docker daemon for workspaces


eivantsov:
If an On Prem DNS can be resolved since only a custom DNS server resolves it, there's no way a Codenvy workspace can inherit dns from Docker daemon options.

Steps to reproduce:

1. Configure Docker daemon to be using a custom DNS resolver
2. Restart Codenvy
3. Start a workspace.

TylerJewell:
Is this a CLI improvement in that the CLI needs to detect the DNS and pass it to the codenvy server?  Or would the codenvy server handle this outside the CLI?

This is probably a Che issue too.

eivantsov:
When a workspace is created we probably send an empty array for dns, thus overriding Docker daemon settings.

TylerJewell:
Ok, if this is an issue for workspaces, then it's not a CLI issue.  It's an issue of how the Codenvy / Che server passes the DNS information into a workspace.

eivantsov:
Yes, not a CLI issue.

2 solutions that I see:

1. Do not send an empty array for dns thus overriding docker daemon settings.
2. Have a property in codenvy.env for a custom dns resolver address.

garagatyi:
It works in such a way because of custom dns resolution technique used in docker.
We need to fix Nginx and manually add resolver into container conf.
Workaround that probably would help right now is to use IP for codenvy master and docker nodes and set dns resolver into /etc/resolv.conf on each docker node host (Not in docker conf, on host where docker runs).
Fix from us would probably take couple of days.

PR #1771, opened 2017-02-13
CODENVY-1749: add possibility to set dns resolver

garagatyi:
### What does this PR do?
Add possibility to set DNS resolving servers.
If Codenvy is set to use custom DNS that is resolvable using custom DNS resolving server admin should set env var: **CODENVY_DNS_RESOLVERS**
It may contain several values in comma-separated format, e.g. 
```
CODENVY_DNS_RESOLVERS=10.10.10.10,8.8.8.8
```

Fixes codenvy/codenvy#1749

#### Changelog
Add possibility to set DNS resolving servers.

#### Release Notes
We now let you override the default DNS resolvers that are created in your workspace runtimes. In Codenvy, each workspace is given a set of runtimes. Agents and other services that run within those runtimes inherit networking properties from Che including proxy and security configuration. In some environments, custom resolution of DNS entries (usually to an internal DNS provider) is required to enable the workspace runtimes to have lookup ability from services within the workspace. You can now add CODENVY_DNS_RESOLVERS=<ip-list> to `che.env to customize the DNS injection into default containers. The default behavior is to inherit the list of DNS resolver servers from the host.

#### Docs PR
??


garagatyi:
@codenvy/pm Please review

dockerfiles/init/manifests/codenvy.pp
```diff
@@ -272,6 +272,10 @@
 # please leave this as it is if you don't need no_proxy configuration
   $no_proxy_for_codenvy_workspaces = getValue("CODENVY_NO_PROXY_FOR_CODENVY_WORKSPACES","")
 
+################################
+# DNS resolver configuration
+  $dns_resolvers = getValue("CODENVY_DNS_RESOLVERS","")
```
riuvshin: default value must be 'NULL' because you check on 'NULL' in template

dockerfiles/init/modules/nginx/templates/http_nginx.conf.erb
```diff
@@ -62,8 +62,7 @@ error_log                     /var/log/nginx/error.log error;
         #gzip_proxied expired no-cache no-store private auth;
         #gzip_types text/plain text/css text/xml text/javascript application/x-javascript application/xml;
 
-        resolver                    127.0.0.1 8.8.8.8 ipv6=off;
-
+        resolver                    <% if @dns_resolvers.empty? -%>8.8.8.8<% else -%><%= @dns_resolvers.gsub(',', ' ') %><% end -%> ipv6=off;
```
riuvshin: you shoul che on != 'NULL' same as in compose file `if @dns_resolvers != 'NULL'`

JamesDrummond:
Needs docs. Waiting on associated Che doc review before creating it for codenvy.

dockerfiles/init/manifests/codenvy.pp
garagatyi: 10x

dockerfiles/init/modules/nginx/templates/http_nginx.conf.erb
garagatyi: 10x

garagatyi:
@riuvshin I fixed PR.
@TylerJewell I commented docs in Che PR. Please consider changes in docs and port them into this PR also.

dockerfiles/init/modules/nginx/templates/http_nginx.conf.erb
```diff
@@ -62,8 +62,7 @@ error_log                     /var/log/nginx/error.log error;
         #gzip_proxied expired no-cache no-store private auth;
         #gzip_types text/plain text/css text/xml text/javascript application/x-javascript application/xml;
 
-        resolver                    127.0.0.1 8.8.8.8 ipv6=off;
-
+        resolver                    <% if @dns_resolvers != 'NULL' -%>8.8.8.8<% else -%><%= @dns_resolvers.gsub(',', ' ') %><% end -%> ipv6=off;
```
riuvshin: if statement look strange maybe I was confused you by asking check != null
but seems in that case it should looks like
`<% if @dns_resolvers == 'NULL' -%>8.8.8.8<% else -%><%= @dns_resolvers.gsub(',', ' ') %><% end -%>`
otherwise 8.8.8.8 will never be by defaul

garagatyi: Good catch, thx

garagatyi:
@riuvshin I had to rework PR, so now empty string is used instead of NULL. It is needed because Java code can't interpret NULL as empty array

riuvshin:
@garagatyi I saw this and approved

bmicklea:
@garagatyi was the merged to 5.3?  There's no milestone set.

PR #1772, opened 2017-02-14
do not redirect /agent-binaries to from http to https

riuvshin:
### What does this PR do?
allows download agents via http

related to https://github.com/eclipse/che/pull/4029

#### Changelog
allows download agents via http


benoitf:
@riuvshin could we merge this PR ?

PR #1781, opened 2017-02-15
add script to build agents init and codenvy images with custom name

riuvshin:
### What does this PR do?
allows build and push docker images with custom name to be able to update acceptances

#### Changelog
allows build and push docker images with custom name to be able to update acceptances

### DIFF
--- a/dockerfiles/custom_images.sh
+++ b/dockerfiles/custom_images.sh
@@ -1,0 +1,24 @@
+#!/bin/sh
+
+set -e
+
+HUB_USER=$1
+if [ -z $HUB_USER ]; then echo "please pass your docker hub username as an arg"; exit 2; fi
+
+is_logged=$(docker info | grep -c Username) || true
+if [ $is_logged -ne 1 ]; then
+    docker login
+fi
+
+sh agents/build.sh
+sh codenvy/build.sh
+sh init/build.sh
+
+# retag
+docker tag codenvy/init:nightly $HUB_USER/init:nightly
+docker tag codenvy/codenvy:nightly $HUB_USER/codenvy:nightly
+docker tag codenvy/agents:nightly $HUB_USER/agents:nightly
+# push
+docker push $HUB_USER/init:nightly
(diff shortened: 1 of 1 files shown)


dockerfiles/custom_images.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/sh
```
benoitf: missing copyright headers
Also I'm not sure that we require another top level script
IMHO it should be integrated in build.include

skabashnyuk:
how to use it?

TylerJewell:
The build.include is sourced in the build.sh scripts, so it's just a way to have it as a library if that makes sense.

dockerfiles/custom_images.sh
riuvshin: I don;t know how build.include works I need something right now because devs blocked to update acceptances. maybe in future we will merge them

riuvshin:
@skabashnyuk execute and follow.

dockerfiles/custom_images.sh
benoitf: I suppose it won't be merged in future if it's not properly done at the beginning.
if ppl need a fragment of script immediately they can use a gist to get that.
But when we submit something into code repository there is a review

riuvshin: in future that script will check if targets are there and other things which is needed only for devs so I don;t know if we should have everything in build.include
for me build.include is source for build.sh but I using build.sh to do some other things.

skabashnyuk:
@riuvshin not clear. Please provide more information how your code will be used

dockerfiles/custom_images.sh
riuvshin: it also may trigger jobs on CI to update acceptances, so...

riuvshin:
@skabashnyuk for now that script is used by devs to prepare images to update acceptances. I just wanted to make their life easier by adding this right in repo. full instruction HOW to update acceptance will be introduced later, but that script is part of that guide

PR #1842, opened 2017-02-28
Revert "INFRA-68; remove compiling-ide module (#1766)"

mkuznyetsov:
This reverts commit fd4fa31b5e7ed01cdbe83ef9f46819bf17790bec. related to https://github.com/codenvy/codenvy/pull/1766

### What does this PR do?
Going back to have a module for compiling GWT for IDE, and another module for it's assembly. It is needed, to be able to reuse GWT sources in other assemblies (like archetype samples), as well as keep assembly-ide-war clean

#### Changelog
This reverts PR https://github.com/codenvy/codenvy/pull/1766

#### Docs PR
https://github.com/codenvy/docs/pull/86


skabashnyuk:
@bmicklea @JamesDrummond  note after we merge this pr 
Docs and changes log from https://github.com/codenvy/codenvy/pull/1766 this pr have to be reverted too.

skabashnyuk:
Right now we trying to solve a couple of issues.
1. We want to not to include gwt dependencies that need for gwt compilation in resulting war.
2. We want to include runtime(not gwt related) jars in resulting war.
3. We want to be able to have a single artifact so plugin developers can include it to his gwt compiling war. 
4. In future we would like to decentralize gwt.xml between many maven artifacts. 

In this pr #1766 we solve **1-2** but forget about **3**.
I also have discussion about maven structure of gwt compiling with  @benoitf and @vparfonov 
During this discussion we have an idea to create uber jar with gwt dependencies but it conflicts **4**  Other idea was to keep gwt dependencies as provided but generate classes war dependency during build. But it require to write new maven plugin and we decided to postpone this task.

Issue #1729, opened 2017-02-08
Change resource distribution to setting resource CAPs


ashumilova:
The resource CAPs may have any value and it will limit the resource usage. But in case there is no CAP set - the team just used the available resources of the parent. The CAP doesn't guarantee there are as many resources available to reach that CAP.

slemeur:
Complete but not merged: PR to be created today.

PR #1845, opened 2017-02-28
Implement resources capping for suborganizations

sleshchenko:
### What does this PR do?
Rework organization resources in following way:
- remove moving resources from parent organization to its suborganization
- add ability suborganizations to use parent organization resources by default
- add ability to parent to set cap on shared resources usage

https://github.com/codenvy/codenvy/issues/1729

#### Changelog
- Rework organization distribution service to cap resource usage.

#### Release Notes
N/A part of teams

#### Docs PR
N/A part of teams


sleshchenko:
Fixing of reviewers' comments and adding migration scripts(renaming column name) will be done in another PR.

PR #1848, opened 2017-03-01
[UD] fixes for "Get More RAM" widget and Billing info tab

akurinnoy:
### What does this PR do?
This PR adds fixes for Billing info tab and "Get More RAM" widget:
- restrict some inputs to enter only valid symbols;
- set focus to input when widget is shown;
- apply filter to round fractional part of calculated values in widget;

N/A

#### Changelog

[UD] make inputs on "Get More RAM" widget and Billing info tab to allow to enter only valid symbols

#### Release Notes

N/A - bugfix

#### Docs PR

N/A - bugfix

### DIFF
--- a/dashboard/src/app/billing/card-info/card-info.html
+++ b/dashboard/src/app/billing/card-info/card-info.html
@@ -38,6 +38,7 @@
                che-label-name="City"
                che-name="city"
                che-place-holder=""
+               che-type-city
                aria-label="City"
                ng-model="cardInfoController.creditCard.city"
                ng-change="cardInfoController.infoChanged(cardInfoForm.$valid)"
@@ -51,6 +52,7 @@
                che-label-name="State"
                che-name="state"
                che-place-holder=""
+               che-type-city
                aria-label="State"
                ng-model="cardInfoController.creditCard.state"
                ng-change="cardInfoController.infoChanged(cardInfoForm.$valid)"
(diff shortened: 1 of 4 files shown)


dashboard/src/app/billing/card-info/card-info.html
```diff
@@ -51,6 +52,7 @@
                che-label-name="State"
                che-name="state"
                che-place-holder=""
+               che-type-city
```
benoitf: why do we have validator of cities for the state ?

akurinnoy: It's not validator of cities, it is just directive which prevent to enter not valid symbols to input.

benoitf: ok but my issue is that the name of validator contains "city" while it is more generic

akurinnoy: Could you suggest more suitable name?

benoitf: @akurinnoy well this change is not urgent I approved the PR. I don't have more suitable name. Maybe che-type-city and che-type-state directives that can be based on the same class.

akurinnoy: Thanks! I will update it soon.

Issue #1739, opened 2017-02-08
Do not make additional http request on each license check in SystemLicenseLoginFilter


skabashnyuk:
Making additional http request on each license check in SystemLicenseLoginFilter is very slow. Please use some smart logic to reduce time of license checking

skabashnyuk:
P1 blocker critical. Since it' double time of each request and made huge performance lost.

dmytro-ndp:
@skabashnyuk: You set **status/blocker** for this issue. Could you, please, explain which function of Codenvy it blocks (that is makes as inaccessible)?

skabashnyuk:
The function is our performance. Making for almost each http request additional http request that sometimes (not sure is it always or not ) make query to database   - is unacceptable solution.

dmytro-ndp:
Our selenium tests demonstrates almost the same performance for Che (without licensing) and Codenvy (with licensing):
- [4h 2 minutes for 326 tests for Che](https://ci.codenvycorp.com/view/qa/job/che-integration-tests/213/);
- [4h 10 minutes for 360 tests for Codenvy](https://ci.codenvycorp.com/view/qa/job/nightly-onprem-selenium-tests/265/).

IMHO, it disapproves assumption that current implementation of licensing leads to huge performance lost.

skabashnyuk:
@dmytro-ndp  you comparing hot and round. 
Can you give me result of simple http request. For instance ```-X OPTIONS /api``` or ```GET /api/user``` with and without license check?
How many times did we make DB query ```select count(id) from user``` ? when loading ide or dashboard. What for? Do we really need that?

dmytro-ndp:
> Can you give me result of simple http request. For instance -X OPTIONS /api or GET /api/user with and without license check?

I could check it. But, such kind of testing should precede conclusion about huge impact on performance.

> How many times did we make DB query select count(id) from user ? when loading ide or dashboard. What for? Do we really need that?

Requesting number of registered users is needed to protect our product from illegal usage: when this number reaches or exceeds allowed limit defined by license, and so stimulate users to pay for usage.

We do it on ~every requests so as new user can be registered at any time.

skabashnyuk:
Thx @dmytro-ndp

dmytro-ndp:
@skabashnyuk: you are welcome :-)

Also, we check license on ~every requests which require login to Codenvy so as we doesn't allow users to login when _Fair Source License_ doesn't accepted by admin.

[...]

PR #1863, opened 2017-03-02
CODENVY-1739: Do not make additional http request on each license check

mmorhun:
### What does this PR do?
Caches Fair Source License acceptance to prevent multiple requests with the same result.

#1739 

#### Changelog
Optimizes Fair Source license checks to improve performance.

#### Release Notes
We've optimized the performance of all HTTP requests by caching the necessary license checks.


skabashnyuk:
Will it cache requests if license is **NOT** accepted?

skabashnyuk:
Ok. And what is the problem? What if it will be 1 minute? And the opposite task. If license was ok but expired during tomcat life time should it be revalidated after X minutes?

tolusha:
> And what is the problem? What if it will be 1 minute?
I user could login after license had been accepted it would be awkward.
 
> If license was ok but expired during tomcat life time should it be revalidated after X minutes?
Agree.
@mmorhun  Take it into account.

mmorhun:
ci-build

Issue #1909, opened 2017-03-13
Unable to use all parent organization resources


sleshchenko:
Unable to use all parent organization resources.

Steps to reproduce:
1. Login as admin.
2. Set free resource limit to 4 GB.
3. Create team and two workspaces there which require 2GB to start.
4. Try to start both of workspaces.

Expected: Both workspaces start successfully.
Actual: The first workspace starts successfully. Start of the second workspace fails with error `Workspace admin/test/ws1 needs 2048MB to start. Your account has 0MB available and 2048MB in use. The workspace can't be start. Stop other workspaces or grant more resources.`

TylerJewell:
Yes, agree with the characterization as severity/blocker.

PR #1914, opened 2017-03-14
CODENVY-1909 Fix providing resources for suborganizations

sleshchenko:
### What does this PR do?
Fixes providing resources for suborganizations.
Resource reservation concept is not flexible enough for implementing different resource distribution for different account types. It was replaced with AvailableResourcesProvider it allows to implement different amount of available resources for different account types.

If account type doesn't have own resource distribution logic, then will be used default one which defines amount of available resources in following way: total resources(license resources) - resources which are used by account.

Personal account type has default available resources provider.

Organizational account type has own available resources provider that considers sharing resources for suborganizations and resources capping.
Provider calculates available resources for organizational account in following way: total resources(license resources) - resources which are used by organization - resources which are used by its suborganizations.
Available resources for suborganizations has a bit different formula: min of available resources for parent and for suborganization (available resources is calculated here by the formula that is listed above). This is required for correct working of resources capping.

https://github.com/codenvy/codenvy/issues/1909

#### Changelog
Fixed providing resources for suborganizations.

#### Release Notes
Fixeed issue when providing resources for suborganizations.

#### Docs PR
N/A


wsmaster/codenvy-hosted-api-organization/src/main/java/com/codenvy/organization/api/resource/OrganizationalAccountAvailableResourcesProvider.java
```diff
@@ -0,0 +1,184 @@
...
+                                                 .values());
+    }
+
+    /**
+     * Returns formatted string for list of resources.
+     */
+    private static String format(Collection<? extends Resource> resources) {
+        return '[' +
+               resources.stream()
+                        .map(resource -> resource.getAmount() + resource.getUnit() + " of " + resource.getType())
```
evoevodin: Does it makes sense to move this code to `resource.toString`  ?

wsmaster/codenvy-hosted-api-organization/src/main/java/com/codenvy/organization/api/resource/SuborganizationResourcesProvider.java
```diff
@@ -83,7 +83,7 @@ public SuborganizationResourcesProvider(AccountManager accountManager,
         }
 
         // given account is suborganization's account and can have resources provided by parent
-        List<? extends Resource> parentResources = usageManagerProvider.get().getAvailableResources(parent);
+        List<? extends Resource> parentResources = usageManagerProvider.get().getTotalResources(parent);
```
evoevodin: i would call the variable differently e.g. `parentTotal` or `parentAvaialble` then when you use it in the code it would be more _declarative_, so you could determine the problem earlier

wsmaster/codenvy-hosted-api-resource/src/main/java/com/codenvy/resource/api/ResourceAggregator.java
```diff
@@ -105,7 +104,61 @@ public ResourceAggregator(Set<ResourceType> resourcesTypes) {
...
+     * | ----------- | --------- | --------- | -------- |
+     * | Resource3   | 1         | -         | 1        |
+     * | ----------- | --------- | --------- | -------- |
+     * </pre>
+     *
+     * @param sourceResources
+     *         the source resources
+     * @param resourcesToCompare
+     *         the resources which should be compared to {@code sourceResources}
+     * @throws IllegalStateException
```
evoevodin: Why `IllegalStateException` but not `IllegalArgumentException` ?

wsmaster/codenvy-hosted-api-organization/src/main/java/com/codenvy/organization/api/resource/OrganizationalAccountAvailableResourcesProvider.java
sleshchenko: `resource.toString` returns `ResourceImpl{id=null, type='RAM', amount=1024, unit='mb'}` but in this case `1024 mb of RAM` is much more readable. So I would not use resource.toString here

evoevodin: The proposal wasn't to use current version of to string, but make it to return `1024 mb of RAM`

wsmaster/codenvy-hosted-api-organization/src/main/java/com/codenvy/organization/api/resource/SuborganizationResourcesProvider.java
sleshchenko: Makes sense

wsmaster/codenvy-hosted-api-organization/src/main/java/com/codenvy/organization/api/resource/OrganizationalAccountAvailableResourcesProvider.java
```diff
@@ -0,0 +1,184 @@
...
+     * <p>Note that if one of the lists don't contain resource
+     * than resource also won't contain it.
+     *
+     * @param resources1
+     *         the fist resources list to compare
+     * @param resources2
+     *         the second resources list to compare
+     * @return resources with minimum amounts.
+     */
+    private List<Resource> min(Collection<? extends Resource> resources1, Collection<? extends Resource> resources2) {
```
evoevodin: Does it make sense to separate this method on 2, 
1 is intersect, the second is aggregation by minimal amount? 
I spend some time getting what the method is doing.

wsmaster/codenvy-hosted-api-organization/src/main/java/com/codenvy/organization/api/resource/OrganizationalAccountAvailableResourcesProvider.java
sleshchenko: Yeah. Sorry, I missed your point. I would do `toString` methods for entities in the same way. For example, `toString()` methods for our entities can be useful for debugging and id is present there. If we change it then it may be a little problem.
WDYT? Will it be OK?

wsmaster/codenvy-hosted-api-resource/src/main/java/com/codenvy/resource/api/ResourceAggregator.java
```diff
@@ -70,33 +68,34 @@ public ResourceAggregator(Set<ResourceType> resourcesTypes) {
...
+    public List<? extends Resource> deduct(List<? extends Resource> sourceResources,
                                            List<? extends Resource> resourcesToDeduct) throws NoEnoughResourcesException {
-        checkSupporting(totalResources);
+        checkSupporting(sourceResources);
         checkSupporting(resourcesToDeduct);
 
-        final Map<String, Resource> result = totalResources.stream()
-                                                           .collect(Collectors.toMap(Resource::getType,
-                                                                                     Function.identity()));
+        final Map<String, Resource> result = sourceResources.stream()
```
evoevodin: usually you deal with resources as with a map<type, resouce>, and this code is repeated throughout all the codebase.
Let's consider collection -> map conversion as grouping by type, so `ResourceAggregator` can do that + it can describe and document override strategies for the resources with repeated type.

wsmaster/codenvy-hosted-api-resource/src/main/java/com/codenvy/resource/api/ResourceAggregator.java
sleshchenko: Good catch. I'll change throwing of IllegalStateException to IllegalArgumentException

wsmaster/codenvy-hosted-api-resource/src/main/java/com/codenvy/resource/api/ResourceAggregator.java
```diff
@@ -105,7 +104,61 @@ public ResourceAggregator(Set<ResourceType> resourcesTypes) {
...
+     * | Resource3   | 1         | -         | 1        |
+     * | ----------- | --------- | --------- | -------- |
+     * </pre>
+     *
+     * @param sourceResources
+     *         the source resources
+     * @param resourcesToCompare
+     *         the resources which should be compared to {@code sourceResources}
+     * @throws IllegalStateException
+     *         when {@code sourceResources} or {@code deduction} contain resource with not supported type
```
evoevodin: What is `deduction` ?

wsmaster/codenvy-hosted-api-organization/src/main/java/com/codenvy/organization/api/resource/OrganizationalAccountAvailableResourcesProvider.java
sleshchenko: Good proposal. In your case code looks much more clear

wsmaster/codenvy-hosted-api-organization/src/main/java/com/codenvy/organization/api/resource/OrganizationalAccountAvailableResourcesProvider.java
sleshchenko: There is necessity to have resources formatter that will return readable string for resource with converting to unit that is the most suitable, e.g. instead of 1024 mb of RAM -> 1Gb of RAM.
So it will be implemented later, and then we'll remove this private method

sleshchenko:
@codenvy/pm Please approve my changes

wsmaster/codenvy-hosted-api-resource/src/main/java/com/codenvy/resource/api/ResourceAggregator.java
sleshchenko: Sounds good. But I would not do such refactoring in this PR. It will be reworked later.

PR #1931, opened 2017-03-16
add possibility to configure docker namespace for snapshots

riuvshin:
### What does this PR do?
add possibility to configure docker namespace for pushing snapshots

#### Changelog
add possibility to configure docker namespace for pushing snapshots


garagatyi:
I think it is a bit confusing to call it docker namespace. Since it will go into changelog can you rephrase it.

richtera:
:+1:

Issue #1688, opened 2017-02-03
Strange behavior when user is not authorized


vitaliy-guliy:
Dashboard does not redirect to the login page when the user session is expired.

![user is not authorized](https://cloud.githubusercontent.com/assets/1655894/22590312/3ff9c710-ea17-11e6-8448-bc371b233972.gif)

sleshchenko:
Your session is valid. Looks like steps to reproducing are following:
1. Create team.
2. Add any other admin there.
3. Remove himself from team.
3. Try to open this team directly by URL {host}/dashboard/#team/{teamName}.

Then team exists but you don't have permissions to open it by Dashboard.
Dashboard should handle it correctly.

Another issue - looks like we should allow to see team, manage workspace and members to Team owner, even when he doesn't have permissions in this team

sleshchenko:
Also I've checked database on a1 instance and there is no any member in adminTeam1. So API doesn't allow to remove last member with `setPermissions` action. Looks like there is issue with concurrency.

sleshchenko:
When user tries to open team where he is not a member he'll following page
![image](https://cloud.githubusercontent.com/assets/5887312/24137487/a7200778-0e1c-11e7-8907-f82f6dee435c.png)

PR #1935, opened 2017-03-16
Add ability for users to remove own permissions

sleshchenko:
### What does this PR do?
Adds ability for users to remove their own permissions.

Also this PR contains changes for locking by instance on updating permissions for it. It is needed to avoid cases like https://github.com/codenvy/codenvy/issues/1688#issuecomment-277239223

https://github.com/codenvy/saas/issues/168

#### Changelog
Added ability for users to remove their own permissions.

#### Release Notes
Added ability for users to remove their own permissions.

#### Docs PR
N/A

### DIFF
--- a/wsmaster/codenvy-hosted-api-permission/src/main/java/com/codenvy/api/permission/server/PermissionsManager.java
+++ b/wsmaster/codenvy-hosted-api-permission/src/main/java/com/codenvy/api/permission/server/PermissionsManager.java
@@ -25,9 +25,12 @@ import com.google.common.collect.ImmutableMap;
 import org.eclipse.che.api.core.ConflictException;
 import org.eclipse.che.api.core.NotFoundException;
 import org.eclipse.che.api.core.Page;
+import org.eclipse.che.api.core.Pages;
 import org.eclipse.che.api.core.ServerException;
 import org.eclipse.che.api.core.notification.EventService;
 import org.eclipse.che.commons.env.EnvironmentContext;
+import org.eclipse.che.commons.lang.concurrent.StripedLocks;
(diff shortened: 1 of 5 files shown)


wsmaster/codenvy-hosted-api-permission/src/main/java/com/codenvy/api/permission/server/PermissionsManager.java
```diff
@@ -73,6 +77,7 @@ public PermissionsManager(EventService eventService,
         }
         this.domains = ImmutableList.copyOf(domains);
         this.domainToDao = ImmutableMap.copyOf(domainToDao);
+        this.updateLocks = new StripedLocks(16);
```
evoevodin: Does it really have to deal with distributed locks?
It looks like a single lock instance should be fine.

sleshchenko: Single lock instance will lock updating of all permissions.
StripedLocks distribute locking of permissions update by instance id.

Do you think it would be better to lock all permissions here?

evoevodin: up to you

sleshchenko:
@codenvy/pm Please approve

bmicklea:
Will the dashboard allow users to remove their own permissions to? Or is this just an API change?

sleshchenko:
This's just an API change. For allowing users to leave teams will be created another PR by @ashumilova

slemeur:
linking with https://github.com/codenvy/saas/issues/188

Issue #1940, opened 2017-03-17
Swarm container became inaccessible after rebooting computer with Codenvy


dmytro-ndp:
`docker ps -a` output fragment:
```
eclipse-che/workspacec4g5745jgi6l268i_machineiv8josjzhjpf9hx2_iedexmain_ws-machine   "/bin/sh -c 'sudo /us"   About an hour ago   Exited (1) 25 minutes ago       0.0.0.0:35052->22/tcp, 0.0.0.0:35051->4401/tcp, 0.0.0.0:35050->4403/tcp, 0.0.0.0:35049->4411/tcp, 0.0.0.0:35048->8000/tcp, 0.0.0.0:35047->8080/tcp, 0.0.0.0:35046->9876/tcp   workspacec4g5745jgi6l268i_machineiv8josjzhjpf9hx2_iedexmain_ws-machine
codenvy/codenvy:nightly  "/entrypoint.sh postf"   23 hours ago        Up 24 minutes                   0.0.0.0:32001->32001/tcp, 0.0.0.0:32101->32101/tcp    codenvy_codenvy_1
swarm:1.2.6              "/swarm manage -H 0.0"   23 hours ago        Exited (1) 25 minutes ago codenvy_swarm_1
haproxy:1.5.18-alpine    "/docker-entrypoint.s"   23 hours ago        Up 24 minutes                   0.0.0.0:80->80/tcp, 0.0.0.0:443->443/tcp              codenvy_haproxy_1
zookeeper:3.4.9          "/docker-entrypoint.s"   23 hours ago        Exited (1) 25 minutes ago codenvy_zookeeper_1
registry:2.5.0           "/entrypoint.sh /etc/"   23 hours ago        Exited (1) 25 minutes ago codenvy_registry_1                                        
postgres:9.5.4           "/docker-entrypoint.s"   23 hours ago        Up 25 minutes                   5432/tcp                                              codenvy_postgres_1                                        
codenvy/rsyslog:nightly  "rsyslogd -n"            23 hours ago        Restarting (1) 10 minutes ago   0.0.0.0:5140->514/tcp                                 codenvy_rsyslog_1                                         
codenvy/agents:nightly   "lighttpd -f /etc/lig"   23 hours ago        Up 25 minutes                   9000/tcp                                              codenvy_agents_1                                          
[...]

PR #1964, opened 2017-03-21
add rsyslog entrypoint

riuvshin:
### What does this PR do?
add entrypoint to rsyslog image where we kill pid file if it exist before start rsyslog, that may happen on force restart container and rsyslog wount be able to start.

fixes https://github.com/codenvy/codenvy/issues/1940

#### Changelog
add rsyslog entrypoint

### DIFF
--- a/dockerfiles/rsyslog/entrypoint.sh
+++ b/dockerfiles/rsyslog/entrypoint.sh
@@ -1,0 +1,14 @@
+#!/bin/sh
+# Copyright (c) 2016 Codenvy, S.A.
+# All rights reserved. This program and the accompanying materials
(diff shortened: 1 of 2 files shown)


dockerfiles/rsyslog/entrypoint.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
+# Copyright (c) 2016 Codenvy, S.A.
```
benoitf: if you can bump the date of the copyright to 2017

riuvshin: sure

PR #1975, opened 2017-03-23
Codenvy-1940: some containers became inaccessible after rebooting

riuvshin:
### What does this PR do?
in order to fix: https://github.com/codenvy/codenvy/issues/1940 afret some investigation i've figured out that our issues were due to using docker logging with syslog which we running as a container. That was a mistake, we should not do that because it works only with compose where we define start order, but when instance or docker daemon reboot containers start in random order which makes failure when image which wants to send logs to syslog started before syslog. And this kind of failures is kind of fatal for docker and docker daemon even don't trying to restart failed container although we've specified `restart: always`.

So here i;ve removed using docker logging with syslog which should fix the issue, but we still need to get logs from containers so i did some changes to make this works same as before.
we had 3 affected images: zookeeper, registry, swarm
For zookeeper i figured how to configure it to write log in to file.
For registry it was enough just to add custom entrypoint where I did log forwarding to file
Unfortunately swarm packaged as just a binary in `scratch` image so there is no bash and no any way to get log, so I added codenvy/swarm image where i get swarm and put in in alpine image which allows me to finally get logs in to a file.

https://github.com/codenvy/codenvy/issues/1940

#### Changelog
fix bug where some containers became inaccessible after rebooting

### DIFF
--- a/dockerfiles/cli/version/5.6.0/images
+++ b/dockerfiles/cli/version/5.6.0/images
@@ -1,10 +1,10 @@
 IMAGE_HAPROXY=haproxy:1.5.18-alpine
 IMAGE_NGINX=nginx:1.10-alpine
-IMAGE_SWARM=swarm:1.2.6
 IMAGE_REGISTRY=registry:2.5.0
 IMAGE_POSTGRES=postgres:9.5.4
 IMAGE_COMPOSE=docker/compose:1.8.1
 IMAGE_ZOOKEEPER=zookeeper:3.4.9
+IMAGE_SWARM=codenvy/swarm:5.6.0
 IMAGE_SOCAT=codenvy/socat:5.6.0
 IMAGE_INIT=codenvy/init:5.6.0
 IMAGE_AGENTS=codenvy/agents:5.6.0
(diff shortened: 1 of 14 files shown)


garagatyi:
AFAIR we have dependencies between codenvy containers. Am I right? If so then does docker daemon respect it after restart?

riuvshin:
@garagatyi unfortunately - no, as docker guys told me "all depends and ordering things implemented in compose, not docker."

garagatyi:
Then I believe we should not support reloading codenvy on daemon reloading. With your fix it won't be broken because of logging, but because of other things. And it will be broken only sometimes which may make support team life much harder.

riuvshin:
@garagatyi other things covered with entrypoints, for now the only issue with reboot is this one.
and just imagine your prod server was rebooted during the night and system was not self recovered :)

garagatyi:
What do you think about replacing rsyslog container with another container that reads logs of containers using `docker logs -f <container>` and puts logs into our volume mounted directory?

riuvshin:
we've added rsyslog just to get haproxy logs, if i will find a way how to get haproxy logs without rsyslog i will remove rsyslog container. (i'm investigating this now)

benoitf:
BTW it would be nice to get rid of rsyslog container

riuvshin:
@benoitf yeah.. but from what i found haproxy log only to syslog :(
maybe we will replace someday haproxy with nginx but this could be hard...

Issue #1851, opened 2017-03-01
Add mecanism to allow invitating users who are not registered


ashumilova:
# Goal
Since we have introduced teams we have improved the way users can share their work with other users. 

Today, the problem we face is that when a user is inviting other users into a team (or a workspace), they need to already have registered onto codenvy.io.
We want to improve this so that it allow to invite people who are not yet registered onto codenvy.io.

The benefits are that it helps to get viral and get more account created. Those accounts created are usually interesting because they are introduced to codenvy by other people that they know and that they trust. 

## Flow
1. From dashboard, user can provide emails of un-registered users when adding members to a team
2. An email is sent to the un-registered users with a link to the create account page
3. User fill the form and email will automatically be pre-filled with the same email adress than in `1.` 
4. User is created onto the system
5. User is added into the team he was initially invited

## Technical tradeoff
- We might limit the number of invitation a user can send per a certain period of time
- We might use reCapcha
- We can envision using hash to identify email that a team admin wanted to invite and compare it with the one given in account creation form

## Long term ideas
- We can imagine a mecanism to check if the invitation sent to an email have been accepted (=account created) and resend an email for those who have not yet registered.

TylerJewell:
Some important aspects here:
1. If a user initiates the account creation process from an invitation email, it should not be necessary for the user to have to re-verify their email address.  It is known from the initiation link in the email sent to their address.

2. Once the user creates their account in Codenvy, then they should not have to take additional steps to join the team.

3. If an invitation over email address is sent to an existing user associated with that same email address, we should recognize that in the invitation window and add them like a permanent user.

slemeur:
Cross linking with initial issue: https://github.com/codenvy/saas/issues/161

slemeur:
Updating description of the issue.

sleshchenko:
@slemeur Do we have template for inviting members via email?

PR #1978, opened 2017-03-23
Add ability to invite unregistered users to organization and workspace

sleshchenko:
### What does this PR do?
Implements Invite API that allows to invite unregistered users to organizations and workspaces.

Workflow:
1. Admin of organization or workspace can invite unregistered user by email. Then invitations can be updated (list of permissions) or removed at all.
2. Invited user will receive email invite with link for automatically registration (bearer token responsible for it, so it invalidates after configured period of time or after tomcat restarting).
3. Invited user can register himself by email or oauth without using provided by email invite token. Then all invitations will be accepted automatically.
4. If user is invited in more than one organization/workspace he will become a member/worker of all organizations/workspaces where he was invited.

Rest API description:
```
@Path("/invite")
InviteService {
    @POST
    void invite(InviteDto inviteDto);

    @GET
    @Path("/{domain}")
    Page<InviteDto> getInvites(@PathParam String domain, @QueryParam String instance,
                               @QueryParam long skipCount, @QueryParam int maxItems);

    @DELETE
    @Path("/{domain}")
    void remove(@PathParam String domain, @QueryParam String instance, @QueryParam String email);
}
```
```
Invite {
  String getEmail();
  String getDomainId();
  String getInstanceId();
  List<String> getActions();
}
```
The following values are applicable for domain id: organization and workspace.

Invites will be stored in one table. Integrity will be provided by two nullable fields which will reference organization or workspace, which field will be used depends on value in domain_id field. See [create_invites.sql](https://github.com/codenvy/codenvy/pull/1978/files#diff-23c43f7076435b88e7c94a63ed75172eR16)

Invite API is not module expansible. So when we'll need to have invites for another domains we'll must to edit invites module instead of adding new one.

It is not implemented in module expansible way like we did permissions because the best way to do it is rework existing permisisons with ability to link to email or user and ability to accept or decline invite. It requires more investigation and work and can be implemented later.

#### Emails
image
image

https://github.com/codenvy/codenvy/issues/1851

#### Changelog
Added ability to invite unredistered users to organization and workspace.

#### Release Notes
Added ability to invite unredistered users to organization and workspace

#### Docs PR
N/A


wsmaster/codenvy-hosted-api-invite/src/main/java/com/codenvy/api/invite/InviteImpl.java
```diff
@@ -0,0 +1,194 @@
...
+        this.domainId = domainId;
+        switch (domainId) {
+            case OrganizationDomain.DOMAIN_ID:
+                this.organizationId = instance;
+                break;
+            case WorkspaceDomain.DOMAIN_ID:
+                this.workspaceId = instance;
+                break;
+            default:
+                //do nothing
```
evoevodin: throw exception ?

PR #1990, opened 2017-03-27
allow to enable/disable exposing swarm and socat ports

riuvshin:
### What does this PR do?
allow to enable/disable exposing swarm and socat ports

#### Changelog
allow to enable/disable exposing swarm and socat ports

### DIFF
--- a/dockerfiles/init/manifests/codenvy.env
+++ b/dockerfiles/init/manifests/codenvy.env
@@ -144,6 +144,11 @@
 #     configure this as we manage it with the CLI.
 #CODENVY_SWARM_NODES=172.17.0.1:2375
 
+# Expose swarm port
+#     Expose swarm port tobe available on host as 23751
+#     Enabled by default, to disable please uncomment and set CODENVY_EXPOSE_SWARM_PORT=false
+#CODENVY_EXPOSE_SWARM_PORT=true
+
 # Docker Host
 #     How Codenvy will connect to the Docker host if CODENVY_DOCKER_DAEMON__URL is not 
 #     set. This value can be provided on the command line as well. If this value is
@@ -622,3 +627,16 @@
 #CODENVY_JMX_USERNAME=admin
 #CODENVY_JMX_PASSWORD=codenvy
 
+
+########################################################################################
+#####                                                                              #####
+#####                                     SOCAT                                    #####
+#####                                                                              #####
+# Expose socat port
+#     Expose socat port tobe available on host as 23750. It is needed only in all-in-one
+#     (basic mode), this will allow start a workspace without adding remote nodes for
+#     running worksoaces (production) mode.
+#     So if you are using remote nodes for running workspaces (production mode) this expose
+#     could be disabled.
+#     Enabled by default, to disable please uncomment and set CODENVY_EXPOSE_SWARM_PORT=false
+#CODENVY_EXPOSE_SWARM_PORT=true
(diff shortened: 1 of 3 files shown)


dockerfiles/init/manifests/codenvy.env
```diff
@@ -144,6 +144,11 @@
 #     configure this as we manage it with the CLI.
 #CODENVY_SWARM_NODES=172.17.0.1:2375
 
+# Expose swarm port
+#     Expose swarm port tobe available on host as 23751
```
benoitf: typo after to before be

dockerfiles/init/manifests/codenvy.env
```diff
@@ -622,3 +627,16 @@
...
+#####                                     SOCAT                                    #####
+#####                                                                              #####
+# Expose socat port
+#     Expose socat port tobe available on host as 23750. It is needed only in all-in-one
+#     (basic mode), this will allow start a workspace without adding remote nodes for
+#     running worksoaces (production) mode.
+#     So if you are using remote nodes for running workspaces (production mode) this expose
+#     could be disabled.
+#     Enabled by default, to disable please uncomment and set CODENVY_EXPOSE_SWARM_PORT=false
+#CODENVY_EXPOSE_SWARM_PORT=true
```
benoitf: looks like an invalid name for the property SWARM/SOCAT

dockerfiles/init/manifests/codenvy.env
```diff
@@ -622,3 +627,16 @@
 #CODENVY_JMX_USERNAME=admin
 #CODENVY_JMX_PASSWORD=codenvy
 
+
+########################################################################################
+#####                                                                              #####
+#####                                     SOCAT                                    #####
+#####                                                                              #####
+# Expose socat port
+#     Expose socat port tobe available on host as 23750. It is needed only in all-in-one
```
benoitf: same typo at `tobe`

dockerfiles/init/manifests/codenvy.env
riuvshin: not sure I understand this comment, can you please elaborate?

benoitf: name is CODENVY_EXPOSE_SWARM_PORT=true
while it should be _SOCAT_ (no SWARM here)

riuvshin: ahhh, silly copy paste!

dockerfiles/init/manifests/codenvy.pp
```diff
@@ -20,6 +20,7 @@
 # swarm nodes management
 # coma separated list of <IP>:<PORT>
   $swarm_nodes = getValue("CODENVY_SWARM_NODES","172.17.0.1:2375")
+  $expose_swarm_port = getValue("CODENVY_EXPOSE_SWARM_PORT","true")
```
garagatyi: It should be false by default to increase security level

dockerfiles/init/manifests/codenvy.pp
```diff
@@ -333,6 +334,10 @@
   $rsync_ssh_log_level=getValue("RSYNC_SSH_LOG_LEVEL","INFO")
 
 ###############################
+# socat
+#
+  $expose_socat_port = getValue("CODENVY_EXPOSE_SOCAT_PORT","true")
```
garagatyi: It should be false by default to increase security level

riuvshin: if this will be false all-in-one mode will not work.

TylerJewell:
@roman - configuring these properties probably needs to be added to the configuration docs, or probably more importantly the run book, as this is a security measure to take once you roll it out for adoption.

FYI - we are not using the terms "basic", "production" or "all-in-one" in our docs in how we describe the system to end users.  The system is always a production system, and the architecture for a single node vs. multiple nodes is also the same.  the only things that are different are some configuration properties - so we are avoiding language of talking about the system in some sort of mode, as it's really only got one mode with many configuration properties.

Issue #1995, opened 2017-03-28
codenvy cli stop works very slow for some containers


riuvshin:
**Reproduction Steps:**
just try to stop codenvy with no running workspaces
**Codenvy version:**    5.6.0-SNAPSHOT

PR #1997, opened 2017-03-28
Codenvy-1995: fix codenvy cli stop very slow

riuvshin:
### What does this PR do?
Fix very slow codenvy cli stop operation.

Thanks @akorneta for reporting this issue. After some investigation I figured out that this was introduced by: https://github.com/codenvy/codenvy/commit/79079d048c22d5c3cdd1e6c75ab7d41f1dc201de

That happened due to wrong entrypoint which spawned more than 1 process in a container which is antiPattern for docker flow. During docker stop docker send special signal to notify main process (usually pid = 1) in container that this container is going to be stopped, but in our case process with pid = 1 was entrypoint script instead of swarm. As result swarm didn't received stop signal and keep working so docker killed that container with force after timeout **~5min**.

To fix this issue I added logic to entrypoint scripts to handle termiation and properly stop needed process.

Fixes: https://github.com/codenvy/codenvy/issues/1995

#### Changelog
Fix very slow codenvy cli stop operation.

### DIFF
--- a/dockerfiles/init/modules/compose/templates/docker-compose.yml.erb
+++ b/dockerfiles/init/modules/compose/templates/docker-compose.yml.erb
@@ -109,9 +109,7 @@ services:
     image: <%= ENV["IMAGE_SWARM"] %>
     volumes:
       - '<%= scope.lookupvar('compose::codenvy_folder') -%>/config/swarm/node_list:/node_list'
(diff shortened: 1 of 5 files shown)


riuvshin:
@skryzhny yeah we already discussed this with @garagatyi, I'll try to improve this.

riuvshin:
OK so I've updated both entrypoints for swarm and registry, now `docker logs -f` works also.
@garagatyi @benoitf @skryzhny please review this again

Issue #1734, opened 2017-02-08
Unexpected errors after consuming a factory and creation of workspace


musienko-maxim:
**Reproduction Steps:** Sometimes after creation workspaces from stacks or consuming a factory we have exceptions like: 

```
2017-02-08 10:57:32,170[aceSharedPool-2]  [WARN ] [c.c.machine.RemoteDockerNode 101]    - The container '9026c92563bf0d6a447fefc4d4a08417bdd2f6befc163f1ac2eb7a5e2107c370' in the workspace 'workspacegziu8a0lih0agv
kw' won't be backed up. The reason is that the workspace wasn't restored from backup
2017-02-08 10:57:33,419[aceSharedPool-2]  [ERROR] [o.e.c.a.w.s.WorkspaceManager 779]    - Error occurs while executing command [/bin/sh, -c, trap '[ -z "$(jobs -p)" ] || kill $(jobs -p); [ -e /tmp/docker-exec-13.
pid ] && rm /tmp/docker-exec-13.pid' EXIT; echo $$>/tmp/docker-exec-13.pid; command -v pidof >/dev/null 2>&1 && {
    pidof sshd >/dev/null 2>&1 && echo 0 || echo 1
} || {
    ps -fC sshd >/dev/null 2>&1 && echo 0 || echo 1
}]: Error response from docker API, status: 404, message: Exec b98fb0f4a66de8046125e262ca520f7f6af68783440ac0032cf4baf095626779 not found

org.eclipse.che.api.core.ServerException: Error occurs while executing command [/bin/sh, -c, trap '[ -z "$(jobs -p)" ] || kill $(jobs -p); [ -e /tmp/docker-exec-13.pid ] && rm /tmp/docker-exec-13.pid' EXIT; echo 
$$>/tmp/docker-exec-13.pid; command -v pidof >/dev/null 2>&1 && {
    pidof sshd >/dev/null 2>&1 && echo 0 || echo 1
} || {
    ps -fC sshd >/dev/null 2>&1 && echo 0 || echo 1
}]: Error response from docker API, status: 404, message: Exec b98fb0f4a66de8046125e262ca520f7f6af68783440ac0032cf4baf095626779 not found
        at org.eclipse.che.api.agent.server.launcher.AbstractAgentLauncher.launch(AbstractAgentLauncher.java:100) ~[che-core-api-agent-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
[...]

PR #2011, opened 2017-03-29
CODENVY-1734: workaround docker exec not found problem in swarm

garagatyi:
### What does this PR do?
Workaround docker exec not found problem in swarm by waiting extra second and sending request again.
Also Swarm connector tries to log docker exec inspect info. 

Related to #1734 

#### Changelog
Add workaround for bug in Swarm when exec can't be found by Swarm

### DIFF
--- a/plugins/plugin-hosted/codenvy-machine-hosted/src/main/java/com/codenvy/swarm/client/SwarmDockerConnector.java
+++ b/plugins/plugin-hosted/codenvy-machine-hosted/src/main/java/com/codenvy/swarm/client/SwarmDockerConnector.java
@@ -17,18 +17,25 @@ package com.codenvy.swarm.client;
 import com.codenvy.swarm.client.model.DockerNode;
 import com.google.common.base.Strings;
 
+import org.eclipse.che.commons.annotation.Nullable;
 import org.eclipse.che.plugin.docker.client.DockerApiVersionPathPrefixProvider;
 import org.eclipse.che.plugin.docker.client.DockerConnector;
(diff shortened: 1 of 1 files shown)


plugins/plugin-hosted/codenvy-machine-hosted/src/main/java/com/codenvy/swarm/client/SwarmDockerConnector.java
```diff
@@ -112,6 +121,33 @@ public ContainerCreated createContainer(CreateContainerParams params) throws IOE
...
+            throws IOException {
+        try {
+            super.startExec(params, execOutputProcessor);
+        } catch (ExecNotFoundException e) {
+            // Sometimes swarm returns this error for unknown reason, see https://github.com/docker/swarm/issues/2664
+            // Log additional info to find out if this endpoint knows about exec at exactly that time
+            logMissingExecInfo(params.getExecId());
+            try {
+                // Wait in case swarm needs some time to find exec
+                Thread.sleep(1000);
```
mmorhun: I'm afraid that 1 second could be not enough. Please set a bit longer delay (maybe 5 seconds is long delay, but it will be better to wait it (almost for sure that exec will be visible by swarm) than give workspace starting error to user)

garagatyi: I think it is not so critical because we still trying to figure out proper fix on Swarm side. But I set it to 3 seconds. Hopefully it will be enough.

PR #2115, opened 2017-04-20
Removes all master authentication cookies from ws-agent request on haproxy

riuvshin:
### What does this PR do?
Removes all master authentication cookies from ws-agent request on haproxy

https://github.com/codenvy/infrastructure/issues/84

#### Changelog
Remove all master authentication cookies from ws-agent request on haproxy

### DIFF
--- a/dockerfiles/init/modules/haproxy/templates/haproxy.cfg.erb
+++ b/dockerfiles/init/modules/haproxy/templates/haproxy.cfg.erb
@@ -35,9 +35,9 @@ backend tomcat_bk
 
 backend nginx_bk
     mode http
-    http-request del-header token-access-key
-    http-request del-header session-access-key
-    http-request del-header logged_in
+    http-request replace-header Cookie (.*)session-access-key=([^;]*);(.*) \1\3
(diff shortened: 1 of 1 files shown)


dockerfiles/init/modules/haproxy/templates/haproxy.cfg.erb
```diff
@@ -35,6 +35,9 @@ backend tomcat_bk
 
 backend nginx_bk
     mode http
+    http-request replace-header Cookie (.*)session-access-key=([^;]*);(.*) \1\3
```
skryzhny: Probably we need to add case where our cookie is last and don't have trailing ';'

garagatyi: +1

skryzhny: Like `other-cookie=rdwdr;session-access-key=mhgh`

garagatyi: So correct regexp is: `http-request replace-header Cookie (.*)session-access-key=([^;]*);?(.*) \1\3`

Issue #994, opened 2016-10-20
There is no user name in the bottom left corner in dashboard.


vkuznyetsov:
There is no user name in the bottom left corner in dashboard for user created with an email and a username.

Step to reproduce:
1. Create a new account with email and username
2. Click on the link in confirmation email
3. Goto Dashboard

Expected result:
- There is Username in the left bottom corner if a user filled out his profile or any other user friendly text if user name empty in a user profile.

Actual result:
- Only Gravatar Icon in the corner.
  ![codenvy account](https://cloud.githubusercontent.com/assets/2101815/19566277/6030abe8-96f2-11e6-8477-c17c4074713c.png)

slemeur:
+1 with this bug The information given to create the account without oauth, must be taken in the user profile

PR #2163, opened 2017-05-05
CODENVY-994 Show the user nickname in the left bottom corner of UD

olexii4:

### What does this PR do?
Show the user email as user nickname in the left bottom corner of UD NavBar in the case when user first name and user last name are empty.

https://github.com/codenvy/codenvy/issues/994

#### Changelog
-- Show the user email in the case when user first name and user last name are empty.

#### Release Notes
N/A

selection_021
selection_022

### DIFF
--- a/dashboard/src/app/navbar/navbar.controller.ts
+++ b/dashboard/src/app/navbar/navbar.controller.ts
@@ -160,6 +160,17 @@ export class CodenvyNavBarController {
   }
 
   /**
+   * Returns user nickname.
+   * @return {string}
+   */
+  getUserName(): string {
+    const {attributes, email} = this.profile;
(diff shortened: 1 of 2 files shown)


dashboard/src/app/navbar/navbar.controller.ts
```diff
@@ -160,6 +160,17 @@ export class CodenvyNavBarController {
   }
 
   /**
+   * Returns user nickname.
+   * @return {string}
+   */
+  getUserNickname(): string {
```
ashumilova: can we rename it to getUserName()?

ashumilova: or getUserDisplayName

olexii4: OK
done

Issue #1812, opened 2017-02-21
Limiting of resources works wrong when configured value equals to -1


sleshchenko:
When there are configured [default accounts' resources](https://github.com/codenvy/codenvy/blob/master/dockerfiles/init/manifests/codenvy.env#L171-L205) with -1 value, then limiting of resources works incorrect.

Expected:
User will be able to use unlimited resources.

Actual: 
User gets error message when he tries to use resource
`You are only allowed to create -1 workspaces.`
`You are only allowed to start -1 workspaces.`
`Workspace needs 1000mb to start. Your account has -1 mb and 0mb in use. The workspace can't be start. Stop other workspaces or grant more resources.`

bmicklea:
@sleshchenko is this something you're working on in this sprint?

sleshchenko:
@bmicklea I've prepared changes that are needed on server side https://github.com/codenvy/codenvy/pull/2174 and @ashumilova will adapt dashboard. So then we'll merge our changes together.

ashumilova:
Took to the next sprint - will start work on it tomorrow.

PR #2174, opened 2017-05-10
CODENVY-1812 Add ability to use -1 in the amount of resources

sleshchenko:
### What does this PR do?
Adds ability to use -1 in the amount of resources while setting resources limits via external configurations or via free resources service.

https://github.com/codenvy/codenvy/issues/1812

#### Changelog
Added ability to use -1 in the amount of resources.

#### Release Notes
N/A

#### Docs PR
N/A

### DIFF
--- a/dashboard/src/app/billing/ram-info/ram-info.controller.ts
+++ b/dashboard/src/app/billing/ram-info/ram-info.controller.ts
@@ -105,7 +105,7 @@ export class RamInfoController {
     let ram = this.lodash.find(resources, (resource: any) => {
       return resource.type === CodenvyResourceLimits.RAM;
     });
-    return ram ? (ram.amount / 1024) : 0;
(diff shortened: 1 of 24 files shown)


dashboard/src/app/billing/ram-info/ram-info.controller.ts
```diff
@@ -105,7 +105,7 @@ export class RamInfoController {
     let ram = this.lodash.find(resources, (resource: any) => {
       return resource.type === CodenvyResourceLimits.RAM;
     });
-    return ram ? (ram.amount / 1024) : 0;
+    return (ram && ram.amount > -1) ? (ram.amount / 1024) : -1;
```
benoitf: should it be instead of `> -1` or `> 0` (or `!= -1)`

ashumilova: fixed

Issue #2164, opened 2017-05-05
Provide a way of searching users by part of their e-mail or name for system admins only


akorneta:
We have several use cases where system administrators needs to filter/search users by part of their e-mail addresses, e.g.

1. When the administrator adds user to root organization (_onprem_)
2. When the administrator filters the list of all users for the keyword phrase their email address (_onprem/saas_)

PR #2191, opened 2017-05-15
CODENVY-2164: Add search user method by email/username for admins

akorneta:
### What does this PR do?
Provides a method for administrators that allow to search users by email/name parts. New method is case insensitive and finds users with names/emails that contains given part. 
**method path**: `api.endpoint/admin/user/find`

#2164 

#### Changelog
Add search user method by email/username for admins

#### Release Notes
Add new API to allow system admin to search users by names and emails

### DIFF
--- a/wsmaster/codenvy-hosted-api-admin/src/main/java/com/codenvy/api/user/server/AdminUserService.java
+++ b/wsmaster/codenvy-hosted-api-admin/src/main/java/com/codenvy/api/user/server/AdminUserService.java
@@ -23,12 +23,12 @@ import io.swagger.annotations.ApiResponses;
 import org.eclipse.che.api.core.BadRequestException;
 import org.eclipse.che.api.core.Page;
 import org.eclipse.che.api.core.ServerException;
+import org.eclipse.che.api.core.model.user.User;
 import org.eclipse.che.api.core.rest.Service;
 import org.eclipse.che.api.core.rest.annotations.GenerateLink;
 import org.eclipse.che.api.user.server.DtoConverter;
 import org.eclipse.che.api.user.server.UserLinksInjector;
 import org.eclipse.che.api.user.server.UserManager;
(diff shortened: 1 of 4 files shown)


wsmaster/codenvy-hosted-api-admin/src/main/java/com/codenvy/api/user/server/AdminUserService.java
```diff
@@ -91,4 +92,40 @@ public Response getAll(@ApiParam(value = "Max items") @QueryParam("maxItems") @D
...
     }
+
+    @GET
+    @Path("/find")
+    @Produces(APPLICATION_JSON)
+    @ApiOperation(value = "Gets all users whose e-mail/name contains(case insensitive) given part")
+    @ApiResponses({@ApiResponse(code = 200, message = "The response contains searching result"),
+                   @ApiResponse(code = 400, message = "Missed required parameters, parameters are not valid"),
+                   @ApiResponse(code = 500, message = "Internal server error occurred")})
+    public Response find(@ApiParam("User e-mail fragment")
```
tolusha: Usually searching criteria tends to extend. It is better to pass `DTO` to avoid method signature refactoring in the future.

tolusha:
Why `n/a` for changes log?

akorneta:
added changes log

wsmaster/codenvy-hosted-api-admin/src/main/java/com/codenvy/api/user/server/AdminUserService.java
akorneta: I guess that in this case it is not critical, because user do not provide to many fields to search by, but for me it's ok to user DTO class instead of query parameters.

slemeur:
Is this API documented in swagger?

akorneta:
@slemeur yes

PR #2213, opened 2017-05-18
Add CSRF prevention filter and adapt clients

evoevodin:
### What does this PR do?
Adds CSRF protection. Resolves codenvy/infrastructure#86.

#### Serverside 
Each request filtered by `LoginFilter` now also filtered by `CodenvyCsrfFilter`.
`CodenvyCsrfFilter` delegates csrf validation routine to the standard [filter](https://tomcat.apache.org/tomcat-7.0-doc/config/filter.html#CSRF_Prevention_Filter_for_REST_APIs)
when _COOKIE_ based authentication is used, which means that clients like CLI(those that use token) will continue to work without any refactoring.
Clients scenario:

- Get token
```
Request headers:
...
session-access-key=0XABCDE...
X-CSRF-Token=Fetch
...

Response headers:
...
X-CSRF-Token=0ABCD(token value)
...
```
- Use token for modification requests:
```
Request headers:
...
session-access-key=0XFFFABCDE...
X-CSRF-Token=0ABCD(token value)
...
```

- Use invalid token or no token in modification request
```
Request headers:
...
session-access-key=0XABCDE...
X-CSRF-Token=hello
...

response is negative code = 403
Response headers:
...
X-CSRF-Token=Required
...
```

#### Changelog
Enabled [CSRF](https://en.wikipedia.org/wiki/Cross-site_request_forgery) [protection](https://tomcat.apache.org/tomcat-7.0-doc/config/filter.html#CSRF_Prevention_Filter_for_REST_APIs) for clients that use cookies

#### Release Notes
Enabled CSRF protection which affects all clients that use cookies for authentication, 
now those clients must pass a valid `X-CSRF-Token` header within each modifying request(POST, PUT, DELETE).

Merge depends on https://github.com/codenvy/qa/issues/872

### DIFF
--- a/assembly/assembly-ide-war/src/main/webapp/_app/activity.js
+++ b/assembly/assembly-ide-war/src/main/webapp/_app/activity.js
@@ -75,7 +75,7 @@ var ActivityTracker = new function () {
         var xhr = new XMLHttpRequest();
 
         xhr.onreadystatechange = function() {
-            if (xhr.readyState == 4) {
+            if (xhr.readyState == XMLHttpRequest.DONE) {
                 csrfToken = xhr.getResponseHeader("X-CSRF-Token");
                 if (csrfToken) {
                     tokenPresentHandler(csrfToken);
--- a/assembly/assembly-ide-war/src/main/webapp/_app/activity.js
+++ b/assembly/assembly-ide-war/src/main/webapp/_app/activity.js
@@ -19,12 +19,19 @@ var ActivityTracker = new function () {
     var maxErrors = 10;
     var errors = 0;
     var active;
+    var csrfToken;
 
     this.init = function (restContext, workspaceId) {
         this.url = restContext + "/activity/" + workspaceId;
         document.addEventListener("mousemove",  ActivityTracker.setActive);
         document.addEventListener("keypress", ActivityTracker.setActive);
-        setInterval(ActivityTracker.sendRequest, timeoutInterval);
+        ActivityTracker.requestCsrfToken(restContext,
+                                         function() {
+                                            setInterval(ActivityTracker.sendRequest, timeoutInterval);
+                                         },
+                                         function() {
+                                            console.log("Could not fetch X-CSRF-Token, activity tracking won't be started");
+                                         });
     };
 
     this.setActive = function() {
@@ -58,7 +65,28 @@ var ActivityTracker = new function () {
             }
         };
         request.open("PUT", ActivityTracker.url, true);
+        if (csrfToken) {
+            request.setRequestHeader("X-CSRF-Token", csrfToken);
+        }
         request.send();
     };
 
+    this.requestCsrfToken = function(restContext, tokenPresentHandler, tokenMissingHandler) {
+        var xhr = new XMLHttpRequest();
+
+        xhr.onreadystatechange = function() {
+            if (xhr.readyState == 4) {
+                csrfToken = xhr.getResponseHeader("X-CSRF-Token");
+                if (csrfToken) {
+                    tokenPresentHandler(csrfToken);
+                } else {
+                    tokenMissingHandler();
+                }
+            }
+        };
+
+        xhr.open("GET", restContext + "/profile");
+        xhr.setRequestHeader("X-CSRF-Token", "Fetch");
+        xhr.send();
+    }
 };
--- a/dashboard/src/app/index.module.ts
+++ b/dashboard/src/app/index.module.ts
@@ -218,8 +218,62 @@ initModule.factory('AddMachineTokenToUrlInterceptor', ($injector, $q) => {
   };
 });
 
+// Prevents CSRF(see https://en.wikipedia.org/wiki/Cross-site_request_forgery)
+// using additional token header, see
+// https://tomcat.apache.org/tomcat-7.0-doc/config/filter.html#CSRF_Prevention_Filter_for_REST_APIs
+initModule.factory('CsrfPreventionInterceptor', ($injector, $rootScope) => {
+  const CSRF_TOKEN_HEADER_NAME = 'X-CSRF-Token';
+  let csrfToken = "";
+
+  function isModifyingMethod(method: string): boolean {
+    return method === "POST" || method === "PUT" || method === "DELETE";
+  };
+
+  function isMachineRequest(url: string): boolean {
+    let agents = $injector.get('cheWorkspace').getWorkspaceAgents();
+    for (let agent of agents.values()) {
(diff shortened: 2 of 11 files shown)


benoitf:
it looks like there is a conflict

evoevodin:
yeah i see going to fix it

assembly/assembly-ide-war/src/main/webapp/_app/activity.js
```diff
@@ -58,7 +65,28 @@ var ActivityTracker = new function () {
...
+            request.setRequestHeader("X-CSRF-Token", csrfToken);
+        }
         request.send();
     };
 
+    this.requestCsrfToken = function(restContext, tokenPresentHandler, tokenMissingHandler) {
+        var xhr = new XMLHttpRequest();
+
+        xhr.onreadystatechange = function() {
+            if (xhr.readyState == 4) {
```
benoitf: can we use https://xhr.spec.whatwg.org/#xmlhttprequest DONE attribute ?

dashboard/src/app/index.module.ts
```diff
@@ -218,8 +218,62 @@ initModule.factory('AddMachineTokenToUrlInterceptor', ($injector, $q) => {
   };
 });
 
+// Prevents CSRF(see https://en.wikipedia.org/wiki/Cross-site_request_forgery)
+// using additional token header, see
+// https://tomcat.apache.org/tomcat-7.0-doc/config/filter.html#CSRF_Prevention_Filter_for_REST_APIs
+initModule.factory('CsrfPreventionInterceptor', ($injector, $rootScope) => {
+  const CSRF_TOKEN_HEADER_NAME = 'X-CSRF-Token';
+  var csrfToken = "";
```
benoitf: in TS, you could use let

dashboard/src/app/index.module.ts
```diff
@@ -218,8 +218,62 @@ initModule.factory('AddMachineTokenToUrlInterceptor', ($injector, $q) => {
...
+
+  function isModifyingMethod(method) {
```
benoitf: return type (boolean) should be added

assembly/assembly-ide-war/src/main/webapp/_app/activity.js
evoevodin: sure, thx

assembly/assembly-ide-war/src/main/webapp/_app/activity.js
```diff
@@ -58,7 +65,28 @@ var ActivityTracker = new function () {
             }
         };
         request.open("PUT", ActivityTracker.url, true);
+        if (csrfToken) {
+            request.setRequestHeader("X-CSRF-Token", csrfToken);
+        }
         request.send();
     };
 
+    this.requestCsrfToken = function(restContext, tokenPresentHandler, tokenMissingHandler) {
```
mshaposhnik: Why not obtain token in extension using RequestFactory and just pass as init param?

dashboard/src/app/index.module.ts
```diff
@@ -218,8 +218,62 @@ initModule.factory('AddMachineTokenToUrlInterceptor', ($injector, $q) => {
...
+    return method === "POST" || method === "PUT" || method === "DELETE";
+  };
+
+  function isMachineRequest(url) {
```
benoitf: return type ?

site/app/site/scripts/models/account.js
```diff
@@ -576,4 +617,5 @@
             }
         };
     });
+
```
benoitf: up to you with this new line :-)

assembly/assembly-ide-war/src/main/webapp/_app/activity.js
mshaposhnik: Ahh seems is used alwo in Websocket terminal, so no way..

dashboard/src/app/index.module.ts
evoevodin: boolean :smile: 
i will add return types

dashboard/src/app/index.module.ts
evoevodin: ok

wsmaster/codenvy-hosted-sso-client/src/main/java/com/codenvy/auth/sso/client/CodenvyCsrfFilter.java
```diff
@@ -0,0 +1,84 @@
...
+        if (containsSessionCookie(((HttpServletRequest)request).getCookies())) {
+            csrfPreventionFilter.doFilter(request, response, chain);
+        } else {
+            chain.doFilter(request, response);
+        }
+    }
+
+    private boolean containsSessionCookie(Cookie[] cookies) {
+        if (cookies != null) {
+            for (Cookie cookie : cookies) {
```
benoitf: should we use streams ? 
Optional.ofNullable / anyMatch

evoevodin: well, this code is going to be executed for each request so it has to be fast and consume minimum resources, i would stick with the current approach + i like it more =)

site/app/site/scripts/views/form.js
```diff
@@ -29,6 +29,10 @@ define(["jquery", "underscore", "backbone", "models/account", "validation"],
                         window.location = "/site/maintenance";
                     }
                 });
+
+                // make first get request which fetches csrf token
+                Account.getUserSettings();
```
vkuznyetsov: Account.getUserSettings() method returns success in both cases (if user logged in it returns his settings, in other case it returns an empty object).  What will happen if you get unexpected error in response?

evoevodin: AFAIU `initialize` is called each time html page is loaded, the idea is that if user is logged in 
`Account.getUserSettings()` will be intercepted by `CsrfPreventionRequestInteceptor` which will get the value from  _X-CSRF-Token_ header. If user is not logged in, it's not a big deal then request will fail unauthorized exception and token won't be fetched -  which is okay because unauthorized user is not vulnerable to CSRF attacks.
I considered another approach, which is to go for token when the first modification request is made but i found it more complex. 
Do you have any idea how to implement it in a better way?

evoevodin:
The merge of this PR depends on https://github.com/codenvy/qa/issues/872.

evoevodin:
[qa passed](https://ci.codenvycorp.com/view/qa/job/a3-selenium-tests/15/)

Issue #2221, opened 2017-05-22
Make it possible to customize the oauth buttons order in custom assemblies.


vkuznyetsov:
Now we have oauth buttons ordered in login page randomly.

It needs to allow to set the oauth buttons order.

vkuznyetsov:
needed for https://github.com/codenvy/redhat/issues/18

PR #2225, opened 2017-05-23
show oauth buttons in defined order

vkuznyetsov:
resolves https://github.com/codenvy/codenvy/issues/2221


pom.xml
```diff
@@ -69,7 +69,7 @@
         <org.osgi.version>5.0.0</org.osgi.version>
         <org.scala.lang.version>2.10.4</org.scala.lang.version>
         <org.thymeleaf.version>3.0.3.RELEASE</org.thymeleaf.version>
-        <site.classifier>onprem-se</site.classifier>
+        <site.classifier>stage</site.classifier>
```
vkuznyetsov: should be onprem-se

Issue #2228, opened 2017-05-23
Regression: Fallback to start from recipe if snapshot is unavailable doesn't work


skryzhny:
As described in #2227 start from snapshots doesn't work after renaming host.
It is expected to start workspace from recipe if snapshot is unavailable.
But I'm getting errors like:
```
[DOCKER] Step 1/1 : FROM some-host.com:5000/machine_snapshot_z7mij6c43mghah5d:latest
[DOCKER] [ERROR] Get https://some-host.com:5000/v1/_ping: http: server gave HTTP response to HTTPS client
```

**Reproduction Steps:**
Same as in #2227 

**Codenvy version:**   5.10.0

mmorhun:
Done

PR #2262, opened 2017-06-13
CODENVY-2888: Fallback to start from recipe if snapshot is unavailable

mmorhun:
### What does this PR do?
Handles situation when docker failed to pull snapshot image.

https://github.com/codenvy/codenvy/issues/2228

#### Changelog
Fallback to container recipe if snapshot image is unavailable.

#### Release Notes
N/A

#### Docs PR
N/A

### DIFF
--- a/plugins/plugin-hosted/codenvy-machine-hosted/src/main/java/com/codenvy/machine/HostedMachineProviderImpl.java
+++ b/plugins/plugin-hosted/codenvy-machine-hosted/src/main/java/com/codenvy/machine/HostedMachineProviderImpl.java
@@ -37,6 +37,7 @@ import org.eclipse.che.plugin.docker.client.DockerConnectorProvider;
 import org.eclipse.che.plugin.docker.client.ProgressMonitor;
 import org.eclipse.che.plugin.docker.client.UserSpecificDockerRegistryCredentialsProvider;
 import org.eclipse.che.plugin.docker.client.exception.ContainerNotFoundException;
+import org.eclipse.che.plugin.docker.client.exception.DockerException;
 import org.eclipse.che.plugin.docker.client.exception.ImageNotFoundException;
 import org.eclipse.che.plugin.docker.client.json.ContainerConfig;
 import org.eclipse.che.plugin.docker.client.params.BuildImageParams;
@@ -210,7 +211,7 @@ public class HostedMachineProviderImpl extends MachineProviderImpl {
                                               .withBuildArgs(buildArgs),
                               progressMonitor);
 
-        } catch (ImageNotFoundException e) {
+        } catch (DockerException e) {
             throw new SourceNotFoundException(e.getLocalizedMessage(), e);
         } catch (IOException e) {
             throw new MachineException(e.getLocalizedMessage(), e);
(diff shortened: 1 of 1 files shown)


plugins/plugin-hosted/codenvy-machine-hosted/src/main/java/com/codenvy/machine/HostedMachineProviderImpl.java
```diff
@@ -210,7 +211,7 @@ protected void pullImage(CheServiceImpl service,
                                               .withBuildArgs(buildArgs),
                               progressMonitor);
 
-        } catch (ImageNotFoundException e) {
+        } catch (DockerException e) {
```
garagatyi: I wonder how this behaves in case image is present but some other error occurred?

mmorhun: I believe this behaves in the correct way here.
Please, take into account that this happens inside `HostedMachineProvider.pullImage`and `DockerException` can be thrown only by `DockerConnector.buildImage` method. Looking deeply, `DockerException` can be thrown from only one place where docker service response is analyzed.
According to [docker reference](https://docs.docker.com/engine/api/v1.29/#operation/ImageBuild) build endpoint could return 200, 400 and 500 error only. In our case only 500 error code is possible. Also, build is made from the following recipe: `FROM <snapshot image>`. There is no 'recipe failure'. This means, that it could be failure of: Codenvy node or network or registry. No other options. Hence:
- If the problem is on Codenvy side (i.g. not enough resources) then this should be fixed by admins and we cannot do anything here (it will fail anyway).
- If the problem is network connection or docker registry is unavailable  then fallback to the workspace recipe is the right behaviour according to spec.

garagatyi: Pull is used in case snapshot is not used for machine. This means that this change will affect no-snapshot case. 
Do you think that in this case any internal error should be treated as image not found error?

mmorhun: Yeah, good catch!

plugins/plugin-hosted/codenvy-machine-hosted/src/main/java/com/codenvy/machine/HostedMachineProviderImpl.java
```diff
@@ -209,9 +211,13 @@ protected void pullImage(CheServiceImpl service,
                                               .withCpuQuota(cpuQuota)
                                               .withBuildArgs(buildArgs),
                               progressMonitor);
-
         } catch (ImageNotFoundException e) {
             throw new SourceNotFoundException(e.getLocalizedMessage(), e);
+        } catch (DockerException e) {
+            if (image != null && SNAPSHOT_LOCATION_PATTERN.matcher(image).matches()) {
```
garagatyi: Can you add a comment to the code that explains why we do this tricky thing?

PR #2271, opened 2017-06-17
CHE-4518: change styles for input boxes

olexii4:
### What does this PR do?
Change styles for input boxes.

https://github.com/eclipse/che/issues/4518

#### Changelog
-  Change styles for input boxes for UD.

#### Release Notes
On our road to Che 6, we have refined the input boxes in the Dashboard so that they look prettier and better to interact with. The different states of the input boxes are well defined and they allow the user to properly identify if the input is having the focus and is in edit mode. Similarly, it is now clearer to understand the mandatory fields and when the values are incorrect. As for the styles of the input boxes, they are simpler and easier to identify in the page.

#### Docs PR
N/A
1.When password changing
selection_018


dashboard/bower.json
```diff
@@ -31,7 +31,6 @@
     "jsonlint": "1.6.0",
     "ng-clip": "0.2.6",
     "ng-lodash": "0.2.3",
-    "ng-password-strength": "0.2.1",
```
benoitf: hello Oleksii, could you explain why you need to remove this library to switch style for input boxes ?

olexii4: I have a template for password input fields.
selection_019

We had the password-strength library which cannot do what we need. That is why I removed it.
FYI: The last release of our password-strength library was in 2015(last commit - on Jan 9, 2016).

benoitf: ok

slemeur:
@olexii4 I'm not sure the green shown in your screenshot is the one we defined in the mockup.
In the mokcup the color is #48D170

olexii4:
@slemeur I fixed it:
    .weak
      color #EA6759
selection_024
    .fair
      color #55A8F9
selection_023
    .good
      color #48D170
selection_022
WDYT?

olexii4:
In case with error:
selection_025

olexii4:
@slemeur Do you want to change global constant 'error color' to #48D170?

slemeur:
Good for me

PR #2297, opened 2017-07-03
Reuse workspace activity managing components from Che

mkuznyetsov:
### What does this PR do?
Use idling tracking from Che (https://github.com/eclipse/che/pull/5512)

https://github.com/eclipse/che/pull/5386

#### Changelog
Activity tracking components are now re-used from Che. Configuration properties remain the same for Codenvy.

#### Release Notes
N/A

#### Docs PR
N/A


slemeur:
change log , docs are missing

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/codenvy-api-configuration.properties
```diff
@@ -43,8 +43,6 @@ password.recovery.expiration_timeout_hours=1
 # path to templates' description folder
 che.template.storage=${catalina.base}/templates
 
-stop.workspace.scheduler.period=60
```
garagatyi: Looks like this property is missing now

mkuznyetsov: added property back

mkuznyetsov:
@slemeur Added Changelog. There are no changes needed for docs, as this feature will still works in the same way.

PR #2314, opened 2017-07-12
Update build.include

svor:


benoitf:
I would have approved this one ^^

PR #2336, opened 2017-07-21
Introduce selenium tests

tolusha:
Introduces selenium tests for Codenvy on-prem project (an issue https://github.com/codenvy/qa/issues/978).


pom.xml
```diff
@@ -830,6 +836,16 @@
                 <version>${che.version}</version>
             </dependency>
             <dependency>
+                <groupId>org.eclipse.che.selenium</groupId>
+                <artifactId>che-selenium-core</artifactId>
```
skabashnyuk: no need to declare it here. It will come with che parent.

pom.xml
```diff
@@ -830,6 +836,16 @@
...
+                <version>${che.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>org.eclipse.che.selenium</groupId>
```
skabashnyuk: no need to declare it here. It will come with che parent.

PR #2346, opened 2017-08-03
Use HttpSshServiceClient instead of TestSshServiceClient

dmytro-ndp:
This request fixes issue https://github.com/codenvy/qa/issues/1067: replaces **TestSshServiceClient** by original client [HttpSshServiceClient](https://github.com/eclipse/che/blob/master/wsagent/che-core-ssh-key-server/src/main/java/org/eclipse/che/plugin/ssh/key/HttpSshServiceClient.java).

### DIFF
--- a/selenium/codenvy-selenium-test/pom.xml
+++ b/selenium/codenvy-selenium-test/pom.xml
@@ -77,6 +77,10 @@
         </dependency>
         <dependency>
             <groupId>org.eclipse.che.core</groupId>
+            <artifactId>che-core-api-ssh-shared</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.eclipse.che.core</groupId>
             <artifactId>che-core-api-user-shared</artifactId>
         </dependency>
         <dependency>
@@ -98,6 +102,10 @@
         <dependency>
             <groupId>org.eclipse.che.core</groupId>
             <artifactId>che-core-commons-lang</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.eclipse.che.plugin</groupId>
+            <artifactId>che-plugin-ssh-key-server</artifactId>
         </dependency>
         <dependency>
             <groupId>org.eclipse.che.selenium</groupId>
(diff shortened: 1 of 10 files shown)


selenium/codenvy-selenium-test/pom.xml
```diff
@@ -147,6 +155,11 @@
             <scope>runtime</scope>
         </dependency>
         <dependency>
+            <groupId>org.eclipse.che.plugin</groupId>
+            <artifactId>che-plugin-github-server</artifactId>
+            <scope>runtime</scope>
```
skabashnyuk: runtime?

dmytro-ndp: Yes, because there is implementation of **GitHubKey** DTO class in **che-plugin-github-server** which isn't referenced in tests directly.

skabashnyuk: Why it's declared explicitly?

dmytro-ndp: Because default scope is **compile**.

dmytro-ndp: I checked it again and realized that **runtime** scope is really redundant here.
Thank you for notice, Sergii!

tolusha: Tests require DTOs implementation at runtime.

dmytro-ndp:
Close without merging so as **HttpSshServiceClient** [is being re-written](https://github.com/eclipse/che/blob/dd665669589f24d05a8a068d72faa60d956008f0/ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/ssh/SshServiceClientImpl.java#L53) to use [Promises](https://github.com/eclipse/che/blob/master/ide/commons-gwt/src/main/java/org/eclipse/che/api/promises/client/js/Promises.java#L45) class which in turn is based on native JavaScript code. Another issue that it will use [DtoFactory](https://github.com/eclipse/che/blob/master/ide/commons-gwt/src/main/java/org/eclipse/che/ide/dto/DtoFactory.java) differ for [server-side one](https://github.com/eclipse/che/blob/master/core/che-core-api-dto/src/main/java/org/eclipse/che/dto/server/DtoFactory.java) and it requires [dedicated code](https://github.com/eclipse/che/blob/master/ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/client/BootstrapController.java#L91-L92) to generate Dto-classes implementations.

Issue #2351, opened 2017-08-04
Remove the fair source license check in the CLI during installation


riuvshin:
needed for https://github.com/codenvy/codenvy/issues/2350

riuvshin:
@TylerJewell I did quick investigation on this and here is my thoughts on what should we do.
I don;t think we have to remove license check functionality from CLI.
In CHE we have this:
[DEFAULT_CHE_LICENSE=false](https://github.com/eclipse/che/blob/master/dockerfiles/base/scripts/base/startup_01_init.sh#L154) which means that no license accept will be shown to user.
[DEFAULT_CHE_LICENSE_URL="https://www.eclipse.org/legal/epl-v10.html"](https://github.com/eclipse/che/blob/master/dockerfiles/base/scripts/base/startup_01_init.sh#L125)

In Codenvy:
[CHE_LICENSE=true](https://github.com/codenvy/codenvy/blob/master/dockerfiles/cli/scripts/pre_init.sh#L26)
[CHE_LICENSE_URL="https://codenvy.com/legal/fair-source/"](https://github.com/codenvy/codenvy/blob/master/dockerfiles/cli/scripts/pre_init.sh#L27)

So my proposal is just make license settings for Codenvy / saas same as in CHE WDYT?

TylerJewell:
+1

I forgot that we made it that flexible!

riuvshin:
@TylerJewell you did :)

Issue #2350, opened 2017-08-04
EPL Codenvy on-prem / remove licenses


riuvshin:
- [x] Remove the fair source license check in the CLI during installation https://github.com/codenvy/codenvy/issues/2351 @riuvshin
- [x] Remove the fair source license check in the site during first login https://github.com/codenvy/codenvy/issues/2352 @vkuznyetsov 
- [x] Remove all elements of license enforcement from the server - Codenvy binary will be unrestricted for total seats, total ram, total servers. @tolusha 
- [x] Remove any license key input from the UD. @ashumilova  
- [x] Remove any restrictions that prevent team / group / scale capabilities from the UD. @ashumilova 
- [x] docs clean up https://github.com/codenvy/docs/issues/131 @bmicklea or @slemeur

riuvshin:
@bmicklea @slemeur hey guys we need remove license stuff before end of current sprint we all are almost ready but we need help on docs. We have to be ready to close this today.

bmicklea:
I'm on PTO today but I'll try and get to it later this afternoon

riuvshin:
@benoitf Anatoly already proposed PR we just need review.

PR #2353, opened 2017-08-04
CLI: disable license check, update license url, remove license props

riuvshin:
### What does this PR do?
CLI: disable license check, update license url, remove license props

https://github.com/codenvy/codenvy/issues/2351

#### Changelog
CLI: disable license check, update license url, remove license props


riuvshin:
@tolusha this PR should be merged ONLY along with another PR (probably on your side) with removing license stuff from codebase because I've removed license props here.

dockerfiles/cli/scripts/pre_init.sh
```diff
@@ -23,8 +23,8 @@ pre_init() {
   WS_AGENT_IN_REPO="${WS_AGENT_IN_REPO_MODULE_NAME}/target/assembly-wsagent-server-*.tar.gz"
   WS_AGENT_ASSEMBLY="ws-agent.tar.gz"
   CHE_SCRIPTS_CONTAINER_SOURCE_DIR="/repo/dockerfiles/cli/scripts"
-  CHE_LICENSE=true
-  CHE_LICENSE_URL="https://codenvy.com/legal/fair-source/"
+  CHE_LICENSE=false
+  CHE_LICENSE_URL="https://www.eclipse.org/legal/epl-v10.html"
```
benoitf: we can remove these two lines (defaut is false and EPL)

riuvshin: I also was thinking about that because it is defined in base so it could be taken from the parent, but then I decided to keep it there to make it clear that it should be defined  in case of overriding. And if we will need back this it will be much easier to understand what to do. So if you don't mind I would like to keep this.

dmytro-ndp:
@riuvshin: we need to merge it into master ASAP so as https://github.com/codenvy/codenvy/issues/2350 has already been there.

riuvshin:
@dmytro-ndp to merge that PR @tolusha should remove components that using props that this PR removes because product will be broken. @tolusha is that ready? last time I've checked this week or   two ago it was still failing.

Issue #2400, opened 2017-09-05
build codenvy UD in docker same as in CHE


riuvshin:
we have an issue with build codenvy UD with different versions of node, to avoid that we have to build in docker with predictable env. In CHE we already doing that but in codenvy we still have a risk that same sources may work in different way depending on node version used to build UD.
node version should be 6.11.2

depends on: https://github.com/codenvy/codenvy/issues/2399

riuvshin:
no longer actual.

PR #2401, opened 2017-09-06
build codenvy dashboard in docker

riuvshin:
- build codenvy dashboard in docker
- add native profile
needed for: https://github.com/codenvy/codenvy/issues/2400


dashboard/Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+# It builds an archive file that can be used by doing later
+#  $ docker run --rm eclipse-che-dashboard | tar -C target/ -zxf -
+FROM mhart/alpine-node:6.11.2
+
+RUN apk update && \
+    apk add --no-cache git
+COPY package.json /dashboard/
+RUN cd /dashboard && npm install
+COPY bower.json /dashboard/
+RUN cd /dashboard && ./node_modules/.bin/bower install --allow-root
```
benoitf: I think we will have an issue there because inside bower.json we're specifying to clone upstream che
but if we execute this Dockerfile a second time it won't clone again che, then we'll use previous cloned version

dashboard/Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+COPY . /dashboard/
+RUN cd /dashboard && ./node_modules/.bin/typings install
+COPY . /dashboard/
+RUN cd /dashboard && ./node_modules/.bin/gulp test
```
benoitf: I think we can replace gulp test by adding build as well
(no need to copy twice)

dashboard/Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+
+RUN apk update && \
+    apk add --no-cache git
+COPY package.json /dashboard/
+RUN cd /dashboard && npm install
+COPY bower.json /dashboard/
+RUN cd /dashboard && ./node_modules/.bin/bower install --allow-root
+COPY . /dashboard/
+RUN cd /dashboard && ./node_modules/.bin/typings install
+COPY . /dashboard/
```
benoitf: why do we need a new copy ?

dashboard/Dockerfile
```diff
@@ -12,17 +12,19 @@ FROM node:6.11.2
 
 RUN apt-get update && \
     apt-get install -y git
-COPY package.json /dashboard/
+RUN mkdir -p /dashboard/.tmp
+RUN useradd -ms /bin/bash node-user
```
benoitf: it's a wip but before merging we should merge all RUN operations and add cleanup calls https://github.com/eclipse/che/blob/master/dashboard/Dockerfile#L15-L16

riuvshin:
@benoitf finally! it seems now it works :) need to add native profile and fix TODOs in docker file :)

benoitf:
wow

dashboard/Dockerfile
riuvshin: sure :)

riuvshin:
btw we will need introduce same 
```
 +RUN useradd -ms /bin/bash node-user
 +COPY . /dashboard/
 +RUN chown -R node-user:node-user /dashboard/
```

in che

benoitf:
is there a way to avoid the chown operation ?

riuvshin:
@benoitf no :( docker ADD and COPY commands adding files as root and then our user won't be able to write. Well it was the fastest solution maybe we could improve that

benoitf:
@riuvshin yes sure it's not mandatory, just thinking if we could use another way to avoid chown operations

riuvshin:
@benoitf @ashumilova now it is ready for review.
But I still doubt about this:

```
# workaround to prevent caching old che-dashboard
ADD https://api.github.com/repos/eclipse/che/git/refs/heads/master /tmp/che-dashboard-version.json
```

benoitf:
@riuvshin should we close this PR ?

riuvshin:
yes, this is no longer actual.

Issue #2399, opened 2017-09-05
adapt codenvy UD to node 6.11.2


riuvshin:
we need to sync node version that we use for CHE and codenvy, in che we use node 6.11.2

riuvshin:
set as blocker as it will block next release.

PR #2403, opened 2017-09-08
Revisit dependencies to support newer node version

ashumilova:
Signed-off-by: Anna Shumilova <ashumilo@redhat.com>

### What does this PR do?
Reduces dashboard build time, fixes stylus plugin fails on node version >= 6.

https://github.com/codenvy/codenvy/issues/2399

#### Changelog
Reduced dashboard build time, fixed stylus plugin fails on node version 6.

#### Release Notes
N/A

#### Docs PR
N/A

### DIFF
--- a/dashboard/src/app/index.styl
+++ b/dashboard/src/app/index.styl
@@ -69,6 +69,7 @@
     padding 0
     min-width 40px
     min-height 40px
+
   button.md-button:disabled
     border 1px solid $cat-gray-color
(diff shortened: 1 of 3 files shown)


dashboard/src/app/index.styl
```diff
@@ -69,6 +69,7 @@
     padding 0
     min-width 40px
     min-height 40px
+
```
benoitf: maybe not required for this PR

PR #2424, opened 2017-09-20
interrupt build scripts for docker files if failure occurs

riuvshin:
### What does this PR do?
interupt build scripts for docker files if failure occurs

#### Changelog
interupt build scripts for docker files if failure occurs

### DIFF
--- a/dockerfiles/build.include
+++ b/dockerfiles/build.include
@@ -8,6 +8,9 @@
 # Contributors:
 #   Florent Benoit - Initial Implementation
 
+set -e
+set -u
+
 skip_update() {
   for i in "$@" ; do
     if [ $i = "--skip-update" ]; then


dockerfiles/cli/build.sh
```diff
@@ -8,6 +8,8 @@
 # Contributors:
 #   Florent Benoit - Initial Implementation
 
+set -e
```
benoitf: do we need it as we include build.include ?

riuvshin: not sure, but just in case I added this :)

benoitf: I think we can drop on all .sh that imports build.include

also you should add `set -u` as well in build.include

riuvshin: are you sure to remove on sh? my issue was that this https://github.com/codenvy/codenvy/blob/master/dockerfiles/agents/build.sh#L16 line failed but script passed. it was before build.include was used.

you mean `. ${DIR}/../build.include`  will apply `set -e` for build.sh ?

riuvshin: tested, `. ${DIR}/../build.include` will apply `set -e`

benoitf: yes

riuvshin: done

Issue #2483, opened 2017-10-12
Fix problem with run organization tests


Ohrimenko1988:
After merge che-multiuser branch to the master we have problem with run organizations selenium tests. We need to investigate, find reason and solve this problem.

Ohrimenko1988:
Done https://github.com/codenvy/codenvy/pull/2487

PR #2484, opened 2017-10-12
Fix organizations selenium tests

Ohrimenko1988:
### What does this PR do?
Fix problem with run organizations selenium tests

Resolve this issue https://github.com/codenvy/codenvy/issues/2483


selenium/codenvy-selenium-test/src/test/java/com/codenvy/selenium/dashboard/onprem_organization/BulkOnpremOrganizationDeletionTest.java
```diff
@@ -58,7 +59,7 @@
 
   @BeforeClass
   public void setUp() throws Exception {
-    dashboard.open();
+    dashboard.open(adminTestUser.getName(), adminTestUser.getPassword());
```
dmytro-ndp: If we need to add almost complete duplicates of organization tests of Che only for usage of `dashboard.open(adminTestUser.getName(), adminTestUser.getPassword())` I would prefer to use `dashboard.open(String userName, String userPassword)` method in Che organization tests with nullable paremeters, and in case when they = null we can use default user credentials instead.

PR #2485, opened 2017-10-12
Update mockito to v2.10

benoitf:
requires
- [ ] https://github.com/eclipse/che-dependencies/pull/62
- [ ] https://github.com/eclipse/che/pull/6394

### What does this PR do?
Update mockito to v2.10
it allows to start to use Java9 as runtime

#### Release Notes
N/A

#### Docs PR
N/A

### DIFF
--- a/plugins/plugin-hosted/codenvy-machine-hosted/src/test/java/com/codenvy/machine/backup/WorkspaceFsBackupSchedulerTest.java
+++ b/plugins/plugin-hosted/codenvy-machine-hosted/src/test/java/com/codenvy/machine/backup/WorkspaceFsBackupSchedulerTest.java
@@ -12,6 +12,7 @@ package com.codenvy.machine.backup;
 
 import static java.lang.Thread.sleep;
 import static org.mockito.Matchers.eq;
+import static org.mockito.Mockito.after;
 import static org.mockito.Mockito.doAnswer;
 import static org.mockito.Mockito.doThrow;
 import static org.mockito.Mockito.mock;
@@ -148,7 +149,7 @@ public class WorkspaceFsBackupSchedulerTest {
     scheduler.scheduleBackup();
 
     // then
-    verify(backupManager, timeout(2000).never()).backupWorkspace(wsId);
+    verify(backupManager, after(2000).never()).backupWorkspace(wsId);
   }
 
   @Test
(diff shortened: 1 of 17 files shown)


benoitf:
ci-build

plugins/plugin-hosted/codenvy-machine-hosted/src/test/java/com/codenvy/machine/backup/WorkspaceFsBackupSchedulerTest.java
```diff
@@ -171,7 +172,7 @@ public void shouldNotBackupWSWithNonRunningStatus() throws Exception {
     // add this verification with timeout to ensure that thread executor had enough time before verification of call
     verify(backupManager, timeout(2000)).backupWorkspace(WORKSPACE_ID_1);
     verify(backupManager, timeout(2000)).backupWorkspace(WORKSPACE_ID_2);
-    verify(backupManager, timeout(2000).never()).backupWorkspace("ws3");
```
garagatyi: @benoitf Can you elaborate on why did you change it?

benoitf: sure, 
>
https://static.javadoc.io/org.mockito/mockito-core/1.10.19/org/mockito/verification/VerificationWithTimeout.html
 never()
Deprecated. 
Validation with timeout combined with never simply does not make sense, as never() will typically immediately pass, and therefore not wait the timeout. The behaviour you may be looking for is actually provided by after().never().
To avoid compilation errors upon upgrade the method is deprecated and it throws a "friendly reminder" exception.
In a future release we will remove timeout(x).atMost(y) and timeout(x).never() from the API.
Do you want to find out more? See issue 235

In 2.x the deprecated stuff has been removed so I switch to after(timeout).never()

riuvshin:
ci-test

riuvshin:
ci-build

plugins/plugin-hosted/codenvy-machine-hosted/src/test/java/com/codenvy/machine/backup/WorkspaceFsBackupSchedulerTest.java
garagatyi: tnx for the explanation

benoitf:
ci-build

Issue #2477, opened 2017-10-11
Fix tests in "factory" package


Ohrimenko1988:
After merge "che-multiuser" branch into master we need to fix next test: DirectUrlFactoryWithRootFolder

Ohrimenko1988:
Done 
https://github.com/codenvy/codenvy/pull/2494
https://github.com/eclipse/che/pull/6789

PR #2494, opened 2017-10-18
Fix problem with redirecting to the factory page after login

Ohrimenko1988:
### What does this PR do?
Fix problem with redirecting to the factory page after login, when webdriver.get() method used very fast several times one by one, this method does not work as we expected.

Issue https://github.com/codenvy/codenvy/issues/2477


dmytro-ndp:
ci-built

Issue #2471, opened 2017-10-10
Rework AuthenticateAndAcceptFactoryThroughGitHubOAuthTest test to remove github user after test execution


dmytro-ndp:
There is **AuthenticateAndAcceptFactoryThroughGitHubOAuthTest** which has [_tearDown()_](https://github.com/codenvy/codenvy/blob/master/selenium/codenvy-selenium-test/src/test/java/com/codenvy/selenium/factory/AuthenticateAndAcceptFactoryThroughGitHubOAuthTest.java#L69) method aimed to remove github test user, but it is broken and remove default user instead.
To fix it we should provide github user email so as **testUserServiceClient.findByName(gitHubUsername)** doesn't work as expected.

**Codenvy version:**    
5.19.0-SNAPSHOT

**OS and version:**    

**Docker version:**    

**Codenvy cli.log output:**

Ohrimenko1988:
Done

PR #2498, opened 2017-10-20
Rework AuthenticateAndAcceptFactoryThroughGitHubOAuthTest 

Ohrimenko1988:
### What does this PR do?
Rework cleanup after test .

Issue https://github.com/codenvy/codenvy/issues/2471

### DIFF
--- a/selenium/codenvy-selenium-test/src/test/java/com/codenvy/selenium/factory/AuthenticateAndAcceptFactoryThroughGitHubOAuthTest.java
+++ b/selenium/codenvy-selenium-test/src/test/java/com/codenvy/selenium/factory/AuthenticateAndAcceptFactoryThroughGitHubOAuthTest.java
@@ -74,6 +74,13 @@ public class AuthenticateAndAcceptFactoryThroughGitHubOAuthTest {
 
   @AfterClass
   public void tearDown() throws Exception {
+
+    seleniumWebDriver.get(getDashboardWorkspaceUrl());
+    dashboardWorkspace.waitToolbarTitleName("Workspaces");
+    dashboardWorkspace.waitListWorkspacesOnDashboard();
+
+    deleteAllWorkspaces(getAllWorkspaceNames());
+
     String userGitHubEmail =
         testGitHubServiceClient.getUserPublicEmail(gitHubUsername, gitHubPassword);
 
@@ -103,13 +110,6 @@ public class AuthenticateAndAcceptFactoryThroughGitHubOAuthTest {
 
     projectExplorer.waitProjectExplorer();
     notificationsPopupPanel.waitExpectedMessageOnProgressPanelAndClosed("Project Spring imported");
-
-    seleniumWebDriver.get(getDashboardWorkspaceUrl());
-    dashboardWorkspace.waitToolbarTitleName("Workspaces");
-    dashboardWorkspace.waitListWorkspacesOnDashboard();
-
-    deleteAllWorkspaces(getAllWorkspaceNames());
-    testFactory.delete();
   }
 
   private void deleteAllWorkspaces(List<String> workspaces) {
--- a/selenium/codenvy-selenium-test/src/test/java/com/codenvy/selenium/factory/AuthenticateAndAcceptFactoryThroughGitHubOAuthTest.java
+++ b/selenium/codenvy-selenium-test/src/test/java/com/codenvy/selenium/factory/AuthenticateAndAcceptFactoryThroughGitHubOAuthTest.java
@@ -13,10 +13,13 @@ package com.codenvy.selenium.factory;
 import com.codenvy.selenium.pageobject.site.LoginAndCreateOnpremAccountPage;
 import com.google.inject.Inject;
 import com.google.inject.name.Named;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
 import org.eclipse.che.api.core.model.user.User;
 import org.eclipse.che.selenium.core.SeleniumWebDriver;
+import org.eclipse.che.selenium.core.client.TestGitHubServiceClient;
 import org.eclipse.che.selenium.core.client.TestUserServiceClient;
-import org.eclipse.che.selenium.core.client.TestWorkspaceServiceClient;
 import org.eclipse.che.selenium.core.client.TestWorkspaceServiceClientFactory;
 import org.eclipse.che.selenium.core.factory.FactoryTemplate;
 import org.eclipse.che.selenium.core.factory.TestFactory;
@@ -29,6 +32,8 @@ import org.eclipse.che.selenium.pageobject.Ide;
 import org.eclipse.che.selenium.pageobject.NotificationsPopupPanel;
 import org.eclipse.che.selenium.pageobject.Profile;
 import org.eclipse.che.selenium.pageobject.ProjectExplorer;
+import org.eclipse.che.selenium.pageobject.dashboard.DashboardWorkspace;
+import org.testng.annotations.AfterClass;
 import org.testng.annotations.BeforeClass;
 import org.testng.annotations.Test;
 
@@ -56,38 +61,30 @@ public class AuthenticateAndAcceptFactoryThroughGitHubOAuthTest {
   @Inject private TestApiEndpointUrlProvider apiEndpointUrlProvider;
   @Inject private TestUserNamespaceResolver testUserNamespaceResolver;
   @Inject private TestWorkspaceServiceClientFactory testWorkspaceServiceClientFactory;
+  @Inject private DashboardWorkspace dashboardWorkspace;
+  @Inject private TestGitHubServiceClient testGitHubServiceClient;
 
   private TestFactory testFactory;
+  private final String WORKSPACE_TOOLBAR = "Workspaces";
 
   @BeforeClass
   public void setUp() throws Exception {
     testFactory = testFactoryInitializer.fromTemplate(FactoryTemplate.MINIMAL).build();
   }
 
-  //  @AfterClass
-  // This method removes default user instead of github user.
-  // Need to be reworked https://github.com/codenvy/codenvy/issues/2471
+  @AfterClass
   public void tearDown() throws Exception {
-    User user = testUserServiceClient.findByEmail(testUser.getEmail());
-    TestWorkspaceServiceClient workspaceServiceClient =
-        testWorkspaceServiceClientFactory.create(testUser.getEmail(), testUser.getPassword());
-    workspaceServiceClient
-        .getAll()
-        .forEach(
-            ws -> {
-              try {
-                workspaceServiceClient.delete(ws, user.getName());
-              } catch (Exception e) {
-                throw new RuntimeException(e);
-              }
-            });
+    String userGitHubEmail =
+        testGitHubServiceClient.getUserPublicEmail(gitHubUsername, gitHubPassword);
 
-    testUserServiceClient.remove(testUserServiceClient.findByEmail(testUser.getEmail()).getId());
+    User user = testUserServiceClient.findByEmail(userGitHubEmail);
+    testUserServiceClient.remove(user.getId());
     testFactory.delete();
   }
 
   @Test
   public void loginThroughGitHubOAuthAndAcceptFactory() throws Exception {
+
     testFactory.open(ide.driver());
 
     loginPage.waitMainElementsOnLoginPage();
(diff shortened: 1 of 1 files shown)


selenium/codenvy-selenium-test/src/test/java/com/codenvy/selenium/factory/AuthenticateAndAcceptFactoryThroughGitHubOAuthTest.java
```diff
@@ -107,4 +111,43 @@ public void loginThroughGitHubOAuthAndAcceptFactory() throws Exception {
     projectExplorer.waitProjectExplorer();
     notificationsPopupPanel.waitExpectedMessageOnProgressPanelAndClosed("Project Spring imported");
   }
+
+  private void deleteAllWorkspaces(List<String> workspaces) {
```
dmytro-ndp: This method is about action inside Workspaces page of dashboard and so shall be places in the appropriate page object.

selenium/codenvy-selenium-test/src/test/java/com/codenvy/selenium/factory/AuthenticateAndAcceptFactoryThroughGitHubOAuthTest.java
```diff
@@ -107,4 +111,43 @@ public void loginThroughGitHubOAuthAndAcceptFactory() throws Exception {
...
+          dashboardWorkspace.waitListWorkspacesOnDashboard();
+          dashboardWorkspace.selectWorkspaceItemName(wsName);
+          dashboardWorkspace.waitToolbarTitleName(Arrays.asList(wsName.split("/")).get(1));
+          dashboardWorkspace.clickOnDeleteWorkspace();
+          dashboardWorkspace.clickOnDeleteDialogButton();
+          dashboardWorkspace.waitToolbarTitleName(WORKSPACE_TOOLBAR);
+        });
+  }
+
+  private List<String> getAllWorkspaceNames() {
```
dmytro-ndp: see comment here https://github.com/codenvy/codenvy/pull/2498/files#r145934139

selenium/codenvy-selenium-test/src/test/java/com/codenvy/selenium/factory/AuthenticateAndAcceptFactoryThroughGitHubOAuthTest.java
```diff
@@ -107,4 +111,43 @@ public void loginThroughGitHubOAuthAndAcceptFactory() throws Exception {
...
+            workspaceName -> {
+              if (isWorkspaceName(workspaceName)) {
+                workspaces.add(workspaceName);
+              }
+            });
+
+    return workspaces;
+  }
+
+  private List<String> getNotFilteredWorkspaceNames() {
```
dmytro-ndp: see comment here https://github.com/codenvy/codenvy/pull/2498/files#r145934139

selenium/codenvy-selenium-test/src/test/java/com/codenvy/selenium/factory/AuthenticateAndAcceptFactoryThroughGitHubOAuthTest.java
```diff
@@ -107,4 +111,43 @@ public void loginThroughGitHubOAuthAndAcceptFactory() throws Exception {
...
+    return Arrays.asList(dashboardWorkspace.getTextFromListWorkspaces().split("\n"));
+  }
+
+  private boolean isWorkspaceName(String workspaceName) {
```
dmytro-ndp: see comment here https://github.com/codenvy/codenvy/pull/2498/files#r145934139

selenium/codenvy-selenium-test/src/test/java/com/codenvy/selenium/factory/AuthenticateAndAcceptFactoryThroughGitHubOAuthTest.java
```diff
@@ -107,4 +111,43 @@ public void loginThroughGitHubOAuthAndAcceptFactory() throws Exception {
...
+    return workspaceName.contains("/") && workspaceName.length() > 3;
+  }
+
+  private String getDashboardWorkspaceUrl() {
```
dmytro-ndp: see comment here https://github.com/codenvy/codenvy/pull/2498/files#r145934139

Ohrimenko1988: On my opinion, it's wrong way, we have necessary methods for work with this page. And methods which were created in the test are complicated for page object. It may be more difficult for debugging and support

dmytro-ndp: 1. Logic of removing of workspaces from dashboard is not a propose of this class named **AuthenticateAndAcceptFactoryThroughGitHubOAuthTest**.
1. Test classes should be the easiest ones for debugging and support.
1. Some tests might require the same functionality and in this case we will need to duplicate the code.
1. This test could be removed in future and in this case we will lose the code.

Ohrimenko1988: Yes, your arguments make sense, but it`s only one side of coin
1) PageObject must be as simple as possible
2) Test class will be easy to debug because if something will be wrong we just see it in the test class. So, i think, In this case, we can apply some code duplication.
3) Besides this methods were tested only for this tests, and i can`t give 100 % that they will be work correctly in all cases. So, at list, this is another separate task.

selenium/codenvy-selenium-test/src/test/java/com/codenvy/selenium/factory/AuthenticateAndAcceptFactoryThroughGitHubOAuthTest.java
```diff
@@ -56,38 +61,37 @@
...
-            ws -> {
-              try {
-                workspaceServiceClient.delete(ws, user.getName());
-              } catch (Exception e) {
-                throw new RuntimeException(e);
-              }
-            });
 
-    testUserServiceClient.remove(testUserServiceClient.findByEmail(testUser.getEmail()).getId());
+    seleniumWebDriver.get(getDashboardWorkspaceUrl());
```
dmytro-ndp: It could be simplified by using `dashboard.selectWorkspacesItemOnDashboard();` command.

selenium/codenvy-selenium-test/src/test/java/com/codenvy/selenium/factory/AuthenticateAndAcceptFactoryThroughGitHubOAuthTest.java
dmytro-ndp: Please, take a look https://en.wikipedia.org/wiki/Single_responsibility_principle

selenium/codenvy-selenium-test/src/test/java/com/codenvy/selenium/factory/AuthenticateAndAcceptFactoryThroughGitHubOAuthTest.java
Ohrimenko1988: Yes, I tried to do as you wrote, but it not worked well in "@AfterClass" method , so I used another construction.

selenium/codenvy-selenium-test/src/test/java/com/codenvy/selenium/factory/AuthenticateAndAcceptFactoryThroughGitHubOAuthTest.java
Ohrimenko1988: Ok, I can create issue for testing and move this methods to page object, will you apply this?

selenium/codenvy-selenium-test/src/test/java/com/codenvy/selenium/factory/AuthenticateAndAcceptFactoryThroughGitHubOAuthTest.java
dmytro-ndp: By the way - there is a check-box in left top of _Workspaces page_ which allows to select all workspaces and delete all of them at once, instead of removing they one by one:
screenshot-2017-10-20 codenvy workspaces

Ohrimenko1988: Yes its more simply, but we are not have methods for working with this check-box

selenium/codenvy-selenium-test/src/test/java/com/codenvy/selenium/factory/AuthenticateAndAcceptFactoryThroughGitHubOAuthTest.java
dmytro-ndp: It's better to move the method to **DashboardWorkspace** page as part of the current issue.

selenium/codenvy-selenium-test/src/test/java/com/codenvy/selenium/factory/AuthenticateAndAcceptFactoryThroughGitHubOAuthTest.java
dmytro-ndp: It a chance to add it :-)

selenium/codenvy-selenium-test/src/test/java/com/codenvy/selenium/factory/AuthenticateAndAcceptFactoryThroughGitHubOAuthTest.java
Ohrimenko1988: As I saz, I can`t give 100 % for all cases. This is another task that can take half of day. We need extra story points for this. I think the new task will be the better way.

selenium/codenvy-selenium-test/src/test/java/com/codenvy/selenium/factory/AuthenticateAndAcceptFactoryThroughGitHubOAuthTest.java
Ohrimenko1988: But this is not a goal of this issue, I used existed methods :-)

selenium/codenvy-selenium-test/src/test/java/com/codenvy/selenium/factory/AuthenticateAndAcceptFactoryThroughGitHubOAuthTest.java
```diff
@@ -107,4 +111,43 @@ public void loginThroughGitHubOAuthAndAcceptFactory() throws Exception {
...
+    workspaces.forEach(
+        wsName -> {
+          seleniumWebDriver.get(getDashboardWorkspaceUrl());
+          dashboardWorkspace.waitToolbarTitleName(WORKSPACE_TOOLBAR);
+          dashboardWorkspace.waitListWorkspacesOnDashboard();
+          dashboardWorkspace.selectWorkspaceItemName(wsName);
+          dashboardWorkspace.waitToolbarTitleName(Arrays.asList(wsName.split("/")).get(1));
+          dashboardWorkspace.clickOnDeleteWorkspace();
+          dashboardWorkspace.clickOnDeleteDialogButton();
+          dashboardWorkspace.waitToolbarTitleName(WORKSPACE_TOOLBAR);
```
dmytro-ndp: We need to ensure that workspace was completely removed.

Ohrimenko1988: Yes, you are right, good idea

dmytro-ndp: There is notification popup at the top right corner appears after workspace has been successfully removed.

selenium/codenvy-selenium-test/src/test/java/com/codenvy/selenium/factory/AuthenticateAndAcceptFactoryThroughGitHubOAuthTest.java
dmytro-ndp: Movement of code into the **DoashboardWorkspaces** page object will get rid of **getDashboardWorkspaceUrl()** method.

selenium/codenvy-selenium-test/src/test/java/com/codenvy/selenium/factory/AuthenticateAndAcceptFactoryThroughGitHubOAuthTest.java
```diff
@@ -64,30 +68,25 @@ public void setUp() throws Exception {
...
+    dashboardWorkspace.deleteAllWorkspaces();
+
+    User user = testUserServiceClient.findByEmail(userGitHubEmail);
+    testUserServiceClient.remove(user.getId());
     testFactory.delete();
   }
 
   @Test
   public void loginThroughGitHubOAuthAndAcceptFactory() throws Exception {
+
```
dmytro-ndp: Redundant empty line here.

Ohrimenko1988:
@yes, thanks for your reminder

